PR #9, opened 2025-06-04
fix: check for process.on

threepointone:
In Cloudflare Workers / the workerd runtime, `process` can be defined, but `process.on()` isn't (yet). This adds a check for the existence of `process.on` as a function before handing event handlers.

(Hi folks! I'm a tech lead on ai agents at Cloudflare. Big fan of y'all! Working on making this package work with all our stuff. There's a separate issue with the node build of `debug` being loaded, but I don't yet think that's a problem with this package. Working on fixing that separately.)

### DIFF
--- a/.changeset/clear-pans-nail.md
+++ b/.changeset/clear-pans-nail.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: ensure process.on exists and is a function before adding event handlers


dkundel-openai (changes requested):
Thanks for the fix! Looks good overall. Could you run `pnpm changeset` to generate a changeset entry? Will try to get this out ASAP

dkundel-openai:
Btw there is a dedicated test for Cloudflare in the integrations-test/ happy to adapt this with some expert feedback

threepointone:
done! btw you can add the changeset bot https://github.com/apps/changeset-bot which adds a comment on a pr if it's missing a changeset (and gives a link to autogenerate one) It looks like this https://github.com/cloudflare/agents/pull/316#issuecomment-2936725029

threepointone:
I've started experimenting here https://github.com/threepointone/openai-cloudflare-agents I'll keep you in the loop. For now, trying to figure out why `debug`'s node build gets inserted into the build... I'll get back to you on this.

dkundel-openai (approved):

dkundel-openai:
Thanks! Yeah waiting for the bot to be added :)

PR #17, opened 2025-06-04
Fix guardrail interruption for realtime session

dkundel-openai:
## Summary
- trigger `response.cancel` even when audio hasn't started
- mark the response as stopped so a new one can start

## Testing
- `pnpm build`
- `CI=1 pnpm test`
- `pnpm lint`

------
https://chatgpt.com/codex/tasks/task_i_68409f83804883318c5bf86f8a0457ca

### DIFF
--- a/packages/agents-realtime/src/openaiRealtimeEvents.ts
+++ b/packages/agents-realtime/src/openaiRealtimeEvents.ts
@@ -6,15 +6,18 @@ import type { MessageEvent as WebSocketMessageEvent } from 'ws';
 // provide better runtime validation when parsing events from the server.
 
 export const realtimeResponse = z.object({
-  id: z.string().optional(),
-  conversation_id: z.string().optional(),
-  max_output_tokens: z.number().or(z.literal('inf')).optional(),
+  id: z.string().optional().nullable(),
+  conversation_id: z.string().optional().nullable(),
+  max_output_tokens: z.number().or(z.literal('inf')).optional().nullable(),
   metadata: z.record(z.string(), z.any()).optional().nullable(),
-  modalities: z.array(z.string()).optional(),
-  object: z.literal('realtime.response').optional(),
-  output: z.array(z.any()).optional(),
-  output_audio_format: z.string().optional(),
-  status: z.enum(['completed', 'incomplete', 'failed', 'cancelled']).optional(),
+  modalities: z.array(z.string()).optional().nullable(),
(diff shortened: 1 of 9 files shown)


seratch (approved):
LGTM

dkundel-openai:
Still need to validate this myself. So do not merge

packages/agents-realtime/src/openaiRealtimeEvents.ts
```diff
@@ -6,15 +6,18 @@ import type { MessageEvent as WebSocketMessageEvent } from 'ws';
 // provide better runtime validation when parsing events from the server.
 
 export const realtimeResponse = z.object({
-  id: z.string().optional(),
-  conversation_id: z.string().optional(),
-  max_output_tokens: z.number().or(z.literal('inf')).optional(),
+  id: z.string().optional().nullable(),
```
dkundel-openai, line 9: had to mark all of these as nullable since `optional()` in Zod now explicitly requires `nullable()`

packages/agents-realtime/src/realtimeSession.ts
```diff
@@ -460,24 +461,32 @@ export class RealtimeSession<
...
-          this.#context,
-          this.#currentAgent,
-          error,
-        );
-        this.interrupt();
-
-        const feedbackText = getRealtimeGuardrailFeedbackMessage(result);
-        this.sendMessage(feedbackText);
-        break;
+    const firstTripwireTriggered = results.find(
```
dkundel-openai, line 464: This should be the same behavior as before. Just makes it more explicit that we trip on the first guardrail

packages/agents-realtime/src/realtimeSession.ts
```diff
@@ -460,24 +461,32 @@ export class RealtimeSession<
...
+      (result) => result.output.tripwireTriggered,
+    );
+    if (firstTripwireTriggered) {
+      // this ensures that if one guardrail already trips and we are in the middle of another
+      // guardrail run, we don't trip again
+      console.log(
+        'interruptedByGuardrail',
+        this.#interruptedByGuardrail[responseId],
+      );
+      if (this.#interruptedByGuardrail[responseId]) {
```
dkundel-openai, line 470: Without this there might be jankyness since other guardrail checks might have been kicked off already. So this makes sure that we are only tripping once per response

PR #22, opened 2025-06-05
fix: Add init option validation to detect client side agents as tools to RealtimeAgent

seratch:
This pull request proposes adding a new input validation to RealtimeAgent constructor. This check is helpful for developers to learn why local agents as tools on the client side does not work and what to do next.

Implementation details can be changed as necessary. Still in draft mode but please feel free to add comments and let me know what you think.

### DIFF
--- a/.changeset/wide-results-invite.md
+++ b/.changeset/wide-results-invite.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-realtime': patch
+'@openai/agents-core': patch
+---
+
+- Add safeToRunInBrowser property to FunctionTool inteface
+- Add init option validation to detect local agents as tools in web browser to RealtimeAgent
--- a/packages/agents-core/src/agent.ts
+++ b/packages/agents-core/src/agent.ts
@@ -437,6 +437,7 @@ export class Agent<
         additionalProperties: false,
       },
       strict: true,
+      localAgentAsTool: true,
       execute: async (data, context) => {
         if (!isAgentToolInput(data)) {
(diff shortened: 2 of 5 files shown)


seratch:
i like this bot, will add a changeset

dkundel-openai:
I think the fixes of running regular Agents in the browsers should make it sufficient now to only block RealtimeAgent's to be used as tool. If a text based Agent is now used in a RealtimeAgent as a tool it should appropriately throw an error that the API key is not set for the model

seratch:
IIUC, for text agents, if a developer calls `setDefaultOpenAIKey()` on the browser side (client code), the underlying runner (the code in the function created by `asTool()`) could try to run the agent as a tool without a server. This means the app can perform responses api calls on the client side using an api key.

dkundel-openai:
Sort of! Using the set API key method alone doesn't work as it will throw an error by the OpenAI library for being in the browser.

So in order to make it work someone would actually:
1. Create their own OpenAI client instance
2. Pass the API key into that client
3. Set dangerouslyAllowApiKey: true in the OpenAI client
4. Use setDefaultOpenAIClient()

I think that's enough of a burden for people to go through. There is an example of this in #50

seratch:
Yeah, this makes sense. Actually my primary intention here was to improve the cryptic runtime error developers could face in the scenario, but now it seems the situation is better than that. I'll follow your suggestion here early next week!

seratch:
I will come up with a new PR for this when I have time for it. Let me close this now.

PR #30, opened 2025-06-05
feat: add hosted mcp tool support

yusuf-eren:


examples/hosted-mcp/approvals.ts
```diff
@@ -0,0 +1,79 @@
...
+ * approval callbacks.
+ */
+
+// Create readline interface for user input
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+function approvalCallback(toolName: string): Promise<boolean> {
```
dkundel-openai, line 15: I don't think I understand how this is currently being used

dkundel-openai (changes requested):
Thank you for contributing this! Since we have an approval flow for local function tool calls already, can you make this work with `result.interruptions`? https://openai.github.io/openai-agents-js/guides/human-in-the-loop/

This will likely require:
1. Changing the logic in `runImplementation.ts` to detect if a hosted MCP tool has requested approval and trigger an interruption with an approval item
2. Change the `approve` and `reject` functions to create the respective items for an approval.
3. Make sure that in `run` the run continues appropriately.

Adding @seratch here who had been looking into this as well.

PR #33, opened 2025-06-06
Fix #23 Add Hosted MCP server tool support

### LINKED ISSUE #23: Add Hosted MCP server tool support
### Describe the feature

While Responses API provides hosted MCP server tool support, this SDK does not yet support it yet. We're already working on this feature addition, so it will be available sometime soon!

### References:

- Tool document: https://platform.openai.com/docs/guides/tools-remote-mcp
- Python SDK implementation: https://github.com/openai/openai-agents-python/pull/731

seratch:
This pull request resolves #23

### DIFF
--- a/packages/agents-core/src/runImplementation.ts
+++ b/packages/agents-core/src/runImplementation.ts
@@ -14,7 +14,13 @@ import {
 } from './items';
 import logger, { Logger } from './logger';
 import { ModelResponse, ModelSettings } from './model';
-import { ComputerTool, FunctionTool, Tool, FunctionToolResult } from './tool';
+import {
+  ComputerTool,
+  FunctionTool,
+  Tool,
+  FunctionToolResult,
+  HostedMCPTool,
+} from './tool';
 import { AgentInputItem, UnknownContext } from './types';
 import { Runner } from './run';
 import { RunContext } from './runContext';
@@ -31,6 +37,7 @@ import * as protocol from './types/protocol';
 import { Computer } from './computer';
 import { RunState } from './runState';
 import { isZodObject } from './utils';
+import * as ProviderData from './types/providerData';
 
 type ToolRunHandoff = {
   toolCall: protocol.FunctionCallItem;
@@ -47,11 +54,17 @@ type ToolRunComputer = {
   computer: ComputerTool;
 };
 
+type ToolRunMCPApprovalRequest = {
+  requestItem: RunToolApprovalItem;
+  mcpTool: HostedMCPTool;
+};
+
 export type ProcessedResponse<TContext = UnknownContext> = {
   newItems: RunItem[];
   handoffs: ToolRunHandoff[];
   functions: ToolRunFunction<TContext>[];
   computerActions: ToolRunComputer[];
+  mcpApprovalRequests: ToolRunMCPApprovalRequest[];
   toolsUsed: string[];
   hasToolsOrApprovalsToRun(): boolean;
 };
@@ -69,12 +82,19 @@ export function processModelResponse<TContext>(
   const runHandoffs: ToolRunHandoff[] = [];
   const runFunctions: ToolRunFunction<TContext>[] = [];
   const runComputerActions: ToolRunComputer[] = [];
+  const runMCPApprovalRequests: ToolRunMCPApprovalRequest[] = [];
   const toolsUsed: string[] = [];
   const handoffMap = new Map(handoffs.map((h) => [h.toolName, h]));
   const functionMap = new Map(
     tools.filter((t) => t.type === 'function').map((t) => [t.name, t]),
   );
   const computerTool = tools.find((t) => t.type === 'computer');
+  const mcpToolMap = new Map(
+    tools
+      .filter((t) => t.type === 'hosted_tool' && t.providerData?.type === 'mcp')
+      .map((t) => t as HostedMCPTool)
+      .map((t) => [t.providerData.serverLabel, t]),
+  );
 
   for (const output of modelResponse.output) {
     if (output.type === 'message') {
@@ -82,8 +102,58 @@ export function processModelResponse<TContext>(
         items.push(new RunMessageOutputItem(output, agent));
       }
     } else if (output.type === 'hosted_tool_call') {
-      items.push(new RunToolCallItem(output, agent));
-      toolsUsed.push(output.name);
+      if (
+        output.providerData?.type === 'mcp_approval_request' ||
+        output.name === 'mcp_approval_request'
+      ) {
+        // Hosted remote MCP server support
+        const providerData =
+          output.providerData as ProviderData.HostedMCPApprovalRequest;
+        const mcpServerLabel = providerData.serverLabel;
+        const mcpServerTool = mcpToolMap.get(mcpServerLabel);
+        if (mcpServerTool !== undefined) {
+          const toolName = JSON.stringify({
+            server: providerData.serverLabel,
+            name: providerData.name,
+          });
+          // Do this approval later
+          runMCPApprovalRequests.push({
+            requestItem: new RunToolApprovalItem(
+              {
+                type: 'function_call',
+                name: toolName,
+                callId: providerData.id,
+                arguments: providerData.arguments || '',
+                status: 'in_progress',
+                providerData,
+              },
+              agent,
+            ),
+            mcpTool: mcpServerTool,
+          });
+          items.push(new RunToolCallItem(output, agent));
+          toolsUsed.push(toolName);
+        } else {
+          const message = `MCP server (${mcpServerLabel}) not found in Agent (${agent.name})`;
+          addErrorToCurrentSpan({
+            message,
+            data: { mcp_server_label: mcpServerLabel },
+          });
+          throw new ModelBehaviorError(message);
+        }
+      } else {
+        // the rest of the hosted
+        items.push(new RunToolCallItem(output, agent));
+        const toolName = output.providerData?.serverLabel
+          ? // hosted MCP tool
+            JSON.stringify({
+              server: output.providerData.serverLabel,
+              name: output.name,
+            })
+          : // other hosted tools
+            output.name;
+        toolsUsed.push(toolName);
+      }
     } else if (output.type === 'reasoning') {
       items.push(new RunReasoningItem(output, agent));
     } else if (output.type === 'computer_call') {
@@ -147,6 +217,7 @@ export function processModelResponse<TContext>(
     handoffs: runHandoffs,
     functions: runFunctions,
     computerActions: runComputerActions,
+    mcpApprovalRequests: runMCPApprovalRequests,
     toolsUsed: toolsUsed,
     hasToolsOrApprovalsToRun(): boolean {
       return (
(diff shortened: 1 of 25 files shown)


packages/agents-core/src/runImplementation.ts
```diff
@@ -68,21 +81,78 @@ export function processModelResponse<TContext>(
...
+                providerData,
+              },
+              agent,
+            ),
+            mcpTool: mcpServerTool,
+          });
+          items.push(new RunToolCallItem(output, agent));
+          toolsUsed.push(toolName);
+        } else {
+          const message = `MCP server (${mcpServerLabel}) not found in Agent (${agent.name})`;
```
seratch, line 120: This pattern usually never arises, but I added this to align with Python SDK.

packages/agents-core/src/runImplementation.ts
```diff
@@ -68,21 +81,78 @@ export function processModelResponse<TContext>(
...
+        output.providerData?.type === 'mcp_approval_request' ||
+        output.name === 'mcp_approval_request'
+      ) {
+        // Hosted remote MCP server support
+        const providerData =
+          output.providerData as ProviderData.HostedMCPApprovalRequest;
+        const mcpServerLabel = providerData.serverLabel;
+        const mcpServerTool = mcpToolMap.get(mcpServerLabel);
+        if (mcpServerTool !== undefined) {
+          const toolName = JSON.stringify({
```
seratch: Having both server label and tool name is helpful for MCP server usage, so I've included both in JSON format.

packages/agents-core/src/runImplementation.ts
```diff
@@ -336,6 +407,40 @@ export async function executeToolsAndSideEffects<TContext>(
...
+  if (processedResponse.mcpApprovalRequests.length > 0) {
+    for (const approvalRequest of processedResponse.mcpApprovalRequests) {
+      const toolData = approvalRequest.mcpTool
+        .providerData as ProviderData.HostedMCPTool<TContext>;
+      if (!toolData.onApproval) {
+        throw new UserError(
+          `Hosted remote MCP server tool (${toolData.serverLabel}) does not have an onApproval function`,
+        );
+      }
+      const approvalResult = await toolData.onApproval(
```
seratch: This is the same design with Python SDK. Developers can pass onApproval callback when passing HostedMCPTool to Agent.

packages/agents-core/src/tool.ts
```diff
@@ -110,6 +111,61 @@ export function computerTool(
...
+          serverLabel: options.serverLabel,
+          serverUrl: options.serverUrl,
+          requireApproval: 'always',
+          onApproval: options.onApproval,
+        }
+      : {
+          type: 'mcp',
+          serverLabel: options.serverLabel,
+          serverUrl: options.serverUrl,
+          requireApproval: 'never',
```
seratch: lack of this property results in 400 bad request error, so explicitly having this

packages/agents-core/src/types/providerData.ts
```diff
@@ -0,0 +1,52 @@
+import { HostedMCPApprovalFunction } from '../tool';
```
seratch, line 1: I've added these types to make the internal code changes easier. Handling Record<string, any> everywhere could be sometimes challenging to detect mistakes, so using these types internally and within agents-openai package too.

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -33,13 +33,18 @@ import {
   ImageGenerationStatus,
   WebSearchStatus,
 } from './tools';
+import { camelToSnakeCase, snakeToCamelCase } from './utils/providerData';
+import { ProviderData } from '@openai/agents-core/types';
```
seratch, line 37: importing the types from agents-core.

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -142,7 +147,12 @@ function converTool<_TContext = unknown>(
       return {
         tool: {
           type: 'file_search',
-          vector_store_ids: tool.providerData.vectorStoreId,
+          vector_store_ids:
+            tool.providerData.vectorStoreIds ||
+            // for backwards compatibility
+            (typeof tool.providerData.vectorStoreId === 'string'
```
seratch: we already have vectorStoreId here but it should be vectorStoreIds (plural). allowing both here for backward compatibility and simple use cases w/ a single vector store.

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -198,7 +218,7 @@ function getInputMessageContent(
     return {
       type: 'input_text',
       text: entry.text,
-      ...entry.providerData,
+      ...camelToSnakeCase(entry.providerData),
```
seratch: currently there is no issue but automating the conversion should be helpful for us as maintainers

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -450,12 +470,71 @@ function getInputItems(
...
+        return entry;
+      } else if (
+        item.providerData?.type === 'mcp_approval_request' ||
+        item.name === 'mcp_approval_request'
+      ) {
+        const providerData =
+          item.providerData as ProviderData.HostedMCPApprovalRequest;
+        const entry: OpenAI.Responses.ResponseInputItem.McpApprovalRequest = {
+          type: 'mcp_approval_request',
+          id: providerData.id ?? item.id!,
```
seratch, line 532: having most of the properties in providerData is easier to understand for maintainers on the agents-core side plus the data size is ignorable, so intentionally duplicating the data

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -512,86 +591,117 @@ function convertToOutputItem(
...
+      const { status, ...remainingItem } = item;
+      let outputData = undefined;
+      if ('result' in remainingItem && remainingItem.result !== null) {
+        // type: "image_generation_call"
+        outputData = remainingItem.result;
+        delete (remainingItem as any).result;
+      }
       const output: protocol.HostedToolCallItem = {
         type: 'hosted_tool_call',
-        id,
```
seratch, line 542: The id property could be used in many cases, so i would like to consistently have it also in providerData.

packages/agents-openai/src/tools.ts
```diff
@@ -188,3 +188,41 @@ export function imageGenerationTool(
...
+ * A hosted MCP tool that lets the model call a remote MCP server directly
+ * without a round trip back to your code.
+ */
+export type HostedMCPTool = HostedTool & {
+  name: 'hosted_mcp';
+  providerData: {
+    type: 'mcp';
+    serverLabel: string;
+    serverUrl: string;
+    requireApproval?: 'never' | 'once' | 'always';
```
seratch:
```suggestion
    requireApproval?: 'never' | 'always';
```

packages/agents-openai/src/tools.ts
```diff
@@ -188,3 +188,41 @@ export function imageGenerationTool(
...
+ * Creates a hosted MCP tool definition.
+ *
+ * @param serverLabel - The label identifying the MCP server.
+ * @param serverUrl - The URL of the MCP server.
+ * @param requireApproval - Whether tool calls require approval.
+ */
+export function hostedMcpTool(options: {
+  serverLabel: string;
+  serverUrl: string;
+  requireApproval?: 'never' | 'once' | 'always';
```
seratch:
```suggestion
  requireApproval?: 'never' | 'always';
```

packages/agents-core/src/runImplementation.ts
dkundel-openai: I think the logic here should be that `onApproval` is optional. If it is passed then we execute that to determine approval. If it is not passed we should treat this using interruptions meaning a user will have to call `result.state.approve(interruption)` or `result.state.reject(interruption)`. Essentially the approval code experience should be possible to be the same if someone mixes hosted MCP and regular function tool calls that require approval

packages/agents-core/src/runState.ts
```diff
@@ -152,6 +152,12 @@ const serializedProcessedResponseSchema = z.object({
       computer: z.any(),
     }),
   ),
+  mcpApprovalRequests: z.array(
+    z.object({
+      requestItem: z.any(),
+      mcpTool: z.any(),
+    }),
+  ),
```
dkundel-openai: If we are adding anything to the runState schema we need to make it optional / have default values for backwards compatibility

packages/agents-core/test/runState.test.ts
```diff
@@ -252,6 +252,7 @@ describe('deserialize helpers', () => {
       functions: [],
       handoffs: [],
       computerActions: [{ toolCall: call, computer: tool }],
+      mcpApprovalRequests: [],
```
dkundel-openai, line 255: we should have test that handle the absence of this for backwards compatibility.

packages/agents-openai/src/openaiResponsesModel.ts
dkundel-openai: I had considered this. The main reason I shied away from this was that my original thinking was that `providerData` should be neutral but I guess since we are already reading so many fields from it, this might make sense

packages/agents-openai/src/openaiResponsesModel.ts
dkundel-openai: provider data should always support any properties in my mind. this is essentially the equivalent of providerMetadata in the Vercel AI SDK. But I guess inside the model I'm okay with doing additional processing as long as we continue to allow adding additional properties so that if the API adds them we don't have to make changes to support them

packages/agents-core/src/runState.ts
```diff
@@ -152,6 +153,28 @@ const serializedProcessedResponseSchema = z.object({
...
+        }),
+        // HostedMCPTool
+        mcpTool: z.object({
+          type: z.literal('hosted_tool'),
+          name: z.literal('hosted_mcp'),
+          providerData: z.record(z.string(), z.any()),
+        }),
+      }),
+    )
+    .optional(),
```
seratch, line 177: this is optional for backward compatibility

packages/agents-core/src/runImplementation.ts
seratch: Now this PR supports both HITL and onApproval callback

seratch:
@dkundel-openai Updated this PR based on the previous review comments. It should be ready for review again

packages/agents-openai/src/openaiResponsesModel.ts
seratch: @dkundel-openai As we discussed, I've updated the internal code to consistently use snake_case keys for providerData and it seems working very well: https://github.com/openai/openai-agents-js/pull/33/commits/9779f1c5cb386caaf11f9b9effbebcd67592dc1c

packages/agents-core/test/run.test.ts
```diff
@@ -88,7 +88,7 @@ describe('Runner.run', () => {
 
       const rawItem = {
         name: 'toolZ',
-        call_id: 'c1',
+        callId: 'c1',
```
dkundel-openai, line 91: was this just broken or did this change somewhere? I didn't see the related change anywhere else

packages/agents-core/src/runImplementation.ts
```diff
@@ -236,20 +302,18 @@ export async function executeInterruptedToolsAndSideEffects<TContext>(
...
-    .map((item) => {
-      return item.rawItem.callId;
-    });
-
+  // call_ids for function tools
+  const functionCallIds = originalPreStepItems
+    .filter(
+      (item) =>
+        item instanceof RunToolApprovalItem &&
+        'callId' in item.rawItem &&
```
seratch, line 310: @dkundel-openai this is the change that is not compatible with run.test.ts's call_id; do you think both callId and call_id should be supported? I don't think so though.

packages/agents-core/test/run.test.ts
seratch: this is associated with this change. i don't have any strong opinion about the additional validation but it should be good to go: https://github.com/openai/openai-agents-js/pull/33/files#r2139160148

dkundel-openai: ah this was actually broken. It's always been callId for a `function_call` but because in this test it was typed as `any` it didn't cause trouble

dkundel-openai (approved):

PR #37, opened 2025-06-06
fix(agents-core): ensure handoff messages use valid JSON

whoiskatrin:
getTransferMessage isn't returning a valid JSON, added a test to verify the output
added changeset


packages/agents-core/src/handoff.ts
```diff
@@ -47,7 +47,7 @@ export type HandoffInputFilter = (input: HandoffInputData) => HandoffInputData;
 export function getTransferMessage<TContext, TOutput extends AgentOutputType>(
   agent: Agent<TContext, TOutput>,
 ) {
-  return `{'assistant': '${agent.name}'}`;
```
seratch, line 50: The only reason why we went with this string is to align with Python SDK. As long as the standard JSON format works without any issues here, I think it should be fine to change this. @dkundel-openai thoughts?

jhills20: fine to change

PR #40, opened 2025-06-06
fix(shims): Bind crypto to randomUUID

AielloChan:
Fix a bug when call `randomUUID` in the browser.

### DIFF
--- a/.changeset/quiet-flies-march.md
+++ b/.changeset/quiet-flies-march.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


seratch (approved):
Thank you! You're right that this fix is necessary for web browser environment.

seratch:
@dkundel-openai will review this probably tomorrow but before that, could you run `pnpm changeset` to generate changeset file? This one brings changes to agents-core package. Also, we're currently using patch version bumps.

dkundel-openai (approved):

PR #41, opened 2025-06-06
fix(shims): Naively polyfill AsyncLocalStorage in browser

AielloChan:
Some code like `packages/agents-core/src/tracing/context.ts:25` will call this function.


seratch (commented):
Thank you. We actually recognized this and was one of TODOs we haven't moved to the issue tracker.

Issue #16, opened 2025-06-04
tracing/`defaultProcessor()` breaks Cloudflare Workers runtime


threepointone:
### Describe the bug

Cloudflare Workers has a constraint where you can't start async i/o outside of the context of a request (so you can't do a setTimeout/fetch etc until a request comes in). We do this for performance and security reasons. 

OpenAI agents starts collecting traces over here https://github.com/openai/openai-agents-js/blob/b9a655420807a24cfbeb83508bed5996533c9842/packages/agents-core/src/index.ts#L150 as soon as the module is loaded. This makes Workers throw an error, which looks like this: 

```
Uncaught Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or
  connect()), setting a timeout, and generating random values are not allowed within global scope.
  To fix this error, perform this operation within a handler.
  https://developers.cloudflare.com/workers/runtime-apis/handlers/
    at null.<anonymous> (server.js:21280:29) in <instance_members_initializer>
    at null.<anonymous> (server.js:21281:14) in BatchTraceProcessor
    at null.<anonymous> (server.js:21428:25) in defaultProcessor
    at null.<anonymous> (server.js:28636:19)
   [code: 10021]
```

(Annoyingly, while this error does show in `wrangler dev` and in production, it doesn't show with the vite plugin, which is why I missed this until it was reported to me)

So a fix could be that we don't start the batch processor if `navigator.userAgent === 'Cloudflare-Workers`, and it's up to the developer to add trace collection manually. 

Filing this issue to discuss before I start thinking of solutions / sending PRs.  

### Debug information

- Agents SDK version: `v0.0.2`
- Runtime environment `workerd@1.20250604.0`

### Repro steps

https://github.com/RogutKuba/test-openai-sdk `pnpm install && pnpm start`

### Expected behavior

It, um, shouldn't crash.

dkundel-openai:
That makes sense. Since this is a cloudflare workers specific issue, I wonder if adding a flag that is turned off in the `shims-workerd.ts` and potentially in browser as well makes more sense. That way we are not adding a userAgent flag runtime check into the processor part. We already turn it off in the browser here https://github.com/openai/openai-agents-js/blob/main/packages/agents-core/src/tracing/processor.ts#L128 using that same approach.

The only thing with this is that it might run into the same bug as #10

Manouchehri:
I really do want automatic trace collecting in Workers with this SDK though. :( 

[...]

Issue #10, opened 2025-06-04
TypeError: Illegal invocation in browser when using run() with tracing, despite setTracingDisabled(true)


chnl:
Issue Description:
When using `@openai/agents@0.0.1` in a Next.js (v15.2.4, React 19) application, a TypeError: Illegal invocation occurs. This error originates from within the SDK's internal tracing mechanisms when run(agent, ...) is called, specifically in what appears to be utils.mjs according to the stack trace. This happens even when setTracingDisabled(true) is called before any agent operations.

The application uses the standard `@openai/agents` package (not `@openai/agents-realtime`). The error seems to indicate that a trace-related function is being called with an incorrect this context in the browser environment.
```
SDK Version:
@openai/agents": "^0.0.1" (from package.json)
Environment:
Next.js v15.2.4
React v19.0.0
Browser (e.g., Chrome, latest)
Node.js (for build, likely v20+ as per SDK recommendations)
```

Steps to Reproduce:

- Set up a Next.js client component ("use client";).
- Import Agent, run, setDefaultOpenAIClient, setTracingDisabled from `@openai/agents`.
- Import OpenAI from openai.

In the component/hook:

- Call setTracingDisabled(true) as early as possible.
- Initialize an OpenAI client instance.
- Call setDefaultOpenAIClient() with the created client instance.
- Define a simple Agent instance (e.g., new Agent({ name: 'TestAgent', instructions: 'You are a test agent.' })).
- Attempt to call await run(agentInstance, "Test input").

Observed Behavior:
```
A TypeError: Illegal invocation is thrown. The browser console also shows messages like:
BatchTraceProcessor is not supported in the browser. Traces will not be exported.

Stack Trace:
[Orchestrator] Error running WelcomeAgent: TypeError: Illegal invocation
    at tX (utils.mjs:15:21) // Mapped to https://in.ference.ai/_next/static/chunks/682-a6a0afe59dc77b8b.js:42:17358
    at new tq (traces.mjs:12:43) // Mapped to https://in.ference.ai/_next/static/chunks/682-a6a0afe59dc77b8b.js:42:18893
    at new tY (traces.mjs:54:1) // Mapped to https://in.ference.ai/_next/static/chunks/682-a6a0afe59dc77b8b.js:42:19447
    at tV.createTrace (provider.mjs:52:1) // Mapped to https://in.ference.ai/_next/static/chunks/682-a6a0afe59dc77b8b.js:42:19867
[...]

PR #50, opened 2025-06-06
Fix browser and cloudflare trace exports

dkundel-openai:
Both Cloudflare and the Browser should not run export loops by default.

This PR:
1. Properly disables the export loop in Cloudflare Workers by default. In the browser it was already disabled.
2. Provides a way to start the export loop with `startTraceExportLoop` and a way to `forceFlush` the traces (for Cloudflare)
3. Adds some integration tests to test if the SDK runs in the browser


dkundel-openai:
This should fix #16 @threepointone if you want to take a look

This should also further fix #10 removing the warning log

threepointone:
I'll have a look this weekend! thanks so much @dkundel-openai, legend.

integration-tests/cloudflare-workers/worker/src/index.ts
```diff
@@ -12,26 +12,41 @@
...
+      const agent = new Agent({
+        name: 'Test Agent',
+        instructions:
+          'You will always only respond with "Hello there!". Not more not less.',
+      });
+      const result = await run(agent, 'Hey there!');
 
-			setDefaultOpenAIKey(env.OPENAI_API_KEY!);
-			setTracingDisabled(true);
+      // make sure you shut down tracing before exiting to flush any remaining traces
```
threepointone: a problem here is that a worker isolate might be serving multiple requests, so this might shutdown before other requests are complete. I'll sit down this evening after dinner and try to come up with something better (probably using ALS)

threepointone: another issue is that if this returns a websocket, then the response is returned early, and any traces collected during the websocket's lifetime won't be flushed

threepointone:
Ok, I think I have a clean solution without mucking about ALS etc. tldr - 

- let's not run `startTraceExportLoop()`: this has the tendency to keep the worker running because there will always be a running timer
- events start getting flushed anyway when the batch size is hit (because of `BatchProcessor::#safeAddItem`
- cloudflare users will be expected to call `forceFlush()` manually (added as a method to `TraceProvider`
- we (cloudflare/me) can probably have a better integration with out own Agent class that sets this up automatically / removes some boilerplate. 

```diff
diff --git a/integration-tests/cloudflare-workers/worker/src/index.ts b/integration-tests/cloudflare-workers/worker/src/index.ts
index c1d5c72..c132ab8 100644
--- a/integration-tests/cloudflare-workers/worker/src/index.ts
+++ b/integration-tests/cloudflare-workers/worker/src/index.ts
@@ -31,7 +31,6 @@ export default {
     try {
       setDefaultOpenAIKey(env.OPENAI_API_KEY!);
       setTraceProcessors([new BatchTraceProcessor(new ConsoleSpanExporter())]);
-      startTraceExportLoop();
 
       const agent = new Agent({
         name: 'Test Agent',
@@ -40,13 +39,13 @@ export default {
       });
       const result = await run(agent, 'Hey there!');
 
-      // make sure you shut down tracing before exiting to flush any remaining traces
-      await getGlobalTraceProvider().shutdown();
-
       return new Response(`[RESPONSE]${result.finalOutput}[/RESPONSE]`);
     } catch (error) {
       console.error(error);
       return new Response(String(error), { status: 500 });
+    } finally {
+      // make sure to flush any remaining traces before exiting
+      ctx.waitUntil(getGlobalTraceProvider().forceFlush());
     }
   },
 } satisfies ExportedHandler<Env>;
diff --git a/packages/agents-core/src/tracing/provider.ts b/packages/agents-core/src/tracing/provider.ts
index ecbea0a..1b01c0b 100644
--- a/packages/agents-core/src/tracing/provider.ts
+++ b/packages/agents-core/src/tracing/provider.ts
@@ -204,6 +204,10 @@ export class TraceProvider {
       });
     }
   }
+
+  async forceFlush(): Promise<void> {
+    await this.#multiProcessor.forceFlush();
+  }
 }
 
 let GLOBAL_TRACE_PROVIDER: TraceProvider | undefined = undefined;
```

rm-openai (approved):

PR #51, opened 2025-06-07
docs: redesign to fit more new centralized design

dkundel-openai:
This is not implementing the new design entirely and is instead a first iteration


seratch (approved):
Thanks!

PR #57, opened 2025-06-08
fix(agents-core): prevent premature agent termination with text and function calls

DroopyTersen:
## Description
This PR fixes an agent termination issue in the `@openai/agents-core` package:

**Agent Loop Continuation**: Prevents premature agent termination when models return both text output and tool calls

## Problem
Some models (Claude Sonnet 4) return both text and tool calls in their responses. Previously, if there was any potential final output (text), the agent would terminate even if there were pending function results or computer results to process. This caused the agent to stop prematurely before completing all requested tool operations.

## Solution
- Modified the condition to continue the agent loop when there are function results or computer results present, regardless of text output
- Added checks for `functionResults.length > 0` and `computerResults?.length > 0` to the termination condition

## Testing
- [x] Built project successfully with `pnpm build` (had to futz with some types to get `build` working after initial fork, not sure if it was a setup issue on my side)
- [x] All tests pass with `CI=1 pnpm test`
- [x] Linting passes with `pnpm lint`
- [x] Verified fix prevents premature agent termination when models return both text and tool calls

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/.changeset/modern-bananas-greet.md
+++ b/.changeset/modern-bananas-greet.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Prevent premature agent termination when models return both text and tool calls
--- a/packages/agents-core/src/runImplementation.ts
+++ b/packages/agents-core/src/runImplementation.ts
@@ -404,7 +404,13 @@ export async function executeToolsAndSideEffects<TContext>(
       : undefined;
 
   // if there is no output we just run again
-  if (!potentialFinalOutput) {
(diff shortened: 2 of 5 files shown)


dkundel-openai:
Thank you for your contribution @DroopyTersen! Looking at this and #62 I'm going to be looking at merging that one to be closer to the Python implementation but we appreciate your contribution!

PR #68, opened 2025-06-10
fix(realtime): add zod dependency to package.json. fixes #52

### LINKED ISSUE #52: Deno compatibility issue with @openai/agents package
## Issue Description

The @openai/agents package fails to run in Deno with the following error:

```
error: Could not resolve 'npm:@openai/agents@0.0.4'

Caused by:
    [ERR_MODULE_NOT_FOUND] Cannot find module 'file:///Users/michael/Library/Caches/deno/npm/registry.npmjs.org/@openai/agents/0.0.4_1/index.js' imported from 'file:///Users/michael/Developer/deno-openai-agents/mod.ts'
```

## Steps to Reproduce

1. Create a `mod.ts` file with:
```typescript
import { Agent, run } from 'npm:@openai/agents';

const agent = new Agent({
  name: 'Assistant',
});

const result = await run(
  agent,
  'Write a haiku about recursion in programming.',
);
```

2. Create a `deno.json` file with:
```json
{
  "tasks": {
    "dev": "deno run --allow-net --allow-env mod.ts"
  }
}
```

3. Run `deno task dev`

## Expected Behavior

The package should work with Deno as mentioned in the documentation that Deno is a supported platform.

## Environment

- Deno version: deno 2.3.5 (stable, release, aarch64-apple-darwin)
- @openai/agents version: 0.0.4
- Platform: macOS

## Additional Context

Despite Deno being listed as a supported platform, the package fails to resolve properly when imported via npm: specifier.

dkundel-openai:


verdaccio-config.yml
```diff
@@ -9,7 +9,6 @@ packages:
   '@openai/*':
     access: $all
     publish: $all
-    proxy: npmjs
```
dkundel-openai, line 12: This is just a side change since Verdaccio blocks publishing the same version that's already public if this is set

seratch (approved):
Thanks!

PR #69, opened 2025-06-10
chore: update version details incl. linking versions

dkundel-openai:


.changeset/config.json
```diff
@@ -3,7 +3,7 @@
   "changelog": "@changesets/cli/changelog",
   "commit": false,
   "fixed": [],
-  "linked": [],
+  "linked": [["@openai/agents-*"]],
```
dkundel-openai, line 6: This should properly link the versions now

PR #80, opened 2025-06-11
Add remote MCP server (Streamable HTTP) support; fixes #78

### LINKED ISSUE #78: Connect with MCP server running on local
### Please read this first

Yes
- **Have you searched for related issues?** Others may have had similar requests
Yes

### Describe the feature
I am currently have a MCPServer created with model context protocol library and it running locally. In the same express app I want to create a new endpoint to work with multi-agents architecture but currently, we don't support that. Since the MCP tools from what I read only work with stdio MCPServer? 

```typescript
  const mcpServer = new MCPServerStdio({
    name: 'Filesystem Server, via npx',
    fullCommand: `npx -y @modelcontextprotocol/server-filesystem ${samplesDir}`,
    // Or passing command and args
    // command: 'npx',
    // args: ['-y', '@modelcontextprotocol/server-filesystem', samplesDir],
  });
```

And you kind of have to do it like above which doesn't make sense. 

We should just be able to pass in server name and the local url or url to work with and the library should automatically discover the list of mcp tools to use. 

Or we do it similar to what Vercel AI Sdk does and pass it to the model creation call.

```typescript
  const mcpClient = await experimental_createMCPClient({
    transport: mcpTransportInstance,
    onUncaughtError,
  });

  const tools = await mcpClient.tools();
```
 
What is the feature you're requesting? How would it work? Please provide examples and details if possible.

seratch:
This pull request adds remote MCP server support (Streamable HTTP) to the `mcpServers` parameter of the Agent constructor. Having this change resolves #78. We don't plan to support SSE, which is now a legacy protocol.

### DIFF
--- a/.changeset/crazy-coats-give.md
+++ b/.changeset/crazy-coats-give.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Add remote MCP server (Streamable HTTP) support
--- a/examples/mcp/package.json
+++ b/examples/mcp/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "build-check": "tsc --noEmit",
     "start:stdio": "tsx filesystem-example.ts",
+    "start:streamable-http": "tsx streamable-http-example.ts",
     "start:hosted-mcp-on-approval": "tsx hosted-mcp-on-approval.ts",
     "start:hosted-mcp-human-in-the-loop": "tsx hosted-mcp-human-in-the-loop.ts",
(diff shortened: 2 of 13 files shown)


dkundel-openai:
Awesome! Can we add a guide to the docs that explains the three different options for MCP and when to use what. Like how the remote MCP hosted tool only works for Responses models but should be preferred for hosted MCP servers while the other options are good for local MCP use cases or in combination with other models

seratch:
Yeah, I will work on the document too!

PR #85, opened 2025-06-11
Export OpenAI Response object on ResponseSpanData.

tcdent:
Adds `_response` as an attribute to `ResponseSpanData` for use by third-party tracing integrations. 

Adds a test to verify that the new attribute does not get exported to the default tracer.

### DIFF
--- a/packages/agents-core/src/tracing/spans.ts
+++ b/packages/agents-core/src/tracing/spans.ts
@@ -1,3 +1,4 @@
+import OpenAI from 'openai';
 import logger from '../logger';
 import { TracingProcessor } from './processor';
 import { generateSpanId, removePrivateFields, timeIso } from './utils';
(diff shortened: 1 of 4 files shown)


packages/agents-core/src/tracing/spans.ts
```diff
@@ -38,6 +39,7 @@ export type ResponseSpanData = SpanDataBase & {
    * Not used by the OpenAI tracing provider but helpful for other tracing providers.
    */
   _input?: string | Record<string, any>[];
+  _response?: OpenAI.Responses.Response;
```
dkundel-openai: Can we type this as Record<string, any>? We are trying to remove as much of the type reliance in core on OpenAI types as possible

tcdent: No problem. Done!

dkundel-openai (approved):

PR #90, opened 2025-06-12
Run pnpm self-update & pnpm upgrade

seratch:
This pull upgrades pnpm and projects' devDependencies


vitest.config.ts
```diff
@@ -2,7 +2,7 @@ import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
   test: {
-    workspace: ['packages/*'],
+    projects: ['packages/*'],
```
seratch, line 5: Updated this part due to the following vitest deprecation:

>DEPRECATED  The `workspace` option is deprecated and will be removed in the next major. To hide this warning, rename `workspace` option to `projects`.

dkundel-openai (approved):

PR #96, opened 2025-06-12
Improve the types of RealtimeSession configuration

seratch:
- inputAudioTranscription
- turnDetection

### DIFF
--- a/examples/docs/voice-agents/turnDetection.ts
+++ b/examples/docs/voice-agents/turnDetection.ts
@@ -7,8 +7,8 @@ const session = new RealtimeSession(agent, {
     turnDetection: {
       type: 'semantic_vad',
       eagerness: 'medium',
-      create_response: true,
-      interrupt_response: true,
+      createResponse: true,
+      interruptResponse: true,
     },
   },
 });
--- a/packages/agents-realtime/src/clientMessages.ts
+++ b/packages/agents-realtime/src/clientMessages.ts
@@ -33,6 +33,37 @@ export type RealtimeTracingConfig =
     }
   | 'auto';
 
+export type RealtimeInputAudioTranscriptionConfig = {
+  language?: string;
+  model?: 'gpt-4o-transcribe' | 'gpt-4o-mini-transcribe' | 'whisper-1' | string;
+  prompt?: string;
+};
+
+export type RealtimeTurnDetectionConfigAsIs = {
+  type?: 'semantic_vad';
+  create_response?: boolean;
+  eagerness?: 'auto' | 'low' | 'medium' | 'high';
+  interrupt_response?: boolean;
+  prefix_padding_ms?: number;
+  silence_duration_ms?: number;
+  threshold?: number;
+};
+
+// The Realtime API accepts snake_cased keys, so when using this, this SDK coverts the keys to snake_case ones before passing it to the API
+export type RealtimeTurnDetectionConfigCamelCase = {
+  type?: 'semantic_vad';
+  createResponse?: boolean;
+  eagerness?: 'auto' | 'low' | 'medium' | 'high';
+  interruptResponse?: boolean;
+  prefixPaddingMs?: number;
+  silenceDurationMs?: number;
+  threshold?: number;
+};
+
+export type RealtimeTurnDetectionConfig =
+  | RealtimeTurnDetectionConfigAsIs
+  | RealtimeTurnDetectionConfigCamelCase;
+
 export type RealtimeSessionConfig = {
   model: string;
   instructions: string;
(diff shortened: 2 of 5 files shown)


examples/docs/voice-agents/turnDetection.ts
```diff
@@ -7,8 +7,8 @@ const session = new RealtimeSession(agent, {
     turnDetection: {
       type: 'semantic_vad',
       eagerness: 'medium',
-      create_response: true,
```
seratch, line 10: changed the example but either works!

packages/agents-realtime/src/clientMessages.ts
```diff
@@ -33,15 +33,46 @@ export type RealtimeTracingConfig =
     }
   | 'auto';
 
+export type RealtimeInputAudioTranscriptionConfig = {
+  language?: string;
+  model?: 'gpt-4o-transcribe' | 'gpt-4o-mini-transcribe' | 'whisper-1' | string;
```
seratch: allowed this property to pass anything else as we may release new models in the future (plus alpha users may use different ones)

packages/agents-realtime/src/openaiRealtimeBase.ts
```diff
@@ -406,6 +408,43 @@ export abstract class OpenAIRealtimeBase
...
+  private static buildTurnDetectionConfig(
+    c: RealtimeTurnDetectionConfig | undefined,
+  ): RealtimeTurnDetectionConfigAsIs | undefined {
+    if (typeof c === 'undefined') {
+      return undefined;
+    }
+    return {
+      type: c.type,
+      create_response:
+        'createResponse' in c
```
seratch: if snake_case then camelCase is preferred, happy to change this order

packages/agents-realtime/src/clientMessages.ts
dkundel-openai: We should do `string & {}` for better type autocomplete

packages/agents-realtime/src/clientMessages.ts
```diff
@@ -33,15 +33,46 @@ export type RealtimeTracingConfig =
...
+  eagerness?: 'auto' | 'low' | 'medium' | 'high';
+  interrupt_response?: boolean;
+  prefix_padding_ms?: number;
+  silence_duration_ms?: number;
+  threshold?: number;
+};
+
+// The Realtime API accepts snake_cased keys, so when using this, this SDK coverts the keys to snake_case ones before passing it to the API
+export type RealtimeTurnDetectionConfigCamelCase = {
+  type?: 'semantic_vad';
```
dkundel-openai: Type can also be server_vad so we should support both.

packages/agents-realtime/src/clientMessages.ts
```diff
@@ -33,15 +33,46 @@ export type RealtimeTracingConfig =
...
+  createResponse?: boolean;
+  eagerness?: 'auto' | 'low' | 'medium' | 'high';
+  interruptResponse?: boolean;
+  prefixPaddingMs?: number;
+  silenceDurationMs?: number;
+  threshold?: number;
+};
```
dkundel-openai, line 65: Can we make it acceptable for this to also still take other properties inside of these two settings? Thinking how theoretically you could roll your own Realtime Transport Layer right now with other session config. But also fine to guide people to providerData for that and override this entire property

packages/agents-realtime/src/clientMessages.ts
seratch: @dkundel-openai thanks, updated!

dkundel-openai (changes requested):
See open comments

packages/agents-realtime/src/clientMessages.ts
seratch: thanks, updated

packages/agents-realtime/src/openaiRealtimeBase.ts
```diff
@@ -406,6 +408,48 @@ export abstract class OpenAIRealtimeBase
...
+        ? interruptResponse
+        : interrupt_response,
+      prefix_padding_ms: prefixPaddingMs ? prefixPaddingMs : prefix_padding_ms,
+      silence_duration_ms: silenceDurationMs
+        ? silenceDurationMs
+        : silence_duration_ms,
+      threshold,
+      ...rest,
+    };
+    // Remove undefined values from the config
```
seratch, line 446: When I verified the behavior, having undefined values could affect the connection establishment, so I added this logic. but if my observation is wrong or is missing something, please feel free to adjust this part.

packages/agents-realtime/src/realtimeSession.ts
```diff
@@ -519,7 +519,7 @@ export class RealtimeSession<
     this.#transport.on('turn_done', (event) => {
       const item = event.response.output[event.response.output.length - 1];
       const textOutput = getLastTextFromAudioOutputMessage(item) ?? '';
-      const itemId = item.id ?? '';
+      const itemId = item?.id ?? '';
```
seratch, line 522: This is an unrelated existing bug i found while doing tests

dkundel-openai (approved):

PR #105, opened 2025-06-13
Add guardrail itemId details

dkundel-openai:
## Summary
- include itemId in `guardrail_tripped` event details
- handle itemId when running output guardrails
- assert itemId in guardrail test
- document new details argument (English & Japanese)
- use `item.id` when capturing guardrail-triggering output

## Testing
- `pnpm lint`
- `pnpm -r build-check` *(fails to resolve `@openai/agents-core`)*
- `CI=1 pnpm test` *(fails to load `@openai/agents-core/_shims`)*
- `pnpm build`

------
https://chatgpt.com/codex/tasks/task_i_684cabbac64c833193a0e76d458db0bd


packages/agents-realtime/src/realtimeSession.ts
```diff
@@ -447,7 +447,11 @@ export class RealtimeSession<
     }
   }
 
-  async #runOutputGuardrails(output: string, responseId: string) {
+  async #runOutputGuardrails(
+    output: string,
+    responseId: string,
+    itemId: string,
```
seratch, line 453: consistently using a "details" object here would be easier to enhance, but since this is a private method, we can change it if necessary in the future too.

PR #121, opened 2025-06-18
Fix #117 groupId, traceMetadata passed to Runner are not set

### LINKED ISSUE #117: workflowName, groupId and traceId are not set
```
    const runner = new Runner({
      workflowName: 'chatbot_storefront',
      groupId: chatSession.sessionId,
      traceId: `trace_${message.id}`,
      tracingDisabled: false
    })

    const result = await runner.run(
      startAgent,
      message.message.content,
      {
        context: context,
        previousResponseId: chatSession.lastResponseId,
        stream: true
      },
    );
```
I can see traces in Traces Dashboard but their workflowName, groupId, and traceId are not set as defined. 

Am I missing something or is it a bug?

seratch:
This pull request resolves #117


packages/agents-core/src/run.ts
```diff
@@ -127,7 +127,7 @@ export type RunConfig = {
   /**
    * An optional dictionary of additional metadata to include with the trace.
    */
-  traceMetadata?: Record<string, any>;
+  traceMetadata?: Record<string, string>;
```
seratch, line 130: Confirmed the server side behavior and found that only string values are allowed here. Also, it does not support nested data.

dkundel-openai (approved):

PR #125, opened 2025-06-19
Fix #124 by allowing an empty object for Usage#add method

### LINKED ISSUE #124: Custom model provider raises an error when usage object does not have some properties
我按照文档上的例子去运行Echo Model [https://openai.github.io/openai-agents-js/guides/models/#custom-model-providers](url)

但是报错了
Image

```ts
import { Agent, Runner } from "@openai/agents-core";
import { EchoModel, EchoProvider } from "./model";


const agent = new Agent({
  name: "Assistant",
  instructions: "You are a helpful assistant",
  model: new EchoModel(),
  modelSettings: { temperature: 0.7, toolChoice: "auto" },
  tools: [],
});

const runner = new Runner({ modelProvider: new EchoProvider() });

const result = await runner.run(agent, "你好，自我介绍下");
console.log('===>result', result.finalOutput);




export class EchoModel implements Model {
  name: string;
  constructor() {
    this.name = 'Echo';
  }
  async getResponse(request: ModelRequest): Promise<ModelResponse> {
    return {
      usage: {},
      output: [{ role: 'assistant', content: request.input as string }],
    } as any;
  }
  async *getStreamedResponse(
    _request: ModelRequest,
  ): AsyncIterable<ResponseStreamEvent> {
    yield {
      type: 'response.completed',
      response: { output: [], usage: {} },
    } as any;
  }
}

export class EchoProvider implements ModelProvider {
  getModel(_modelName?: string): Promise<Model> | Model {
    return new EchoModel();
  }
}
```

seratch:
This pull request resolves #124

### DIFF
--- a/packages/agents-core/src/usage.ts
+++ b/packages/agents-core/src/usage.ts
@@ -62,11 +62,11 @@ export class Usage {
     this.outputTokens += newUsage.outputTokens;
     this.totalTokens += newUsage.totalTokens;
     if (newUsage.inputTokensDetails) {
-      // The type does not allow unddefined, but it could happen runtime
+      // The type does not allow undefined, but it could happen runtime
       this.inputTokensDetails.push(...newUsage.inputTokensDetails);
     }
     if (newUsage.outputTokensDetails) {
-      // The type does not allow unddefined, but it could happen runtime
+      // The type does not allow undefined, but it could happen runtime
       this.outputTokensDetails.push(...newUsage.outputTokensDetails);
     }
   }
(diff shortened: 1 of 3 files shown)


packages/agents-core/src/usage.ts
```diff
@@ -61,8 +61,14 @@ export class Usage {
     this.inputTokens += newUsage.inputTokens;
     this.outputTokens += newUsage.outputTokens;
     this.totalTokens += newUsage.totalTokens;
-    this.inputTokensDetails.push(...newUsage.inputTokensDetails);
-    this.outputTokensDetails.push(...newUsage.outputTokensDetails);
+    if (newUsage.inputTokensDetails) {
+      // The type does not allow unddefined, but it could happen runtime
```
dkundel-openai: There is a typo in undefined both here and below

dkundel-openai (changes requested):

packages/agents-core/src/usage.ts
seratch: Thanks, resolved!

PR #145, opened 2025-06-24
Fix #138 by checking the unexpected absence of currentAgent.handoffs

### LINKED ISSUE #138: TypeError: Cannot read properties of undefined (reading 'map') when using handoffs to agent with tools
When using the handoffs feature in an agent, if the delegated agent has tools, running run(mainAgent, input) triggers a runtime error:

```
TypeError: Cannot read properties of undefined (reading 'map')
    at <anonymous> (/node_modules/@openai/agents-core/dist/run.js:104:71)
    ...
```
•	Main agent contains a handoff to a secondary agent with tools (e.g., getCurrentTime, getCurrentDate).
	•	When asking a question routed via handoff, the server crashes.

⸻

Steps to Reproduce:
	1.	Create two agents:
	•	mainAgent (with handoffs)
	•	assistantAgent (with tools)
	2.	Use the [README example](https://github.com/openai/openai-agents-js#handoffs-between-agents) but add tools to the delegated agent.
	3.	Call run(mainAgent, "What is the current time?")
	4.	See error.

⸻

Expected Behavior:
Handoff should work as documented, and the delegated agent with tools should return the answer.

⸻

Actual Behavior:
Throws TypeError: Cannot read properties of undefined (reading 'map').

⸻

Versions:
	•	@openai/agents: 0.0.9
	•	@openai/agents-core: 0.0.9
	•	Node.js: 20.19.2
	•	OS: macOS

⸻

Additional Context:
Manual agent routing works as a workaround.
Please clarify if this is a known limitation or a bug.

seratch:
This pull request resolves #138

### DIFF
--- a/.changeset/better-kids-trade.md
+++ b/.changeset/better-kids-trade.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Fix #138 by checking the unexpected absence of state.currentAgent.handoffs
--- a/.changeset/upset-baboons-shop.md
+++ b/.changeset/upset-baboons-shop.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Fix #138 by checking the unexpected absence of currentAgent.handoffs
--- a/packages/agents-core/src/run.ts
+++ b/packages/agents-core/src/run.ts
@@ -302,7 +302,12 @@ export class Runner extends RunHooks<any, AgentOutputType<unknown>> {
           }
 
           if (state._currentStep.type === 'next_step_run_again') {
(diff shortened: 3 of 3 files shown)


seratch:
This is a quick fix but there may be something further we should handle. Asking the reporter to provide more information.

dkundel-openai:
Thanks! Indeed odd behavior. Can you add a changeset just to cover it?

seratch:
Good catch; Added changeset file

seratch:
will wait for a response in the issue before merging this change

seratch:
We'll merge this change as a quick fix but we may apply more changes to support more patterns when I learn them

PR #146, opened 2025-06-24
Fix #129 The model in run config should be used over an agent's default setting

### LINKED ISSUE #129: Runner default configuration is not propagated to the Agents
According to the documentation (and the types), it should be possible to provide default values for `model` etc. using a custom `Runner` instance. The values can be set (type safe), but the Agents do not respect them and fall back to their own default values (e.g. for the model it is `gpt-4.1-2025-04-14`).

Furthermore, the [documentation](https://openai.github.io/openai-agents-js/guides/running-agents/#run-config) says `Force a specific model for all agents in the run.` which is likely **not** what we want here. I would assume to set defaults via the `Runner` and have the option to overwrite the values via `Agent`.

Also the `workflowName` is not propagated to the tracing (tracing always shows `Agent workflow`).

### Debug information

- Agents SDK version: (e.g. `v0.0.8`)
- Runtime environment (e.g. `Bun 1.2.16`)

### Repro steps

```typescript
import { Agent, Runner } from "@openai/agents";

async function main() {
  // Create an Agent without model defined
  const agent = new Agent({
    name: "TestAgent",
    instructions: "You are a test agent. Respond with the model name.'",
    // model is intentionally not defined
  });

  // Run the agent - model should be configured by the runner
  const runner = new Runner({
    workflowName: "Bug Report",
    model: "gpt-4.1-nano",
  });
  const result = await runner.run(agent, "What model are you using?");

  console.log("Response:", result.finalOutput);
}

main().catch(console.error);
```

### Expected behavior

Given the example, I would expect the `TestAgent` to use the model `gpt-4.1-nano-2025-04-14` (confirmed by setting it directly on the Agent), but according to the tracing it uses `gpt-4.1-2025-04-14`. Furthermore, I would expect the tracing to show `Bug Report` as workflow name of the trace, but it shows `Agent workflow`.

The documentation should better explain that the `Runner` provides the defaults (up for discussion).

seratch:
This pull request resolves #129

### DIFF
--- a/.changeset/fast-ads-return.md
+++ b/.changeset/fast-ads-return.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Fix #129 The model in run config should be used over an model's default setting
--- a/.changeset/fast-ads-return.md
+++ b/.changeset/fast-ads-return.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
-Fix #129 The model in run config should be used over an model's default setting
+Fix #129 The model in run config should be used over an agent's default setting
--- a/packages/agents-core/src/agent.ts
+++ b/packages/agents-core/src/agent.ts
@@ -310,6 +310,8 @@ export class Agent<
     });
   }
 
+  static DEFAULT_MODEL_PLACEHOLDER = '';
+
   name: string;
   instructions:
     | string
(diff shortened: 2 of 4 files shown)


.changeset/fast-ads-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Fix #129 The model in run config should be used over an model's default setting
```
seratch:
```suggestion
Fix #129 The model in run config should be used over an agent's default setting
```

Issue #139, opened 2025-06-23
Error: Invalid output type


bakikucukcakiroglu:
When using OpenAI’s models with the Agents library, I sometimes receive outputs that don’t conform to the JSON schema (validated by Zod) I provided. This doesn’t seem to be an issue with the library itself, so I suspect I might be overlooking something—structured outputs were always correct before I started using Agents.

I can share trace IDs from example conversations if that would help troubleshoot.

bakikucukcakiroglu:
Also my backend goes down when I get this error even if the controller calling this is wrapped in a catchAsync function to catch any error. I couldn't really understand how this error escapes from that handler. It is probably related to it being stream but not sure. Can you please explain what is going on here?

```
  const runner = new Runner({
    workflowName: 'mas_storefront',
    groupId: chatSession.sessionId,
    traceId: `trace_${message.id}`,
    tracingDisabled: false
  })

  const result = await runner.run(
    startAgent, 
    message.message.content,
    {
      context: context,
      previousResponseId: chatSession.lastResponseId,
      stream: true,
    },
  );

  for await (const event of result.toTextStream()) {
    res.write(event);
  }

  res.end();
```


```
Unhandled rejection ModelBehaviorError: Invalid output type
.executeToolsAndSideEffects ( /app/node_modules/@openai/agents-core/src/runImplementation.ts:614 )
```

bakikucukcakiroglu:
Guys please guide me through the solution. Even though there is an outputType (Structured outputs), model rarely gives an invalid response and my backend service goes down. It may not be possible to prevent model's rare invalid response but how to safely handle them? @seratch 

Here are more details https://community.openai.com/t/gpt4-1-doesnt-follow-strict-json-schema/1296679

seratch:
Hi @bakikucukcakiroglu, thanks for reaching out! Could you share a few more details about your Agent implementation? I understand the error may not occur every time, but we need more information to try to reproduce the issue you're experiencing.

Also, regarding exception handling, it is expected to handle any runtime errors on the user side. So, please include try/catch clauses in the code that runs your agents, and consider rerunning your agents if an exception can be resolved by doing so. For more details, please refer to https://openai.github.io/openai-agents-js/guides/running-agents/#exceptions

bakikucukcakiroglu:
```

export const AgentsResponseStorefront = z.object({
    content: z.string(),
    suggestive_answers: z.array(z.string()),
    recommended_products: z.array(z.string()),
[...]

PR #148, opened 2025-06-24
Fix a bug where some of the exceptions thrown from runImplementation.ts could be unhandled

seratch:
This pull request resolves the issue reported at https://github.com/openai/openai-agents-js/issues/139#issuecomment-2998999137

### DIFF
--- a/.changeset/poor-regions-try.md
+++ b/.changeset/poor-regions-try.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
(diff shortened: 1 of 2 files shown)


bakikucukcakiroglu:
@dkundel-openai I want to state that it is somehow urgent because it causes a production server to go down. It would be super nice if we can merge it as fast as possible and create a new version release!

dkundel-openai (approved):

PR #160, opened 2025-06-29
fix: if prompt is not specified return undefined - fixes #159

### LINKED ISSUE #159: Reponses API with Azure OpenAI - could not pass `null` value
The Azure OpenAI Response API implementation does not accept null input values.

Using `AzureOpenAI ` instance by calling `setDefaultOpenAIClient(client)` with instructions fails with error ` 400 Unknown parameter: ''.`.

```
import { AzureOpenAI } from 'openai';
const client = new AzureOpenAI({ ...azOAIClientOptions, defaultQuery: { 'api-version': 'preview', } });
setDefaultOpenAIClient(client);
```

The error seems to be due to Responses API call including `prompt: null` in body, from line below in code.

https://github.com/openai/openai-agents-js/blob/d1252282930f5e2ae9cc79cebddd963246518c8c/packages/agents-openai/src/openaiResponsesModel.ts#L374

Can this be changed to return `undefined` instead to fix this issue?

slowtick:
Azure OpenAI Responses API does not accept `null` values. It fails with `400 Unknown parameter: ''` error if `prompt` value is passed as `null`. Returning undefined, removes the prompt field and makes it work with Azure Open AI Responses API.

### DIFF
--- a/packages/agents-openai/src/openaiResponsesModel.ts
+++ b/packages/agents-openai/src/openaiResponsesModel.ts
@@ -365,13 +365,15 @@ function isMessageItem(item: protocol.ModelItem): item is protocol.MessageItem {
   return false;
 }
 
-function getPrompt(prompt: ModelRequest['prompt']): {
-  id: string;
-  version?: string;
-  variables?: Record<string, any>;
-} | null {
+function getPrompt(prompt: ModelRequest['prompt']):
+  | {
+      id: string;
+      version?: string;
+      variables?: Record<string, any>;
+    }
+  | undefined {
   if (!prompt) {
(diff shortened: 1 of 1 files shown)


dkundel-openai:
@kwhinnery-openai any concerns with this?

@slowtick can you please add a changeset for this?

kwhinnery-openai:
This should work fine with OAI Responses also

seratch:
OK, this should be good to go. @slowtick could you add chageset file for this? you can use `pnpm changeset` command for it (select agents-openai as the changed package, the version bump for all packages is "patch").

PR #164, opened 2025-07-01
Add MCP server tool filtering support to agents-js

### LINKED ISSUE #162: Add tool filter to MCP support
Implement the feature equivalent to https://github.com/openai/openai-agents-python/pull/861

vrtnis:
Hey all - this PR brings a new tool filtering feature to our MCP integrations.  This resolves #162 based on https://github.com/openai/openai-agents-python/pull/861

You can now:

1. Use `createStaticToolFilter` for quick allow/block lists

2. Provide a custom filter function to make dynamic decisions based on `runContext`

3. Updated core API so `getAllTools` now accepts your `runContext`, ensuring filters are applied correctly

4. Documentation enhancements in `docs/guides/mcp.mdx`

5. New example (`examples/mcp/tool-filter-example.ts`) to demonstrate both static and dynamic filtering

6. Comprehensive tests covering static lists, callable filters, multi-server hierarchies, and cache interactions

All tests pass and you can try it out by `pnpm -F mcp start:tool-filter`

### DIFF
--- a/.changeset/heavy-yaks-mate.md
+++ b/.changeset/heavy-yaks-mate.md
@@ -1,5 +1,0 @@
----
-"@openai/agents-openai": patch
----
-
-Handle function call messages with empty content in Chat Completions
--- a/.changeset/heavy-yaks-mate.md
+++ b/.changeset/heavy-yaks-mate.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents-openai": patch
+---
+
+Handle function call messages with empty content in Chat Completions
--- a/.changeset/hungry-suns-search.md
+++ b/.changeset/hungry-suns-search.md
@@ -1,6 +1,6 @@
 ---
-'@openai/agents-realtime': minor
-'@openai/agents-core': minor
+'@openai/agents-realtime': patch
+'@openai/agents-core': patch
 ---
 
 agents-core, agents-realtime: add MCP tool-filtering support (fixes #162)
--- a/.changeset/hungry-suns-search.md
+++ b/.changeset/hungry-suns-search.md
@@ -1,0 +1,6 @@
+---
+'@openai/agents-realtime': minor
+'@openai/agents-core': minor
+---
+
+agents-core, agents-realtime: add MCP tool-filtering support (fixes #162)
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -105,7 +105,7 @@ using `createStaticToolFilter` or a custom function:
 ```ts
 const server = new MCPServerStdio({
   fullCommand: 'my-server',
-  toolFilter: createStaticToolFilter(['safe_tool'], ['danger_tool']),
+  toolFilter: createStaticToolFilter({ allowed: ['safe_tool'], blocked: ['danger_tool'] }),
 });
 
 const dynamicServer = new MCPServerStreamableHttp({
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -97,6 +97,23 @@ For **Streamable HTTP** and **Stdio** servers, each time an `Agent` runs it may 
 
 Only enable this if you're confident the tool list won't change. To invalidate the cache later, call `invalidateToolsCache()` on the server instance.
 
+### Tool filtering
+
+You can restrict which tools are exposed from each server. Pass either a static filter
+using `createStaticToolFilter` or a custom function:
+
+```ts
+const server = new MCPServerStdio({
+  fullCommand: 'my-server',
+  toolFilter: createStaticToolFilter(['safe_tool'], ['danger_tool']),
+});
+
+const dynamicServer = new MCPServerStreamableHttp({
+  url: 'http://localhost:3000',
+  toolFilter: ({ runContext }, tool) => runContext.context.allowAll || tool.name !== 'admin',
+});
+```
+
 ## Further reading
 
 - [Model Context Protocol](https://modelcontextprotocol.io/) – official specification.
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -10,6 +10,7 @@ import hostedStreamExample from '../../../../../examples/docs/mcp/hostedStream.t
 import hostedHITLExample from '../../../../../examples/docs/mcp/hostedHITL.ts?raw';
 import streamableHttpExample from '../../../../../examples/docs/mcp/streamableHttp.ts?raw';
 import stdioExample from '../../../../../examples/docs/mcp/stdio.ts?raw';
+import toolFilterExample from '../../../../../examples/docs/mcp/tool-filter.ts?raw';
 
 The [**Model Context Protocol (MCP)**](https://modelcontextprotocol.io) is an open protocol that standardizes how applications provide tools and context to LLMs. From the MCP docs:
 
@@ -99,24 +100,13 @@ Only enable this if you're confident the tool list won't change. To invalidate t
 
 ### Tool filtering
 
-You can restrict which tools are exposed from each server. Pass either a static filter
-using `createMCPToolStaticFilter` or a custom function:
+You can restrict which tools are exposed from each server by passing either a static filter via `createMCPToolStaticFilter` or a custom function. Here’s a combined example showing both approaches:
 
-```ts
-const server = new MCPServerStdio({
-  fullCommand: 'my-server',
-  toolFilter: createMCPToolStaticFilter({
-    allowed: ['safe_tool'],
-    blocked: ['danger_tool'],
-  }),
-});
-
-const dynamicServer = new MCPServerStreamableHttp({
-  url: 'http://localhost:3000',
-  toolFilter: ({ runContext }, tool) =>
-    runContext.context.allowAll || tool.name !== 'admin',
-});
-```
+<Code 
+  lang="typescript" 
+  code={toolFilterExample} 
+  title="Tool filtering" 
+/>
 
 ## Further reading
 
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -100,17 +100,21 @@ Only enable this if you're confident the tool list won't change. To invalidate t
 ### Tool filtering
 
 You can restrict which tools are exposed from each server. Pass either a static filter
-using `createStaticToolFilter` or a custom function:
+using `createMCPToolStaticFilter` or a custom function:
 
 ```ts
 const server = new MCPServerStdio({
   fullCommand: 'my-server',
-  toolFilter: createStaticToolFilter({ allowed: ['safe_tool'], blocked: ['danger_tool'] }),
+  toolFilter: createMCPToolStaticFilter({
+    allowed: ['safe_tool'],
+    blocked: ['danger_tool'],
+  }),
 });
 
 const dynamicServer = new MCPServerStreamableHttp({
   url: 'http://localhost:3000',
-  toolFilter: ({ runContext }, tool) => runContext.context.allowAll || tool.name !== 'admin',
+  toolFilter: ({ runContext }, tool) =>
+    runContext.context.allowAll || tool.name !== 'admin',
 });
 ```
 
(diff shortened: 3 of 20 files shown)


dkundel-openai:
@seratch can you review please?

.changeset/hungry-suns-search.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openai/agents-realtime': minor
+'@openai/agents-core': minor
```
seratch: sem versioning can upgrade minor for feature additions, but this project is still in its early stage; so we're doing patch version upgrades for most changes so far. could you go with patch this time?

```suggestion
'@openai/agents-realtime': patch
'@openai/agents-core': patch
```

docs/src/content/docs/guides/mcp.mdx
```diff
@@ -97,6 +97,23 @@ For **Streamable HTTP** and **Stdio** servers, each time an `Agent` runs it may
...
 
+### Tool filtering
+
+You can restrict which tools are exposed from each server. Pass either a static filter
+using `createStaticToolFilter` or a custom function:
+
+```ts
+const server = new MCPServerStdio({
+  fullCommand: 'my-server',
+  toolFilter: createStaticToolFilter(['safe_tool'], ['danger_tool']),
```
seratch: giving names to the arguments should be easier to read for devs; so how about using an argument object instead? The approach is better for future enhancement too.
```suggestion
  toolFilter: createStaticToolFilter({ allowed: ['safe_tool'], blocked: ['danger_tool'] }),
```

packages/agents-core/src/shims/mcp-server/node.ts
```diff
@@ -97,7 +100,52 @@ export class NodeMCPServerStdio extends BaseMCPServerStdio {
   }
 
   // The response element type is intentionally left as `any` to avoid explosing MCP SDK type dependencies.
-  async listTools(): Promise<MCPTool[]> {
+  protected async _applyToolFilter(
```
seratch: I think this filtering should be done on the agent/runner side, not within MCP server wrapper.

seratch (changes requested):
Thanks for sending this! I haven't verified the actual behavior yet but let me share a few quick comments

packages/agents-core/src/shims/mcp-server/node.ts
vrtnis: thanks for the feedback! that makes sense. as such i’ve now removed _applyToolFilter from the server classes and moved the filtering logic up to the agent level instead.

vrtnis:
thanks for the comments, @seratch! esp. the point about moving filtering logic that makes sense for keeping the server implementation focused on transport.
i’ve pushed an update that moves filtering into the agent and runner layer, where it can be handled with proper context. look forward to your feedback!

packages/agents-core/src/mcp.ts
```diff
@@ -138,11 +154,14 @@ export class MCPServerStdio extends BaseMCPServerStdio {
...
-  async listTools(): Promise<MCPTool[]> {
+  async listTools(
+    runContext?: RunContext<any>,
+    agent?: Agent<any, any>,
+  ): Promise<MCPTool[]> {
     if (this.cacheToolsList && this._cachedTools) {
       return this._cachedTools;
     }
-    const tools = await this.underlying.listTools();
+    const tools = await this.underlying.listTools(runContext, agent);
```
seratch: I mean you don't need to pass run context and agent to the underlying MCP server wrapper. You can let the MCP server wrapper to provide the access to the tool filter and you can run the filter after receiving all the tools from the MCP server on the agent side (specifically getAllMcpFunctionTools).

seratch (changes requested):

packages/agents-core/src/mcp.ts
vrtnis: Got it,  that makes sense... i've removed the `runContext` and agent from the call to `underlying.listTools()`, and now the filtering runs entirely within `getAllMcpFunctionTools()` on the agent side.

vrtnis:
Thanks @seratch.  That makes sense and i’ve removed the `runContext` and `agent` from the call to `underlying.listTools()`, so that now filtering happens entirely in `getAllMcpFunctionTools()` on the agent side.

packages/agents-core/src/mcp.ts
```diff
@@ -27,10 +34,14 @@ export const DEFAULT_STREAMABLE_HTTP_MCP_CLIENT_LOGGER_NAME =
  */
 export interface MCPServer {
   cacheToolsList: boolean;
+  toolFilter?: ToolFilterCallable | ToolFilterStatic;
   connect(): Promise<void>;
   readonly name: string;
   close(): Promise<void>;
-  listTools(): Promise<MCPTool[]>;
+  listTools(
```
seratch: Can you revert this interface change? I don't think this is necessary. If I am missing something, please let me know!

vrtnis: good catch. you're right..i’ve just reverted the interface change ...the overload wasn't strictly necessary for the current implementation.

packages/agents-core/src/mcpUtil.ts
```diff
@@ -0,0 +1,46 @@
+import type { Agent } from './agent';
+import type { RunContext } from './runContext';
+import type { MCPTool } from './mcp';
+import type { UnknownContext } from './types';
+
+/** Context information available to tool filter functions. */
+export interface ToolFilterContext<TContext = UnknownContext> {
```
seratch: I should have mentioned this in earlier reviews, but could you rename all these new types and utilities to have "MCP" prefix? "Tool" means a more generic thing including not only MCP's tools but also built-in tools & function tools. 

So, the names would be MCPToolFilterContext, MCPToolFilterCallable, MCPToolFilterStatic, and createMCPToolStaticFilter.

packages/agents-core/src/mcp.ts
seratch: Perhaps, you might haven't push the changes? if you revert the things I am talking about there, the MCP server classes should have any diff.

vrtnis: i might have pasted the comment too soon.. i pushed the revert in commit 4dce2d5

packages/agents-core/src/mcpUtil.ts
vrtnis: got it..no worries..i will update this as well

vrtnis: @seratch i've updated all of the new types and utilities to use the `MCP` prefix where needed i.e., 

- `MCPToolFilterContext`  
- `MCPToolFilterCallable`  
- `MCPToolFilterStatic`  
- `createMCPToolStaticFilter`  
 [4c1adbd](https://github.com/openai/openai-agents-js/pull/164/commits/4c1adbd3dc7631ab01b404ff9b7aa4cdb5ead6b4) 
lmk if any furher tweaks are needed here

packages/agents-core/src/mcp.ts
seratch: Thanks for putting your efforts on this but what I mean here is something like this: https://github.com/seratch/openai-agents-js/commit/73eb37ae81801f9f64c4853e98338c7404a195ff

examples/mcp/tool-filter-example.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  await mcpServer.connect();
+
+  try {
+    await withTrace('MCP Tool Filter Example', async () => {
+      const agent = new Agent({
+        name: 'MCP Assistant',
+        instructions:
+          'Use the filesystem tools to answer questions. The write_file tool is blocked via toolFilter.',
```
seratch: > The write_file tool is blocked via toolFilter

this part needs to be removed

examples/mcp/tool-filter-example.ts
```diff
@@ -0,0 +1,51 @@
...
+      });
+
+      console.log('Listing sample files:');
+      let result = await run(agent, 'List the files in the current directory.');
+      console.log(result.finalOutput);
+
+      console.log('\nAttempting to write a file (should be blocked):');
+      result = await run(
+        agent,
+        'Create a file named test.txt with the text "hello"',
```
seratch: currently this is blocked either way because the mcp server has the access only to sample_files/ 

my diff changed this as well

seratch (changes requested):
Did thorough review and I found your current implementation actually does nothing for allowing/blocking tools. I've implemented draft version of change here: https://github.com/seratch/openai-agents-js/commit/73eb37ae81801f9f64c4853e98338c7404a195ff Hope sharing this clarifies things!

examples/mcp/tool-filter-example.ts
vrtnis: Resolved through merge of 73eb37a, example updated accordingly.

examples/mcp/tool-filter-example.ts
vrtnis: Resolved through merge of 73eb37a, example updated accordingly.  Also, e4ba592 explicitly lists sample_files to allow listing.

vrtnis: thanks for sharing your draft, @seratch! You’re absolutely right, my plumbing additions never actually invoked the filter, so allow/block lists weren’t doing anything.

I’ve since merged your implementation from 73eb37a and used that as the foundation. Appreciate the clarity your diff brought to this.

One question: should we also update the test to explicitly include `toolUseBehavior: 'stop_on_first_tool'` in the agent config? That setting ensures the run exits immediately with a `"Tool \"write_file\" is blocked by MCP filter."` error rather than continuing or looping, which might make the filter behavior clearer and more predictable in test output.

happy to add that if helpful!

seratch: Thanks for pointing this out. I agree the agent should behave in the way you mentioned for the scenario. Adding test patterns would be appreciated!

vrtnis: Just added an integration test, i.e., `mcpToolFilter.integration.test.ts` that spins up the stub FS server and verifies that `list_directory` (allowed) works and `write_file` (blocked) gets immediately rejected with the expected message using `toolUseBehavior: 'stop_on_first_tool'`. 

I also expanded the original unit tests to cover no filter, allow-only, block-only, async filters, thrown errors, and cache invalidation.  This should give us better coverage across edge cases.

Lastly, I added a script to `package.json` so the example can now be run via `pnpm examples:tool-filter`. 

Thoughts? Let me know if we should have anything else covered!

vrtnis: Noticed that CI had failed on the merge commit (b5ff9ac) due to a type issue with modelResponses in the integration test. I fixed it by adding the correct ModelResponse[] type.  It should be good now.

.changeset/heavy-yaks-mate.md
```diff
@@ -0,0 +1,5 @@
+---
```
seratch: can you delete this file? this is not part of this PR

docs/src/content/docs/guides/mcp.mdx
```diff
@@ -97,6 +97,27 @@ For **Streamable HTTP** and **Stdio** servers, each time an `Agent` runs it may
 
 Only enable this if you're confident the tool list won't change. To invalidate the cache later, call `invalidateToolsCache()` on the server instance.
 
+### Tool filtering
+
+You can restrict which tools are exposed from each server. Pass either a static filter
+using `createMCPToolStaticFilter` or a custom function:
+
+```ts
```
seratch: We consistently use compiled code snippets in docs. can you do this?

- Add `examples/docs/mcp/tool-filter.ts`, which compiles and works
- Add `import stdioExample from '../../../../../examples/docs/mcp/tool-filter.ts?raw';` and use it here

If you're unsure or don't have time for this, please just remove this change.

seratch (changes requested):
The changes are now all good, but there are a few things to adjust 🙏

docs/src/content/docs/guides/mcp.mdx
vrtnis: Yes, good point.  I've added the example as suggested. Confirmed that it compiles and works (screenshot attached).
screenshot00-28-00

seratch (approved):
LGTM

seratch:
@vrtnis Thank you so much for working on this PR 👏

vrtnis:
Thanks @seratch, appreciate the thoughtful reviews and guidance throughout. Learned a lot working through this one! 🙌

Issue #44, opened 2025-06-06
Signal handling in Agents SDK interferes with hot reload in development environments


actuallyzefe:
### Describe the bug

Using the @openai/agents SDK in a NestJS app causes the development server to hang or fail to restart properly when running with `nest start --watch`.

This appears to be due to the SDK registering `process.on('SIGTERM')` and other signal handlers that invoke async cleanup (e.g., getGlobalTraceProvider().shutdown()). In environments like NestJS (which doesn’t fully restart the Node process), this cleanup interferes with reload logic.

Notably, the same SDK integration works fine in a regular TypeScript app using `ts-node-dev --respawn` , where the process is properly restarted on each change.

### Debug information

- Agents SDK version: 0.0.3
- Runtime environment: Node.js 22.16.0
- NestJS version: 11.0.0
- OS: macOS

### Repro steps
- 🔴 NestJS issue repo: [NestJS  Hot Reload Issue ](https://github.com/actuallyzefe/nestjs-oepnai-agents-hotreload-issue)
` export OPENAI_API_KEY=sk`
`npm run start:dev`

- ✅ Working TypeScript repo: [Basic Typescript Hot Reload Issue (works fine.)](https://github.com/actuallyzefe/hotreload-issue-ts)

`export OPENAI_API_KEY=sk`
`npm run dev`

### Expected behavior
Signal handling and shutdown logic in the SDK should not interfere with development workflows.

It’s worth noting that NestJS is one of the most widely used Node.js frameworks, and its default development mode 
(`nest start --watch`) is commonly used by many teams.

rossrobino:
I ran into this with Vite as well, just opened PR #45 which fixes for me, I think it's the same issue.

actuallyzefe:
> I ran into this with Vite as well, just opened PR [#45](https://github.com/openai/openai-agents-js/pull/45) which fixes for me, I think it's the same issue.
I was just commenting on your PR, if it gets merged and the issue is resolved, I’ll go ahead and close the issue. 
Thanks for jumping on this! 🙌

PR #184, opened 2025-07-07
Fix #175 by removing system.exit calls in the built-in handlers

### LINKED ISSUE #175: SIGINT/SIGTERM handlers prevent graceful shutdown
### Describe the bug

Instances of the `TraceProvider` class register global `SIGINT`/`SIGTERM` signal handlers upon being instantiated (see [here](https://github.com/openai/openai-agents-js/blob/d1252282930f5e2ae9cc79cebddd963246518c8c/packages/agents-core/src/tracing/provider.ts#L22) and [here](https://github.com/openai/openai-agents-js/blob/d1252282930f5e2ae9cc79cebddd963246518c8c/packages/agents-core/src/tracing/provider.ts#L194-L198)). Those handlers internally call `process.exit()`, which immediately ends the process without allowing time for user-defined signal handlers to run.

Furthermore, it seems that the [GLOBAL_TRACE_PROVIDER](https://github.com/openai/openai-agents-js/blob/d1252282930f5e2ae9cc79cebddd963246518c8c/packages/agents-core/src/tracing/provider.ts#L213-L219) is being instantiated simply by virtue of the `@openai/agents` package being imported. I am not sure where this is happening, but I have verified it experimentally (see minimal reproduction below).

As a result, simply importing this library breaks user-defined `SIGINT`/`SIGTERM` handlers - for example,  handlers that run custom logic to gracefully shut down other parts of an application (e.g. waiting for in-progress HTTP requests to finish, closing long-lived resources, etc.), because the signal handlers installed by this library end up calling `process.exit()` before the user-defined signal handlers have a chance to complete their shutdown routines.

As a temporary workaround, I am removing existing signal handlers with `process.removeAllListeners('SIGINT');` before registering my own, and making sure to call `await getGlobalTraceProvider().shutdown();` in my custom signal handler that performs graceful shutdown. This seems to be work, but feels hacky and liable to break over time (using [removeAllListeners](https://nodejs.org/api/events.html#emitterremovealllistenerseventname) is considered bad practice in general, per the Node.js docs).

Ideally, I think this library wouldn't install any signal handlers of its own, since that interferes with the user's ability to manage the way their application responds to shutdown signals it receives.

I think this bug was likely introduced in #45.

### Debug information

- Agents SDK version: (e.g. `v0.0.9`)
- Runtime environment (e.g. `Node.js 24.3.0`)

### Repro steps

Here's a minimal reproduction:

```javascript
// Prints an empty array of listeners.
console.log('SIGINT listeners before import: ', process.listeners('SIGINT'));

// This library installs a SIGINT handler simply by virtue of being imported.
// Removing this import makes the graceful shutdown logic work as expected.
const agents = require('@openai/agents');

// Prints an array containing a single listener.
console.log('SIGINT handlers after import: ', process.listeners('SIGINT'));

// Install custom SIGINT handler to trigger graceful shutdown.
let shutdown = false;
process.on('SIGINT', async () => {
  console.log('Received SIGINT. Beginning graceful shutdown');
  shutdown = true;
});

// Simulate long-running work that needs to be gracefully shut down.
const interval = setInterval(() => {
  if (shutdown) {
    console.log('Shutting down'); // This never prints! The process exits before it ever gets here.
    clearInterval(interval);
  }
}, 5000);

```

Run this script with Node.js and press `Ctrl-C`. The process will print `Received SIGINT. Beginning graceful shutdown` and then exit immediately, before ever printing `Shutting down`.

### Expected behavior

I expected to be in control of how my application responds to `SIGINT`/`SIGTERM` signals. I expected the app to not exit until the event loop runs out of work or my own code calls `process.exit()`.

seratch:
This pull request resolves #175 and related reports. This commit https://github.com/openai/openai-agents-js/commit/25165dfe29de59014398dea503b4bf5282fb9d23 in #45 introduced system.exit calls, which didn't exist before the fix. So, reverting them should make sense. An alternative approach would be to introduce a new flag env variable to turn this on/off but I don't think it's necessary.

### DIFF
--- a/.changeset/tame-hotels-hope.md
+++ b/.changeset/tame-hotels-hope.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Fix #175 by removing internal system.exit calls
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,8 @@ jobs:
           version: 10.13.1
           run_install: false  # do manual install to avoid heap memory errors
       - name: Install dependencies
+        env:
+          NODE_OPTIONS: --max_old_space_size=4096
         # lower parallelism to save RAM
         run: pnpm install --frozen-lockfile --network-concurrency=1
       - name: Build all packages
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'pnpm'
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,8 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    env:
+      NODE_OPTIONS: "--max_old_space_size=4096"
     strategy:
       matrix:
         # https://nodejs.org/en/about/previous-releases
@@ -15,20 +17,18 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
           version: 10.13.1
-          run_install: false  # do manual install to avoid heap memory errors
+          run_install: false
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'pnpm'
       - name: Install dependencies
-        env:
-          NODE_OPTIONS: --max_old_space_size=4096
-        # lower parallelism to save RAM
-        run: pnpm install --frozen-lockfile --network-concurrency=1
+        run: pnpm install
       - name: Build all packages
         run: pnpm build
       - name: Run linter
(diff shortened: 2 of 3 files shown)


dkundel-openai (approved):

nathanjcochran:
I think the gotcha here is that just registering the `SIGINT`/`SIGTERM` handlers changes Node's default signal handling behavior, causing it to no longer exit when it receives these signals. This is likely why the user's program was hanging in #44, and why the `process.exit()` calls were added in the first place in #45. See [the Node.js docs](https://nodejs.org/api/process.html#signal-events), which say:

> 'SIGTERM' and 'SIGINT' have default handlers on non-Windows platforms that reset the terminal mode before exiting with code 128 + signal number. If one of these signals has a listener installed, its default behavior will be removed (Node.js will no longer exit).

The same is true for the `unhandledRejection`/`uncaughtException` events. Registering handlers for them changes the default behavior (which is to exit the process).

As a result, this PR will likely break apps that _don't_ have their own handlers installed for those signals/events, because the presence of your handlers prevents them from exiting (which would be the normal/default Node.js behavior otherwise). On the flip side, including `process.exit()` calls in your handlers breaks apps that _do_ have their own handlers installed for those signals/events, because in that case, the normal/default behavior would be for the app to _not_ automatically exit (see #175).

At the end of the day, the difficulty is really just that you're registering signal handlers in a library package 😅. The only way to do it "correctly", as far as I can tell, is for your handlers to exit the process if no _other_ handlers are installed for the relevant signal/event, but not to exit the process if there is at least one other handler registered for the signal/event. That would more-or-less preserve the default Node.js behavior, as explained in the docs I linked above.

seratch:
Thanks for your comment; I will look into this later on.

seratch:
Added check logic within those SDK built-in listeners and changed the behavior to exit the process only when there are no other listeners. This does not change the current behavior for simple code snippets + resolves the unintuitive process termination for apps with custom listeners.

Issue #187, opened 2025-07-09
Agent outputType type error with zod@3.25.68+


teetlaja:
### Describe the bug
When providing a properly-typed Zod schema as the property to an instance, TypeScript raises an error, even though the types appear compatible.
This prevents using custom Zod object schemas as the agent's output shape. `outputType``Agent`
### Debug information
- Agents SDK version: `v0.0.10`
- Runtime environment: `Node.js 22.16.0`
- Package manager: `yarn`
- TypeScript version: `5.8.3`
- Zod version: `3.25.40`
- IDE: WebStorm

### Repro steps
1. Install dependencies:
``` sh
    yarn add @openai/agents zod
```
1. Create a file `my-test.ts` with the following minimal code:
``` typescript
    import { Agent } from '@openai/agents';
    import { z } from 'zod';

    const SimpleOutput = z.object({
      hotelName: z.string(),
      hotelWeb: z.string(),
    });

    // This line causes a TypeScript error:
    const agent = new Agent({
      name: 'test agent',
      instructions: 'Test instructions',
      outputType: SimpleOutput, // Error is reported here!
    });
```
1. Observe the TypeScript error:
``` 
    TS2322: Type 'ZodObject<{ hotelName: ZodString; hotelWeb: ZodString; }, ...>' is not assignable to type 'AgentOutputType<unknown> | undefined'.
    ...
    Types have separate declarations of a private property '_cached'.
```
### Expected behavior
The above use of a custom Zod schema should be accepted as a valid for the constructor.
It should be possible to specify any compatible Zod object to shape the output type, without TypeScript errors. `outputType``Agent`

teetlaja:
Full error output
```
TS2322: Type
ZodObject<{
    hotelName: ZodString;
    hotelWeb: ZodString;
}, "strip", ZodTypeAny, {
    hotelName: string;
    hotelWeb: string;
}, {
    hotelName: string;
    hotelWeb: string;
}>
is not assignable to type AgentOutputType<unknown> | undefined
Type 'ZodObject<{ hotelName: ZodString; hotelWeb: ZodString; }, "strip", ZodTypeAny, { hotelName: string; hotelWeb: string; }, { hotelName: string; hotelWeb: string; }>' is not assignable to type 'ZodObject<any, UnknownKeysParam, ZodTypeAny, { [x: string]: any; }, { [x: string]: any; }> | JsonSchemaDefinition'.
[...]

PR #192, opened 2025-07-09
Update quickstart.mdx - outdated example for tool call. 

TimonBerlin:
The given example doesn’t work anymore because the “parameters” field is now required.
I wasent able to run it without providing a JSON Schema for parameters.

### DIFF
--- a/docs/src/content/docs/guides/quickstart.mdx
+++ b/docs/src/content/docs/guides/quickstart.mdx
@@ -85,6 +85,13 @@ const historyFunFact = tool({
   name: 'history_fun_fact',
   // The description is used to describe **when** to use the tool by telling it **what** it does.
   description: 'Give a fun fact about a historical event',
+  // This agent takes no parameters, so we provide an empty JSON schema.
+  parameters: {
+    type: 'object',
+    properties: {},
+    required: [],
+    additionalProperties: false
+  }
   execute: async () => {
     // The output will be returned back to the Agent to use
     return 'Sharks are older than trees.';


docs/src/content/docs/guides/quickstart.mdx
```diff
@@ -85,6 +85,13 @@ const historyFunFact = tool({
   name: 'history_fun_fact',
   // The description is used to describe **when** to use the tool by telling it **what** it does.
   description: 'Give a fun fact about a historical event',
+  // This agent takes no parameters, so we provide an empty JSON schema.
+  parameters: {
+    type: 'object',
+    properties: {},
+    required: [],
+    additionalProperties: false
+  }
```
seratch: comma is missing:
```suggestion
  },
```

seratch: If you have more time to tweak this, we prefer a slightly simpler one like:
```python
import { Agent, run, tool } from '@openai/agents';
import { z } from 'zod';

const historyFunFact = tool({
  // The name of the tool will be used by the agent to tell what tool to use.
  name: 'history_fun_fact',
  // The description is used to describe **when** to use the tool by telling it **what** it does.
  description: 'Give a fun fact about a historical event',
  // This agent takes no parameters, so we provide an empty JSON schema.
  parameters: z.object({}),
  execute: async () => {
    // The output will be returned back to the Agent to use
    return 'Sharks are older than trees.';
  },
});
```

seratch (changes requested):
Thank you for sending this PR! Left a few comments

docs/src/content/docs/guides/quickstart.mdx
TimonBerlin: Hi @seratch, thanks for your response. I had already tried it as you suggested, using an empty Zod object, but I wasn’t able to get it running.

image

seratch: Probably, it is due to https://github.com/openai/openai-agents-js/issues/187, which is zod compat issue that we're going to apply workaround soon

TimonBerlin: Should i then go with the empty zod object or keep the json schema ?

seratch: we consistently use zod as the primary option in the docs and examples, so please go ahead with an empty zod object here.

TimonBerlin: @seratch  alright done ! Thank you very much :)

PR #194, opened 2025-07-10
fix: the key is repeated for the same content

gitplus:
When the user enters the same content, the key is repeated


seratch:
Thank you!

PR #197, opened 2025-07-11
Update Japanese docs to reflect the latest changes

seratch:
This pull request updates all the Japanese document pages.


docs/src/content/docs/guides/release.mdx
```diff
@@ -3,8 +3,6 @@ title: Release process
 description: Learn how we version and release the SDK and recent changes.
 ---
 
-import { Content as AgentsSdkChangelog } from '../../../../../packages/agents/CHANGELOG.md';
```
seratch, line 6: This is unused and can cause an error on the translated page side. I could fix the error, but just removing it is easier.

PR #209, opened 2025-07-14
feat: support noise reduction argument for session

chollier:
adds support noise reduction argument for session


seratch (changes requested):
Thank you! Could you update this method as well? > https://github.com/openai/openai-agents-js/blob/v0.0.11/packages/agents-realtime/src/openaiRealtimeBase.ts#L375-L409

chollier:
@seratch done

seratch (approved):
LGTM; @dkundel-openai if you have time, can you also check this?

packages/agents-realtime/src/clientMessages.ts
```diff
@@ -33,6 +33,10 @@ export type RealtimeTracingConfig =
     }
   | 'auto';
 
+export type RealtimeInputAudioNoiseReductionConfig = {
+  type: 'near_field' | 'far_field';
```
dkundel-openai: Can we change this to be to make sure that if we add more values in the future it doesn't require someone to upgrade the SDK

```suggestion
  type: 'near_field' | 'far_field' | (string & {});
```

dkundel-openai (approved):

PR #212, opened 2025-07-15
Update docs and README to encourage using `@openai/zod` - ref #187

seratch:
To prevent users from being confused with the zod latest version incompatibility, this pull request updates all the code snippets in this repo to use `@openai/zod`, which is the best workaround for #187 issue at this moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ console.log(result.finalOutput);
 ## Functions example
 
 ```js
-import { z } from 'zod';
+import { z } from '@openai/zod';
 import { Agent, run, tool } from '@openai/agents';
 
 const getWeatherTool = tool({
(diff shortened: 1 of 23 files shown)


seratch:
Still trying to eliminate the issue, but I haven't yet found a viable solution for it.

dkundel-openai:
This seems odd? I introduced `@openai/zod` to explicitly be used for internal purposes only. Meaning it should only be used for things like parsing/validating protocols. For anything inferred like Structured Outputs and tool arguments it intentionally uses `zod` in the code bases as to use the peer dependency that we expect customers to install. So asking someone to use `@openai/zod` is essentially relying on an internal abstraction that might break (for example we might bump it to zod 4 if we decided it served our use cases better)

seratch:
Yeah, this makes sense and this is why I wanted to know your thoughts on this idea. Let's continue exploring both short-term workaround and fundamental solutions for this issue.

PR #215, opened 2025-07-15
Feat: api client invalidation

dreammonkey:
Fixes an issue where OpenAI API keys would persist forever once set, preventing proper switching between different API keys/clients in multi-tenant (nextjs) applications.

The issue was that once an OpenAI API key was set using setDefaultOpenAIKey() or setDefaultOpenAIClient(), the OpenAI client would be cached in the OpenAIProvider class and never updated, even when switching to a different API key. This prevented proper API key switching in multi-tenant applications where different requests need different OpenAI clients with different keys/projects.

Changes Made:                                                                                                      

- Provider Registry System: Added a registry to track OpenAI providers that need to be notified when default keys/clients change 
- Automatic Client Invalidation: When `setDefaultOpenAIKey()` or `setDefaultOpenAIClient()` is called, all registered providers invalidate their cached clients

Best Regards,
Diederik

PS: I also fixed a type on the docs ;)

### DIFF
--- a/.changeset/tasty-dots-relax.md
+++ b/.changeset/tasty-dots-relax.md
@@ -1,0 +1,9 @@
+---
+'@openai/agents-extensions': patch
+'@openai/agents-realtime': patch
+'@openai/agents-openai': patch
+'@openai/agents-core': patch
+'@openai/agents': patch
+---
+
+introduce api client invalidation
--- a/packages/agents-core/src/agent.ts
+++ b/packages/agents-core/src/agent.ts
@@ -383,7 +383,7 @@ export class Agent<
   }
 
   /**
-   * Ouput schema name
+   * Output schema name
    */
   get outputSchemaName(): string {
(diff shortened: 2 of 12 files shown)


seratch:
Thanks for sending this! However, instead of trying to use the default global API key and client for your use case, we recommend creating a `Runner` along with a specific `OpenAIProvider`. This means that you can instantiate `OpenAIClient`, `OpenAIProvider` with the API client, and then pass the `modelProvider` to a `Runner`. Please refer to `modelProvider` in [this document page](https://openai.github.io/openai-agents-js/guides/running-agents/#run-arguments) for more details. This approach allows each app in your multi-tenant setup to have its own `OpenAIClient` with its API key.

Regarding your proposal here, we’d like to avoid complicating the default/global key management just for this purpose. It'd be appreciated if you could understand this.

Lastly, if you’re fine to resend only the document typo fix, we’d be happy to merge it right away!

PR #220, opened 2025-07-16
fix(realtime-ws): stop accidental cancellation error

dkundel-openai:
Depending on the turn detection settings the Realtime API might cancel an ongoing response automatically which in the current code might run into occasional race conditions with the `this.#ongoingResponse` check which in turn causes errors.

This change shuffles the logic around so that if an interruption is happening because of turn detection it will check if the session config has cancellation of responses enabled and only cancel if not. For other interruption calls like guardrails or manual interruptions we will still perform an interruption. 

This change isn't needed in WebRTC because the turn detection based interruptions happen server side.

### DIFF
--- a/.changeset/strong-lobsters-repair.md
+++ b/.changeset/strong-lobsters-repair.md
@@ -1,0 +1,6 @@
+---
+"@openai/agents-extensions": patch
+"@openai/agents-realtime": patch
+---
+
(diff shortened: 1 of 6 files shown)


bakks (approved):

seratch:
Create changeset and get it merged 👍

PR #222, opened 2025-07-17
Add input_fidelity parameter support to image generation tool

seratch:
This pull request adds the new parameter to the responses api tool.

image 👉 image


.github/workflows/test.yml
```diff
@@ -13,7 +13,8 @@ jobs:
     strategy:
       matrix:
         # https://nodejs.org/en/about/previous-releases
-        node-version: [20, 22, 24]
+        # Using 24.3.x due to https://github.com/pnpm/pnpm/issues/9743
```
seratch: Spent time to identify the cause of pnpm i errors... this is necessary for now

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -525,7 +526,10 @@ function getInputItems(
           type: 'code_interpreter_call',
           id: item.id!,
           code: item.providerData?.code ?? '',
-          results: item.providerData?.results ?? [],
+          // This property used to be results, so keeping both for backward compatibility
+          // That said, this property cannot be passed from a user, so it's just API's internal data.
+          outputs:
```
seratch, line 531: this should be safe; the underlying sdk made this breaking changes but agents sdk does not touch this property so far

seratch (commented):
a few comments for reviewers

dkundel-openai (approved):

PR #225, opened 2025-07-17
Fix #216 Publicly accessible PDF file URL is not yet supported in the input_file content data

### LINKED ISSUE #216: Publicly accessible PDF file URL is not yet supported in the input_file content data
I'm getting the error `Error: 400 Missing required parameter: 'input[0].content[0]'.` when running this code:

```
const agentInput = [
  { role: "user", content: [{ type: "input_file", file: <pdf_dataurl> }] },
];
const runResult = await run(orderingEntrypointAgent, agentInput, { maxTurns: 500 });
```

From what I can tell from the documentation, I have given the input in the required shape but not sure why its erroring. 

Thanks!

seratch:
This pull request resolves #216

### DIFF
--- a/packages/agents-openai/src/openaiResponsesModel.ts
+++ b/packages/agents-openai/src/openaiResponsesModel.ts
@@ -174,6 +174,7 @@ function converTool<_TContext = unknown>(
         tool: {
           type: 'image_generation',
           background: tool.providerData.background,
+          input_fidelity: tool.providerData.input_fidelity,
           input_image_mask: tool.providerData.input_image_mask,
           model: tool.providerData.model,
           moderation: tool.providerData.moderation,
--- a/.changeset/shiny-beers-tie.md
+++ b/.changeset/shiny-beers-tie.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Add input_fidelity parameter support to image generation tool
--- a/examples/tools/image-generation.ts
+++ b/examples/tools/image-generation.ts
@@ -18,7 +18,7 @@ async function main() {
   const agent = new Agent({
     name: 'Image generator',
     instructions: 'You are a helpful agent.',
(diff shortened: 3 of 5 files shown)


examples/basic/remote-pdf.ts
```diff
@@ -0,0 +1,27 @@
+import { Agent, run } from '@openai/agents';
+
+const URL = 'https://www.berkshirehathaway.com/letters/2024ltr.pdf';
```
seratch: This URL is used in the document, so I went with it: https://platform.openai.com/docs/guides/pdf-files?api-mode=responses

packages/agents-core/src/types/protocol.ts
```diff
@@ -91,9 +91,17 @@ export const InputFile = SharedBase.extend({
    */
   file: z
     .string()
+    .describe(
```
seratch, line 94: developers can pass 1) as string or 2) `{ url: '....' }`

packages/agents-openai/src/openaiResponsesModel.ts
```diff
@@ -525,7 +535,10 @@ function getInputItems(
           type: 'code_interpreter_call',
           id: item.id!,
           code: item.providerData?.code ?? '',
-          results: item.providerData?.results ?? [],
+          // This property used to be results, so keeping both for backward compatibility
```
seratch: same change with https://github.com/openai/openai-agents-js/pull/222

seratch (commented):
comments for reviewers

examples/basic/remote-pdf.ts
dkundel-openai: Can we use small url here? It's a nit pick but URL is a built in class in JS

seratch: thanks, renamed this and the original remote-image.ts as well

PR #232, opened 2025-07-18
Revert "Fix release tagging"

seratch:
Reverts openai/openai-agents-js#231

The root cause is that we are not using `changeset publish` for publishing and our custom publish command does `pnpm publish -r --no-git-checks`. In this case, the step outputs are empty, and there is no workaround so far.

@dkundel-openai was there any specific reason to avoid `changeset publish`? If `changeset publish` works, the following step to push tags should work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,6 +53,7 @@ jobs:
           NPM_CONFIG_ACCESS: public
 
       - name: Tag release
+        # currently this step does not work due to https://github.com/openai/openai-agents-js/pull/232
         if: steps.changesets.outputs.published == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           NPM_CONFIG_ACCESS: public
 
       - name: Tag release
+        if: steps.changesets.outputs.published == 'true'
```
seratch, line 57:
```suggestion
        # currently this step does not work due to https://github.com/openai/openai-agents-js/pull/232
        if: steps.changesets.outputs.published == 'true'
```

PR #241, opened 2025-07-22
Fix #20 Update metadata.ts files when releasing a new version

### LINKED ISSUE #20: Update metadata.ts files when releasing a new version
Currently, the following files are not updated while the release job successfully adds correct version metadata to the deployed npm packages.

```
	modified:   packages/agents-core/src/metadata.ts
	modified:   packages/agents-extensions/src/metadata.ts
	modified:   packages/agents-openai/src/metadata.ts
	modified:   packages/agents-realtime/src/metadata.ts
	modified:   packages/agents/src/metadata.ts
```

We want to keep these files in git version control so that the release CI job should update them to be up-to-date.

seratch:
This pull request resolves #20 

Currently, this project does not block commit push to main branch by bot user, so it should work. That said, let's check if it really works when we make a new release next time.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,7 +53,6 @@ jobs:
           NPM_CONFIG_ACCESS: public
 
       - name: Push metadata changes
-        if: always()
         run: |
           files="packages/agents-core/src/metadata.ts packages/agents-extensions/src/metadata.ts packages/agents-openai/src/metadata.ts packages/agents-realtime/src/metadata.ts packages/agents/src/metadata.ts"
           if git diff --quiet -- $files; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -52,6 +52,20 @@ jobs:
           NPM_CONFIG_PROVENANCE: true
           NPM_CONFIG_ACCESS: public
 
+      - name: Push metadata changes
+        if: always()
```
dkundel-openai: why do we need this `always()`?

seratch:
```suggestion
```

seratch: removed!

PR #242, opened 2025-07-22
fix: include JsonSchema definitions in mcpTool inputSchema

asanzfonetic:
We noticed that the framework’s inputSchema parameter is expected to comply with JSON Schema standards. However, when the schema contains $ref references, the corresponding definitions were not being included. This caused incomplete schemas to be generated.

This PR fixes the issue by modifying a single line of code to ensure that all referenced JSON Schema definitions are properly included within the inputSchema.

This change improves schema completeness and compatibility with JSON Schema validators.

### DIFF
--- a/packages/agents-core/src/metadata.ts
+++ b/packages/agents-core/src/metadata.ts
@@ -3,11 +3,11 @@
 
 export const METADATA = {
   "name": "@openai/agents-core",
-  "version": "0.0.7",
+  "version": "0.0.13",
   "versions": {
-    "@openai/agents-core": "0.0.7",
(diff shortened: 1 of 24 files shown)


packages/agents-core/src/metadata.ts
```diff
@@ -3,11 +3,11 @@
 
 export const METADATA = {
```
seratch: would it be possible to ask you to remove metadata.ts changes? Since the next release, these diff won't show up.

asanzfonetic: done!

seratch (approved):

Issue #247, opened 2025-07-23
stopAtToolNames does not return finalOutput


mhashas:
```
export const createGetActionLogsAgent = (
  model: Model,
  adminApiToolkit: AdminApiToolkit,
) =>
  new Agent({
    name: 'get_action_logs',
    instructions: buildInstruction(),
    model: model,
    tools: [
      createGetActionLogsFiltersTool(),
      createQueryActionLogsTool(adminApiToolkit),
    ],
    toolUseBehavior: { stopAtToolNames: ['query_action_logs'] },
  });

  const tools = [
    actionLogsAgent.asTool({
      toolName: 'get_action_logs',
      toolDescription: GET_ACTION_LOGS_TOOL_DESCRIPTION,
      // in theory should not be needed, but without it the content is '', might be a bug in the library
      customOutputExtractor: (output: RunResult<any, any>) => {
        return output.finalOutput;
      },
    }),
```

I would expect to not have to write this customOutputExtractor. queryActionLogs returns a very long string in case it matters

vrtnis:
Hey @mhashas  – good catch.

The reason, as you probably have determined, you're seeing an empty result (unless you use `customOutputExtractor`) is that when you use `stopAtToolNames`, the sub-agent halts before emitting a final assistant message, and `asTool()` by default just returns that last message. So technically, it's not a bug per se, but something that could be enhanced.

There’s a draft PR open [#249](https://github.com/openai/openai-agents-js/pull/249) which introduces an opt-in `returnRunResult: true` flag to `agent.asTool()`. With that flag set, you’ll get the full `RunResult` (including `finalOutput`, interruptions like approval requests, etc.), without needing to write a custom extractor.

So your example could be simplified to:

```ts
const tool = actionLogsAgent.asTool({
  toolName: 'get_action_logs',
  returnRunResult: true,
});
```
BTW you can also unit test the tool by calling `.invoke()` directly. It takes a `RunContext` (which can be empty unless you rely on extra context), and a json string as input:

```ts
const res = await tool.invoke(new RunContext(), '{"input": "your filters"}');
expect(res.finalOutput).toContain('expected log data');
```

Feel free to give the PR [#249](https://github.com/openai/openai-agents-js/pull/249) a look and lmk if it fits your use case or if there are other improvements you’d like to see there.

mhashas:
Hey, thanks for the answer, I commented in your MR.

vrtnis:
[...]

PR #249, opened 2025-07-23
enhancement(agent): add returnRunResult option to agent.asTool() to expose RunResult and interruptions

### LINKED ISSUE #243: Agent that has Agent `asTool` that has Tool that `needsApproval` does not return interruptions
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/) ✅ 
- **Have you searched for related issues?** Others may have faced similar issues. ✅ 

### Describe the bug

When I have an Agent as a tool that has a tool itself that has `needsApproval` to be `true`, the run `result.interruptions` array remains empty.

### Debug information

- Agents SDK version: v0.0.11
- Runtime environment Node.js 22.15.1

### Repro steps

Agent A has Agent B `asTool` in `tools` array
```javascript
const agentA = new Agent({
  ...
  tools: [
    agentB.asTool({
      ...
    })
  ]
})
```

Agent B has Tool A in `tools` array
```javascript
const agentB = new Agent({
  ...
  tools: [
    toolA
  ]
})
```

Tool A has `needsApproval` to return `true`
```javascript
const toolA = tool({
  ...
  needsApproval: async (_ctx, _input) => {
    console.log('> approval check', _ctx, _input);
    return true;
  }
})
```

Run result does not return any interruptions
```javascript
const result = await runner.run(agentA, conversation, promptOpts);
if(result.interruptions?.length){
  console.log('> pause for approval', JSON.stringify(result.state, null, 2));
}
```

### Expected behavior

Expected behavior is that `result.interruptions` will have items, however it does not.  The conversation just returns a `finalOutput` that implies that something is happening in the background.

What I have noticed is that if I have Agent B has a handoff within Agent A, then it works and I receive `result.interruptions` items, however its conversational behavior as a handoff is not what I'm trying to achieve.

Is this intended SDK behavior?

vrtnis:
This adds a `returnRunResult` option to `agent.asTool()` that allows developers to access the full `RunResult` of the sub-agent when invoked as a tool.

By default, `asTool()` only returns the final output string of the sub-agent, which means any interruptions (e.g. approval requests via `needsApproval`) are hidden from the parent agent.

With this flag enabled:
- The full `RunResult` is returned
- Interruptions are visible to the outer agent
- Human-in-the-loop and nested approval workflows now work as expected

This change is backward-compatible and opt-in. It does not affect any existing users of `asTool()`.

Also, note that we have,

- Updated the "Agents as Tools" section in `tools.mdx` to explain the `returnRunResult` flag.
- Expanded the example in `agentsAsTools.ts` to include an `echoTool` using `returnRunResult`.
- Added a live usage example in `tool-use-behavior.ts` that demonstrates how to inspect `RunResult.newItems`.

Added tests to verify that:
  - `asTool({ returnRunResult: true })` returns a `RunResult` with the correct `finalOutput`.
  - `stopAtToolNames` works without needing a `customOutputExtractor`, fixing [#247](https://github.com/openai/openai-agents-js/issues/247).
  - `RunContext` is properly used in `.invoke()` to satisfy typing and test runtime behavior.

### Example usage:
```ts
agentB.asTool({
  toolName: 'agentB',
  returnRunResult: true,
});
```
Resolves [#243](https://github.com/openai/openai-agents-js/issues/243) and also
Resolves [#247](https://github.com/openai/openai-agents-js/issues/247) – `asTool()` now correctly returns `finalOutput` with `stopAtToolNames`, without requiring a custom extractor

Tested with `pnpm test`, `pnpm link`, and `pnpm build-check`.

### DIFF
--- a/.changeset/cuddly-poets-refuse.md
+++ b/.changeset/cuddly-poets-refuse.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Add `returnRunResult` option to `agent.asTool()` to expose full RunResult, including interruptions such as approvals.
--- a/docs/src/content/docs/guides/tools.mdx
+++ b/docs/src/content/docs/guides/tools.mdx
@@ -85,6 +85,7 @@ Under the hood the SDK:
 - Creates a function tool with a single `input` parameter.
 - Runs the sub‑agent with that input when the tool is called.
 - Returns either the last message or the output extracted by `customOutputExtractor`.
+- If `returnRunResult` is set, returns the full `RunResult` so nested interruptions can be inspected.
 
 ---
 
--- a/examples/basic/tool-use-behavior.ts
+++ b/examples/basic/tool-use-behavior.ts
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { Agent, run, tool } from '@openai/agents';
+import { Agent, run, tool, RunContext, RunResult } from '@openai/agents';
 
 const Weather = z.object({
   city: z.string(),
@@ -46,6 +46,19 @@ const agent3 = new Agent({
   tools: [getWeatherTool, saySomethingTool],
 });
 
+const agentWithRunResult = new Agent({
+  name: 'Tool agent with RunResult',
+  instructions,
+  toolUseBehavior: { stopAtToolNames: ['get_weather'] },
+  outputType: Weather,
+  tools: [getWeatherTool, saySomethingTool],
+});
+
+const weatherToolWithRunResult = agentWithRunResult.asTool({
+  toolName: 'weather_summary',
+  returnRunResult: true,
+});
+
 async function main() {
   const input = 'What is the weather in San Francisco?';
   const result = await run(agent, input);
@@ -65,6 +78,12 @@ async function main() {
   const finalOutput3 = result3.finalOutput;
   // The weather in San Francisco is sunny. Thanks for asking!
   console.log(finalOutput3);
+
+  const result4 = (await weatherToolWithRunResult.invoke(
+    new RunContext(),
+    JSON.stringify({ input }),
+  )) as RunResult<any, Agent<any, any>>;
+  console.log('agentWithRunResult.newItems:', result4.newItems);
 }
 
 main();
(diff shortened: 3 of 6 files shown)


seratch:
Thanks for sending this PR. 

I don't yet have a clear idea of the best approach for this issue, but at a glance, this PR doesn't seem like the best solution for the linked issues to me. When an agent is attached as a tool, it should behave as a tool. Therefore, exposing the agent's run result as a tool output sounds like a different type of tool definition. Also, the agent that calls the sub-agent as a tool shouldn't need to know the difference between agent tools and other types of tools.

There are a few other things I am currently working on, but will check this issue sometime soon.

vrtnis:
Thanks for the quick intial feedback!...yep, totally agree that when an agent is used as a tool, it should behave like one.  The flag felt like the simplest, least disruptive way to support these issues.  When `returnRunResult` is off, everything behaves as before (returning the usual string or object) and so existing code stays unchanged.   

We only get the richer `RunResult` when we explicitly opt in, which means the parent agent doesn’t need to treat agent tools differently unless that extra detail is wanted (like in #247’s missing `finalOutput`). 

We could also ship an `AgentAsToolWithInterruptions` helper (see https://github.com/LuminaX-alt/openai-agents-js/pull/6) to auto-copy interruptions, but that wouldn’t address the `finalOutput` issue either, so the flag felt like the cleaner path overall.  

I guess the broader question is: since there are already other flags and possibly more coming wondering if it might be worth thinking about whether these kinds of options should live under an experimental umbrella, similar to how Cloudflare Workers support is presently marked. Just some thoughts. Happy to address changes or restart with a different approach.

mhashas:
Ok, so as I understand, there's a bit of incompatibility between the two options:

`stopAtToolNames`
- The final output will be the output of the first tool in the list that was called.

`asTool`
- If not provided, the last message from the agent will be used.

When we talk about the last message from the agent, we're talking about the last message from the history that comes from the assistant, not the last message from the history? Any reason why we want the last message to strictly be from the assistant?

vrtnis:
> Ok, so as I understand, there's a bit of incompatibility between the two options:
> 
> `stopAtToolNames`
> 
> * The final output will be the output of the first tool in the list that was called.
> 
> `asTool`
> 
> * If not provided, the last message from the agent will be used.
> 
> When we talk about the last message from the agent, we're talking about the last message from the history that comes from the assistant, not the last message from the history? Any reason why we want the last message to strictly be from the assistant?

@mhashas i've answered this in https://github.com/openai/openai-agents-js/issues/247#issuecomment-3165676649. thought this clarification belongs in #247 as #249 is the proposed implementation path that avoids needing a custom extractor.

Issue #245, opened 2025-07-22
Runtime SyntaxError for typescript 5.8.3 with openai/agents v0.0.12 at agents-core with node v22 locally and AWS lambda


cam-nano:
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/)

Yes

- **Have you searched for related issues?** Others may have faced similar issues.

Closest one I could find was: https://github.com/openai/openai-agents-js/issues/213

### Describe the bug

Getting the issue when using the @openai/agent with a lambda (node22x) and locally (node v22.12.0). I have tried all sorts of permutations of tsconfig and removing dependencies but am still getting this error. I have a test repo for this issue that I can make public if needed.

```
{
  "errorType": "Runtime.UserCodeSyntaxError",
  "errorMessage": "SyntaxError: Cannot use import statement outside a module",
  "trace": [
    "Runtime.UserCodeSyntaxError: SyntaxError: Cannot use import statement outside a module",
    "    at _loadUserApp (file:///var/runtime/index.mjs:1106:17)",
    "    at async UserFunction.js.module.exports.load (file:///var/runtime/index.mjs:1148:21)",
    "    at async start (file:///var/runtime/index.mjs:1332:23)",
    "    at async file:///var/runtime/index.mjs:1339:1"
  ]
}
```

Getting a similar issue when running locally with node v22.12.0 as a unit test:

```
 /Users/cam/Documents/user/dev/test-openai/node_modules/@openai/agents/dist/index.js:1
    import { setDefaultModelProvider } from '@openai/agents-core';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

    > 1 | import { Agent, run, setDefaultOpenAIKey } from '@openai/agents';
        | ^
      2 | import { OPENAI_API_KEY } from "../constants";
      3 | import { ClientError } from '../utils/client-error';
      4 |

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1314:40)
      at Object.<anonymous> (src/commands/process-agent-command.ts:1:1)
      at Object.<anonymous> (tests/agent-command.test.ts:1:1)
```

I have created a 'skinny' repo with only the handler, command and test in it and am getting the same result as in my larger app. 

### Debug information

- Agents SDK version: (e.g. `v0.0.1`)

```
"@openai/agents": "^0.0.12",
```

- Runtime environment (e.g. `Node.js 22.16.0`)

[...]

Issue #213, opened 2025-07-15
Cannot use import statement outside a module Node 20 and older versions


actuallyzefe:
# OpenAI Agents SDK Bug Report

## Have you read the docs?

✅ **Yes** - [Agents SDK docs](https://openai.github.io/openai-agents-js/)

## Have you searched for related issues?

✅ **Yes** - This appears to be a module compatibility issue not yet reported

## Describe the bug

The OpenAI Agents SDK v0.0.10 is not compatible with CommonJS environments. When imported in a CommonJS project, it throws a SyntaxError because the CommonJS distribution files contain ES module import statements instead of proper CommonJS require statements.

**Error Message:**

```
SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:76:18)
    at wrapSafe (node:internal/modules/cjs/loader:1283:20)
    at Module._compile (node:internal/modules/cjs/loader:1328:27)
```

## Debug information

- **Agents SDK version:** `v0.0.10`
- **Runtime environment:** `Node.js v18.20.8` (also tested on v20.17.0 from package.json engines)
- **Module system:** CommonJS (`"module": "commonjs"` in tsconfig.json)
- **Package manager:** npm/yarn

## Repro steps

**Minimal reproduction script:**

1. Create a new Node.js project with CommonJS configuration:

```bash
mkdir test-openai-agents
cd test-openai-agents
npm init -y
npm install @openai/agents@^0.0.10
```

2. Create `tsconfig.json`:

```json
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "commonjs",
    "moduleResolution": "node",
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true
  }
}
```

3. Try to import the SDK using CommonJS:

```bash
node -p "const { Agent } = require('@openai/agents'); console.log('Success');"
```

**Result:** Throws `SyntaxError: Cannot use import statement outside a module`

**Root cause:** The `/dist/index.js` file (CommonJS export) contains ES module syntax:

```javascript
import { setDefaultModelProvider } from "@openai/agents-core";
import { OpenAIProvider } from "@openai/agents-openai";
// ... more ES module imports
```

## Expected behavior

The CommonJS distribution should use `require()` statements and `module.exports` instead of ES module `import`/`export` statements. The package should work seamlessly in both CommonJS and ES module environments as suggested by the dual export configuration in package.json.

**Expected CommonJS output should be:**

```javascript
[...]

PR #258, opened 2025-07-25
issue-245: fix CommonJS interoperability in tsc-multi and shim files

CharGrnmn:
Solves: 

1. https://github.com/openai/openai-agents-js/issues/213
2. https://github.com/openai/openai-agents-js/issues/245

I am building my own personal npm package. Will confirm here when works as expected (for AWS Lambda)


CharGrnmn:
I wasn't quite able to get this deployed and I'm not certain this works. If would like I will close

.husky/pre-commit
```diff
@@ -8,5 +8,9 @@ git update-index --again
 
 # check for secrets
 if [ -z "$CI" ] && [ -z "$GITHUB_ACTIONS" ]; then
-  trufflehog git file://. --since-commit HEAD --fail
+  if command -v trufflehog >/dev/null 2>&1; then
+    trufflehog git file://. --since-commit HEAD --fail
+  else
+    echo "Warning: trufflehog not found, skipping secret scanning"
+  fi
```
seratch: Perhaps, you might have difficulty with this, we would like to avoid this change.
```suggestion
  trufflehog git file://. --since-commit HEAD --fail
```

tsc-multi.json
```diff
@@ -1,6 +1,6 @@
 {
   "targets": [
-    { "extname": ".js", "module": "es2022", "moduleResolution": "node" },
+    { "extname": ".js", "module": "commonjs", "moduleResolution": "node" },
```
seratch, line 3: Thanks for changing this. This change looks reasonable, but we need to verify these before merging the change:
- Do more manual testing
- Add automated tests to ensure this works as expect under https://github.com/openai/openai-agents-js/tree/main/integration-tests

I will spend time on this ~later this week~ soon.

seratch (changes requested):
Thank you so much for sending this. As mentioned at one of the comments, this should be good to go, but we need more tests and add automated tests under https://github.com/openai/openai-agents-js/tree/main/integration-tests; I will work on it later this week. If you're willing to help us complete these too, it'd be appreciated.

CharGrnmn:
@seratch sure, feel to work on and close this pull request whenever. Unfortunately. I'm not sure if I will be able to work on in a timely fashion. This is no longer blocking me(was able to work around albeit took time) so no rush. You have my thanks

vrtnis:
hey @seratch! quick note up front: thanks to @CharlieGreenman for the import.meta shim idea, it’s folded in here.

on the Lambda side: the node 22 runtime still ships without automatic module-syntax detection. i do not believe there is a public commitment from aws to change that e.g. see [aws-lambda-base-images#233](https://github.com/aws/aws-lambda-base-images/issues/233). `NODE_OPTIONS=--experimental-require-module` is ignored, and the original Compute blog post hasn’t been updated since launch.

so here we’re still hitting two gaps that need addressing. (1) build artifacts: we need both the ESM and CJS outputs, properly wired via the exports field, so environments like lambda and jest can safely require() the SDK without breaking compatibility for existing ESM consumers.

(2) there’s runtime safety.  I think the sdk's shims must avoid unguarded access to `import.meta` so that CJS environments don’t crash when executing runtime code like `loadEnv()`.

Given everything that needs to land here, I think a fresh PR is the cleanest path forward. I’m putting one together to handle both the dual-build setup and the shim guard. (and would include integration tests that cover import, require, and execution inside the relevant Lamda runtime).  I'll have it wrapped up tmrw for review.

seratch:
Thanks for the inputs. I believe the direction should be good to go, but I just don't have time to do thorough testing right now. Any inputs / suggestions / contributions for this would be greatly appreciated.

vrtnis:
Just a quick update that this work has been superseded by [PR #333](https://github.com/openai/openai-agents-js/pull/333). which rolls in a full dual esm + cjs build, adds import.meta shim guards, and includes integration tests for Lambda and cjs runtimes

It covers the same cjs interoperability issues raised here (#245, #213) but also covers other serverless environments, so it should (after feedback) fully address the use cases we discussed in this thread.

seratch:
Thanks for your contribution! I've moved your commit to #346; if you have any comments, please feel free to write in!

PR #260, opened 2025-07-25
fix: persist container_id in codeInterpreterTool history

### LINKED ISSUE #253: codeInterpreterTool history causes container_id exception on reuse
# Summary 
When manually managing and reusing the run history of an Agent using the `codeInterpreterTool`, the agent throws an exception during subsequent runs due to a missing `container_id`. This occurs even when the tool is correctly configured during initialization.

### Environment

- Agents SDK version: (e.g. `v0.0.13`)
- Runtime environment (e.g. `Node.js 22.14.0`)

### Reproduction Steps

```ts
let history: AgentInputItem[] = [];

const agent = new Agent({
  name: "Test Agent",
  model: "gpt-4.1",
  tools: [codeInterpreterTool({container:{
    type: "auto",
    file_ids: []
  }})],
});

const runner = new Runner();

const result = await runner.run(
  agent,
  [
    {
      role: "user",
      content: [
        {
          type: "input_text",
          text: "Please generate me a .txt File with the following content: 'Hello, world!'",
        },
      ],
    },
  ],
);

history.push(...result.history);

const result2 = await runner.run(agent, [
  ...history,
  {
    role: "user",
    content: [{ type: "input_text", text: "Test" }],
  },
]);

console.log(result2.history);
```

The second call to `runner.run` fails with the following error:

```text
/Users/malzacher/Documents/dev/playground/node_modules/openai/src/core/error.ts:72
      return new BadRequestError(status, error, message, headers);
             ^


BadRequestError: 400 Missing required parameter: 'input[1].container_id'.
    at Function.generate (/Users/malzacher/Documents/dev/playground/node_modules/openai/src/core/error.ts:72:14)
    at OpenAI.makeStatusError (/Users/malzacher/Documents/dev/playground/node_modules/openai/src/client.ts:438:28)
    at OpenAI.makeRequest (/Users/malzacher/Documents/dev/playground/node_modules/openai/src/client.ts:661:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async OpenAIResponsesModel.#fetchResponse (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-openai/src/openaiResponsesModel.ts:873:22)
    at async <anonymous> (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-openai/src/openaiResponsesModel.ts:894:24)
    at async <anonymous> (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-core/src/tracing/createSpans.ts:41:16)
    at async OpenAIResponsesModel.getResponse (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-openai/src/openaiResponsesModel.ts:893:22)
    at async <anonymous> (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-core/src/run.ts:377:39)
    at async <anonymous> (/Users/malzacher/Documents/dev/playground/node_modules/@openai/agents-core/src/tracing/context.ts:60:20) {
  status: 400,
  headers: Headers {},
  requestID: 'req_9910cbd8b41e044b5a7d65b62405520d',
  error: {
    message: "Missing required parameter: 'input[1].container_id'.",
    type: 'invalid_request_error',
    param: 'input[1].container_id',
    code: 'missing_required_parameter'
  },
  code: 'missing_required_parameter',
  param: 'input[1].container_id',
  type: 'invalid_request_error'
}
```

### Expected behavior

Reusing the full run history (e.g., result.history) as input to a subsequent run should either:

- Succeed with all required tool metadata preserved or reconstructed, or
- Provide a clear utility or API guidance on how to filter or regenerate only the relevant parts of the history for reuse.

vrtnis:
Replaying a full run history with the `codeInterpreterTool` fails because the API’s `container_id` stayed in snake_case and never got mapped to `providerData.containerId`.  As such, I've added a small `snakeToCamelCase` helper so incoming `container_id` becomes `containerId`, and we still convert it back when building the next request. Now the container ID survives in history and a second `runner.run(...)` won’t throw “missing container_id” anymore.

Changes:

- new `snakeToCamelCase` utility  
- apply it in `convertToOutputItem`  
- continue converting back in `getInputItems`  
- unit tests for both directions  

With these changes, replaying the full `result.history` now succeeds on the second `runner.run(...)` and the `container_id` is correctly preserved across runs.

Resolves #253

### DIFF
--- a/.changeset/stale-times-bathe.md
+++ b/.changeset/stale-times-bathe.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Fix codeInterpreterTool run replay by correctly using container_id from providerData (fixes #253)
--- a/packages/agents-openai/src/openaiResponsesModel.ts
+++ b/packages/agents-openai/src/openaiResponsesModel.ts
@@ -541,7 +541,7 @@ function getInputItems(
           outputs:
             item.providerData?.outputs ?? item.providerData?.results ?? [],
           status: CodeInterpreterStatus.parse(item.status ?? 'failed'),
-          container_id: item.providerData?.containerId,
+          container_id: item.providerData?.container_id,
         };
 
         return entry;


seratch:
Hi @vrtnis, thanks for sending this pull request. 

If I understand this pull request correctly,
1. This pull request does not bring anything new to #253, the container_id is already available in providerData
2. I was actually going to take the same approach to covert all the property names to camelCase in the initial phase of this SDK development 😄 but our team decided not to do so. The latest consistent rule is to use snake_case ones (=the original key names) in providerData for OpenAI models

For these two reasons, I don't think this change should be merged.

vrtnis:
Thanks so much for the thoughtful guidance @seratch and context on the camelCase convention 😄! really appreciate you taking the time to clarify.

You are right that the `container_id` is already available in `providerData`. This is true at rest, but `getInputItems` only looks for `containerId`, so the value is never read and the request is sent without it.

Essentially, after the first run, history holds `providerData.container_id` (snake_case).  
`getInputItems`, however, destructures `providerData.containerId` (camelCase) when building the next request. Because that key isn’t present, `container_id` is left out of the payload, and the second run fails with “missing container_id”.

While keeping everything snake case, i'm thinking that we can patch getInputItems in `openaiResponsesModel.ts` like so:

```ts
const containerId =
  providerData.containerId ?? providerData.container_id;
  ```
This keeps the key in snake_case the whole time, avoids any new helpers or conversions, and ensures the replay works as expected.

Would be happy to open a separate (and much slimmer!) PR for this fix if that sounds good. Let me know!

seratch:
Thanks for pointing the cause out. Definitely, the line https://github.com/openai/openai-agents-js/blob/v0.0.13/packages/agents-openai/src/openaiResponsesModel.ts#L544 is the cause of the issue. No need to consider `providerData.containerId` here, so can you just change the property name there? `providerData` is any typed, so this type of bug could happen (although I introduced internal provider data types for casting).

Having the single line of change should resolve the issue. So, including the change and changeset file in this PR would be appreciated.

vrtnis:
Just pushed an update, thanks again for the helpful feedback! I’ve removed the camelCase handling and made the one-line fix to directly use `providerData.container_id`, as suggested

PR #270, opened 2025-07-29
Examples: Update latestAgent before displaying

mattbobambrose:
You need to update the latestAgent from the latest result before displaying it in the console

### DIFF
--- a/examples/basic/chat.ts
+++ b/examples/basic/chat.ts
@@ -56,11 +56,12 @@ async function main() {
       history.push(user(message));
       const result = await run(latestAgent, history);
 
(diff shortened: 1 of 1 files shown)


mattbobambrose:
I realized that my edit was based on a question asking for multiple tool calls at once, and doesn't reflect behavior for a single tool call. The problem is that the input "What's the weather in Dallas and Danville?" gets a different behavior from "What's the weather in Dallas?" Multiple cities result in multiple tool calls being made, while a single city waits for additional interaction to make a tool call, whether that interaction is more text or another tool call. My update is correct for multiple cities, but not for a single city. The existing code is correct for a single city, but not multiple cities.

PR #285, opened 2025-08-01
Hacky way to add Tool Image support

Tianrider:
Chat Completions allow only Tool messages with string content

**How this bypasses the Chat Completions limitation**
- To bypass: 
```
    } else if (item.type === 'function_call_result') {
      flushAssistantMessage();
      const funcOutput = item;
      if (funcOutput.output.type !== 'text') {
        throw new UserError(
          'Only text output is supported for chat completions. Got item: ' +
            JSON.stringify(item),
        );
      }

      result.push({
        role: 'tool',
        tool_call_id: funcOutput.callId,
        content: funcOutput.output.text,
        ...funcOutput.providerData,
      });
```

### DIFF
--- a/packages/agents-core/src/items.ts
+++ b/packages/agents-core/src/items.ts
@@ -156,6 +156,24 @@ export class RunToolApprovalItem extends RunItemBase {
   }
 }
 
+export class RunInjectedMessageItem extends RunItemBase {
+  public readonly type = 'injected_message_item' as const;
+
+  constructor(
+    public rawItem: protocol.MessageItem,
+    public agent: Agent,
+  ) {
+    super();
+  }
+
+  toJSON() {
+    return {
+      ...super.toJSON(),
+      agent: this.agent.toJSON(),
+    };
+  }
+}
+
 export type RunItem =
   | RunMessageOutputItem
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this pull request, and it's an interesting approach. However, we may not be able to accept this change, as we need to stay aligned with our API platform design. We're aware of the image data needs. Hope we'll bring some enhancements for those use cases.

Tianrider:
> Thanks for sending this pull request, and it's an interesting approach. However, we may not be able to accept this change, as we need to stay aligned with our API platform design. We're aware of the image data needs. Hope we'll bring some enhancements for those use cases.

Hi @seratch. Aplogies, this PR was submitted accidentally and contains incomplete changes. We’ll follow up with a more complete and intentional update in the next few days.

seratch:
Thanks for your reply. If I understand correctly, your change enables a function tool to insert a user/assistant message in addition to (or instead of) a tool call output item. I understand the intention, however, we may not accept the workaround because utilizing the technique could make agent apps harder to debug, plus we (as OpenAI platform team) may not be able to continue supporting the approach in the long run.

PR #290, opened 2025-08-01
feat: add sse server implementation for mcp client

### LINKED ISSUE #289: MCP SSE server transport support
There's a lot of stock mcp internally and the sse protocol is still in use Streamable HTTP is great but it's going to take a while to get it right If it's possible I can propose a pr to support the sse protocol.

xierenyuan:
MCP SSE server transport support close #289

### DIFF
--- a/examples/mcp/package.json
+++ b/examples/mcp/package.json
@@ -1,6 +1,7 @@
 {
   "private": true,
   "name": "mcp",
+  "type": "module",
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.12.0",
     "@openai/agents": "workspace:*",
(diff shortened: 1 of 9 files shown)


examples/mcp/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "private": true,
   "name": "mcp",
+  "type": "module",
```
seratch: either is actually fine, but this example should work without this. to make all the examples consistent, can you revert this change?

xierenyuan: image
Actually, if I don't add it, it'll give me an error. I'll remove it.

seratch: thanks, this was due to an overlooked error in a different commit. Fixed it by https://github.com/openai/openai-agents-js/commit/de6e2cd5bba6962350a166203215f4c388ca0a08

seratch (approved):
LGTM; I confirmed the added example works without issues

PR #294, opened 2025-08-05
docs: Clarify `name` parameter in Basic Configuration

MuhammadHamidRaza:
This pull request aims to improve the clarity and consistency of the `Agent` documentation for the JavaScript SDK.

Currently, the `name` parameter is used in the code example but is not explicitly listed in the "Basic Agent definition" bullet points. This can be confusing for new users who may not immediately realize that `name` is a core and required property.

This change aligns the JavaScript documentation with the Python documentation, where `name` is clearly listed as a required parameter in the "Basic configuration" section.

The update will make the documentation more explicit and easier to follow, providing a better experience for developers defining their agents.


seratch:
Thanks, but the list contains the three essential components of an agent, not listing the required properties of an agent object.

PR #327, opened 2025-08-12
Fix #311 - Auto install trufflehog by running prescript

### LINKED ISSUE #311: Auto install trufflehog by running prescript or another way
When I tried to commit new changes, I encountered an unknown error that prevented the commit. After reviewing the CONTRIBUTING.md file again, I discovered that trufflehog needs to be installed.

To improve the contribution process and avoid confusion, I suggest automating this requirement (e.g., via a pre-commit hook or a setup script).

NiteshSingh17:
Fiexs: https://github.com/openai/openai-agents-js/issues/311

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -6,7 +6,28 @@ git update-index --again
 pnpm eslint $(git diff --cached --name-only --diff-filter=ACMR | sed 's| |\\ |g') --fix
 git update-index --again
 
-# check for secrets
+
+# Only adjust PATH on macOS/Linux/WSL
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this. I set "good first issue" to the issue, but I hesitate to accept the script as the one we'll maintain. We mention the necessity to install it in the contributing guide: https://github.com/openai/openai-agents-js/blob/main/CONTRIBUTING.md#submitting-a-pull-request so if you're open to just improving the error message etc., it would be an easier-to-maintain solution for us.

NiteshSingh17:
Not needed

PR #274, opened 2025-07-30
fix: make zod a peerDependency to dedupe installed versions

ianduvall:
### Summary

Moves `zod` to be a peer dependency

### Issue number

Closes #187

### Checks

- [ ] I've added new tests (if relevant)
- [ ] I've added/updated the relevant documentation
- [x] I've run `pnpm test` and `pnpm test:examples`
  - [ ] (If you made a major change) I've run `pnpm test:integration` [(see details)](https://github.com/openai/openai-agents-js/tree/main/integration-tests)
- [x] I've made sure tests pass
- [ ] I've added a changeset using `pnpm changeset` to indicate my changes

PR #342, opened 2025-08-15
Fix #187 Agent outputType type error with zod@3.25.68+

### LINKED ISSUE #187: Agent outputType type error with zod@3.25.68+

seratch:
This pull request adds my additional commit to https://github.com/openai/openai-agents-js/pull/274 and it resolves https://github.com/openai/openai-agents-js/issues/187

### DIFF
--- a/packages/agents-core/package.json
+++ b/packages/agents-core/package.json
@@ -88,12 +88,11 @@
     "@modelcontextprotocol/sdk": "^1.17.2"
   },
   "dependencies": {
-    "@openai/zod": "npm:zod@3.25.40 - 3.25.67",
     "debug": "^4.4.0",
     "openai": "^5.10.1"
   },
   "peerDependencies": {
-    "zod": "3.25.40 - 3.25.67"
+    "zod": "^3.25.40"
   },
   "peerDependenciesMeta": {
     "zod": {
@@ -121,7 +120,7 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.12",
-    "zod": "3.25.40 - 3.25.67"
+    "zod": "^3.25.40"
   },
   "files": [
     "dist"
--- a/packages/agents-core/src/agent.ts
+++ b/packages/agents-core/src/agent.ts
@@ -1,4 +1,4 @@
-import type { ZodObject } from 'zod/v3';
+import type { ZodObject } from 'zod';
 
 import type { InputGuardrail, OutputGuardrail } from './guardrail';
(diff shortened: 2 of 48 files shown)


packages/agents-core/package.json
```diff
@@ -88,12 +88,11 @@
     "@modelcontextprotocol/sdk": "^1.17.2"
   },
   "dependencies": {
-    "@openai/zod": "npm:zod@3.25.40 - 3.25.67",
```
seratch, line 91: This is definitely effective; we should do this

packages/agents-core/src/agent.ts
```diff
@@ -1,4 +1,4 @@
-import type { ZodObject } from 'zod/v3';
```
seratch, line 1: Although `zod/index` refers to `zod/v3/external.js` within zod package, having `zod/v3` inside this Agents SDK causes TS compilation errors. Thus, simply having references to 'zod' is the right way to rely on zod on the SDK side.

seratch (commented):
Thank you so much @ianduvall !

PR #355, opened 2025-08-18
Upgrade underlying openai package to the latest version

seratch:
This pull request ensures Agents SDK to use the latest `openai` package.

### DIFF
--- a/packages/agents-openai/src/openaiChatCompletionsModel.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsModel.ts
@@ -149,24 +149,27 @@ export class OpenAIChatCompletionsModel implements Model {
         });
       } else if (message.tool_calls) {
         for (const tool_call of message.tool_calls) {
-          const { id: callId, ...remainingToolCallData } = tool_call;
-          const {
-            arguments: args,
-            name,
-            ...remainingFunctionData
-          } = tool_call.function;
-          output.push({
-            id: response.id,
-            type: 'function_call',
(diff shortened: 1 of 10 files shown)


packages/agents-openai/src/openaiChatCompletionsModel.ts
```diff
@@ -149,24 +149,27 @@ export class OpenAIChatCompletionsModel implements Model {
...
-            arguments: args,
-            name: name,
-            callId: callId,
-            status: 'completed',
-            providerData: {
-              ...remainingToolCallData,
-              ...remainingFunctionData,
-            },
-          });
+          if (tool_call.type === 'function') {
```
seratch, line 152: custom tools are recently introduced, but both Python and TS Agents SDK do not support the pattern for now.

dkundel-openai: Is this the same behavior in the Python one right now? I think we have an `unknown` type that we could use to return this and potentially stop the loop based on it. That way people could handle it. But if Python doesn't have a similar behavior I'm fine with it for now. Maybe logging it or something would make sense as it seems like a model misbehavior that could really only happen if someone uses prompts

seratch: I made the same change for python and for now we're skipping custom tools. We may support it in the future, but for now this is consistent.

dkundel-openai (approved):

PR #362, opened 2025-08-21
Fix a bug where responses api does not accept both outputType and verbosity parameter for gpt-5

seratch:
This pull request resolves a bug where responses api does not accept both outputType and verbosity parameter for gpt-5.


owengretzinger:
Pls 🙏

dkundel-openai (approved):

PR #363, opened 2025-08-21
Add a quick opt-in option to switch to gpt-5 and fix issues around gpt-5

seratch:
This pull request is based on https://github.com/openai/openai-agents-js/pull/362, and it is still work-in-progress.


.changeset/public-socks-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Fix a bug where responses api does not accept both outputType and verbosity parameter for gpt-5
```
seratch: as mentioned in this PR's description, this is part of https://github.com/openai/openai-agents-js/pull/362; so merge the PR first

examples/mcp/hosted-mcp-human-in-the-loop.ts
```diff
@@ -21,7 +21,8 @@ async function main(verbose: boolean, stream: boolean): Promise<void> {
   };
   const agent = new Agent({
     name: 'MCP Assistant',
-    instructions: 'You must always use the MCP tools to answer questions.',
+    instructions:
+      'You must always use the MCP tools to answer questions. The mcp server knows which repo to investigate, so you do not need to ask the user about it.',
```
seratch, line 25: gpt-5 models tend to be confused with this MCP server tools' rule (having the repo name in the MCP server URL), so added additional instructions. This works for both gpt-4 and gpt-5.

examples/mcp/streamable-http-custom-fetch-example.ts
```diff
@@ -1,49 +0,0 @@
-import { Agent, run, MCPServerStreamableHttp, withTrace } from '@openai/agents';
```
seratch, line 1: this example code no longer works so deleted it; we may have a working example again, but this is just demonstrating MCP hooks, so it's not our priority

examples/tools/file-search.ts
```diff
@@ -1,11 +1,32 @@
...
+    vectorStore.id,
+    { file_id: upload.id },
+  );
+  console.log(indexed);
+
   const agent = new Agent({
     name: 'File searcher',
     instructions: 'You are a helpful agent.',
     tools: [
-      fileSearchTool(['vs_67bf88953f748191be42b462090e53e7'], {
```
seratch, line 8: replaced a specific vector store id

examples/tools/image-generation.ts
```diff
@@ -25,7 +25,7 @@ async function main() {
     console.log('Generating image, this may take a while...');
     const result = await run(
       agent,
-      'Create an image of a frog eating a pizza, comic book style.',
+      'Create an image of a frog eating a pizza, comic book style. Return a text description of the image as a message too.',
```
seratch, line 28:
```suggestion
      'Create an image of a frog eating a pizza, comic book style.',
```

packages/agents-core/src/agent.ts
```diff
@@ -359,6 +366,23 @@ export class Agent<
     this.toolUseBehavior = config.toolUseBehavior ?? 'run_llm_again';
     this.resetToolChoice = config.resetToolChoice ?? true;
 
+    if (
```
seratch, line 369: I image reviewers may think "Is this necessary?" but it's necessary for smoother developer experience. See also: https://github.com/openai/openai-agents-python/pull/1534

packages/agents-core/src/run.ts
```diff
@@ -1029,3 +1051,35 @@ export async function run<TAgent extends Agent<any, any>, TContext = undefined>(
     return await runner.run(agent, input, options);
   }
 }
+
+/**
```
seratch, line 1055: same as above; without this logic, having non-gpt-5 model on the Runner side fails with OPENAI_DEFAULT_MODEL=gpt-5; this SDK can do further instead of asking developers ot be careful.

packages/agents-core/src/runImplementation.ts
```diff
@@ -573,7 +573,7 @@ export async function executeToolsAndSideEffects<TContext>(
       : undefined;
 
   // if there is no output we just run again
-  if (!potentialFinalOutput) {
+  if (typeof potentialFinalOutput === 'undefined') {
```
seratch, line 576: Changed this because GPT-5 models return empty message output text when returning image generation tool output. Checking if the output is undefined is actually intended here.

dkundel-openai (approved):

PR #364, opened 2025-08-21
Update models document page to cover gpt-5

seratch:
This document update needs to be merged after releasing https://github.com/openai/openai-agents-js/pull/363 changes.

### DIFF
--- a/docs/src/content/docs/guides/models.mdx
+++ b/docs/src/content/docs/guides/models.mdx
@@ -29,6 +29,53 @@ In day‑to‑day work you normally only interact with model **names** and occas
   title="Specifying a model per‑agent"
 />
 
+## Default model
+
+When you don't specify a model when initializing an `Agent`, the default model will be used. The default is currently [`gpt-4.1`](https://platform.openai.com/docs/models/gpt-4.1), which offers a strong balance of predictability for agentic workflows and low latency.
+
+If you want to switch to other models like [`gpt-5`](https://platform.openai.com/docs/models/gpt-5), there are two ways to configure your agents.
+
+First, if you want to consistently use a specific model for all agents that do not set a custom model, set the `OPENAI_DEFAULT_MODEL` environment variable before running your agents.
+
+```bash
+export OPENAI_DEFAULT_MODEL=gpt-5
+node my-awesome-agent.js
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/models.mdx
```diff
@@ -29,6 +29,53 @@ In day‑to‑day work you normally only interact with model **names** and occas
...
+When you use any of GPT-5's reasoning models ([`gpt-5`](https://platform.openai.com/docs/models/gpt-5), [`gpt-5-mini`](https://platform.openai.com/docs/models/gpt-5-mini), or [`gpt-5-nano`](https://platform.openai.com/docs/models/gpt-5-nano)) this way, the SDK applies sensible `modelSettings` by default. Specifically, it sets both `reasoning.effort` and `verbosity` to `"low"`. To adjust the reasoning effort for the default model, pass your own `modelSettings`:
+
+```ts
+import { Agent } from '@openai/agents';
+
+const myAgent = new Agent({
+  name: 'My Agent',
+  instructions: "You're a helpful agent.",
+  modelSettings: {
+    reasoning: { effort: 'minimal' },
```
dkundel-openai, line 64: I think we should also encourage people to try out `medium` and `high`. Just always optimizing for latency doesn't seem like an ideal recommendation. We already dumb down the model with `low` to make the latency closer to what people expect from non-reasoning models. However, just because that's the fastest and an okay default doesn't mean it's what people should strive for.

docs/src/content/docs/guides/models.mdx
```diff
@@ -79,6 +116,8 @@ The OpenAI provider defaults to `gpt‑4o`. Override per agent or globally:
 | `truncation`        | `'auto' \| 'disabled'`                     | Token truncation strategy.                                                |
 | `maxTokens`         | `number`                                   | Maximum tokens in the response.                                           |
 | `store`             | `boolean`                                  | Persist the response for retrieval / RAG workflows.                       |
+| `reasoning.effort`  | `'minimal' \| 'low' \| 'medium' \| 'high'` | Reasoning effort for gpt-5 etc.                                           |
```
dkundel-openai, line 119: In your other code samples from other PRs these were still under providerDetails. If they are top level I think that's better. Just calling it out.

Also `GPT-5` should be capitalized in here

dkundel-openai (approved):

Issue #316, opened 2025-08-09
Error `Invalid output type` should provide additional agent state and output information


AndyRightNow:
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/)
- **Have you searched for related issues?** Others may have had similar requests

### Describe the feature

Here this line 
https://github.com/openai/openai-agents-js/blob/0858c98ee2461774610bd3d7b9da15c6c0620d0f/packages/agents-core/src/runImplementation.ts#L614

It simply throws an empty error without any additional information, such as what is the potential output and what is the parsing error. The trace that this run generates also doesn't really contain much information, just `ZodError`.

It would be really helpful if any error after receiving a potential output can have all the states so far tracked not only in the error thrown but also the trace and span stats.

seratch:
Fair enough, the message string of this exception can have more helpful information.

KostaGPT:
Relatedly, can we include some additional information so we can distinguish model refusals?

Currently, if the agent output type is structured output and there's a model refusal, we get a generic `ModelBehaviorError`.

Or is there another way to tell if we got a refusal in such cases?

PR #370, opened 2025-08-25
Fix: update error message

NiteshSingh17:
fixes: https://github.com/openai/openai-agents-js/issues/316

### DIFF
--- a/packages/agents-core/src/runImplementation.ts
+++ b/packages/agents-core/src/runImplementation.ts
@@ -605,13 +605,14 @@ export async function executeToolsAndSideEffects<TContext>(
     );
     const [error] = await safeExecute(() => parser(potentialFinalOutput));
     if (error) {
+      const errorMessage = String(error);
       addErrorToCurrentSpan({
         message: 'Invalid output type',
         data: {
-          error: String(error),
+          error: errorMessage,
         },
       });
-      throw new ModelBehaviorError('Invalid output type');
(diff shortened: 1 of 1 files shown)


packages/agents-core/src/runImplementation.ts
```diff
@@ -605,13 +605,14 @@ export async function executeToolsAndSideEffects<TContext>(
...
+      const errorMessage = String(error);
       addErrorToCurrentSpan({
         message: 'Invalid output type',
         data: {
-          error: String(error),
+          error: errorMessage,
         },
       });
-      throw new ModelBehaviorError('Invalid output type');
+      throw new ModelBehaviorError(errorMessage);
```
seratch, line 615: Just replacing this way may not be helpful enough for developers. We need to check how the actual dev experience looks like.

NiteshSingh17: I think there will be only single error which will be related to JSON.parse which will be understandable.

https://github.com/NiteshSingh17/openai-agents-js/blob/ee31d368be93e09b8d80e6ec100c4d2a5dc6d185/packages/agents-core/src/utils/tools.ts#L48C19-L48C32

SkinnnyJay:
@seratch, let me know if you need me to pick this up. cc: @NiteshSingh17

SkinnnyJay:
@seratch would it be helpful if I took a look at this issue and circled back with a new PR?

PR #399, opened 2025-08-28
Upgrade openai package to the latest patch version

seratch:
This upgrade is necessary to have a few new features.


examples/docs/package.json
```diff
@@ -7,7 +7,6 @@
     "@openai/agents-core": "workspace:*",
     "@openai/agents-extensions": "workspace:*",
     "@openai/agents-realtime": "workspace:*",
-    "openai": "^5.12.2",
```
seratch, line 10: this is actually unnecessary

PR #406, opened 2025-08-28
Fix CURL request for ephemeral key

sultano:
I get "Unknown parameter: 'model' and 'type'" making the request to get an ephemeral key for the Realtime API. Omitting the request body gave me the key I needed.

API Responses
```json
{
  "error": {
    "message": "Unknown parameter: 'type'.",
    "type": "invalid_request_error",
    "param": "type",
    "code": "unknown_parameter"
  }
}
```

```json
{
  "error": {
    "message": "Unknown parameter: 'model'.",
    "type": "invalid_request_error",
    "param": "model",
    "code": "unknown_parameter"
  }
}
```


dkundel-openai:
Thanks the issue was actually the formatting so fixed that instead

PR #413, opened 2025-08-30
feat: Fix #412 add optional details data to function tool execution

### LINKED ISSUE #412: Expose tool call id to tool function for corellating with streaming items
To optimize latency we are hooking up to streaming tool call parameter items from the model to frontload some processing (e.g. if tool arguments have an array we could process items from this array as they stream in). To correlate this processing back to tool call function invocation happening later we would need tool call id which is currently not exposed to tool call function.

Could we please expose call id as a third FunctionTool.invoke argument?

```
export type FunctionTool<
  Context = UnknownContext,
  TParameters extends ToolInputParameters = undefined,
  Result = unknown,
> = {

...

  /**
   * The function to invoke when the tool is called.
   */
  invoke: (
    runContext: RunContext<Context>,
    input: string,
    callId: string,
  ) => Promise<string | Result>;

...

};
```

I am happy to draft a PR, but wanted to discuss first. Thanks!

seratch:
The feature request #412 is valid, so we may want to add further information to function tool execution if a developer needs. This changes do not break any existing apps.

### DIFF
--- a/packages/agents-core/src/tool.ts
+++ b/packages/agents-core/src/tool.ts
@@ -17,6 +17,7 @@ import { getCurrentSpan } from './tracing';
 import { RunToolApprovalItem, RunToolCallOutputItem } from './items';
 import { toSmartString } from './utils/smartString';
 import * as ProviderData from './types/providerData';
+import * as protocol from './types/protocol';
 
 /**
  * A function that determines if a tool call should be approved.
@@ -67,6 +68,7 @@ export type FunctionTool<
   invoke: (
     runContext: RunContext<Context>,
     input: string,
+    details: { toolCall: protocol.FunctionCallItem },
   ) => Promise<string | Result>;
 
   /**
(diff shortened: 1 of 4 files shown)


packages/agents-core/src/tool.ts
```diff
@@ -67,6 +68,7 @@ export type FunctionTool<
   invoke: (
     runContext: RunContext<Context>,
     input: string,
+    details?: { toolCall: protocol.FunctionCallItem },
```
seratch, line 71: if we don't want to expose all the tool call data, I can filter this to have only callId

chernetsov: having FunctionCallItem is actually great 👍

chernetsov: Curious why do we need details to be optional?

dkundel-openai (approved):

PR #421, opened 2025-09-02
feat: New application security agent example

dpitt:
Added a new Agent example that utilizes the SDK agent framework to obtain and determine OWASP application security vulnerabilities and report findings.   Files are read from the current directory and a timestamped markdown report will be generated in the reports folder. An example report generated from the agent is included.


seratch:
Thanks for sending this example! However, we've been trying to keep the examples manageable for our team, so we're unable to accept this example. Sharing this in your own repo is greatly appreciated!

dpitt:
Will do thanks for the quick response.

PR #499, opened 2025-09-22
fix: non-OpenAI models w/ OpenRouter send an empty string when calling tools

ertembiyik:
## Summary
Some models on OpenRouter called tools with empty object params using "" instead of {}, when headed to traces, it's clear that agent is calling tools with empty args:
```
{
  "callId": "toolu_vrtx_019cputrBMsRPi8nKmGGjS1y",
  "name": "LINEAR_LIST_LINEAR_PROJECTS",
  "type": "function_call",
  "arguments": "",
  "providerData": {
    ...
  },
  "status": "completed"
}
```

This results in error:
```
ToolCallError: Failed to run function tools: SyntaxError: Unexpected end of JSON input
```
## Example
Here is a minimal reproducible example:

[agents-bug.zip](https://github.com/user-attachments/files/22465905/agents-bug.zip)

```typescript
import { Agent, run } from "@openai/agents";
import { aisdk } from "@openai/agents-extensions";
import { createOpenRouter } from "@openrouter/ai-sdk-provider";
import { Composio } from "@composio/core";
import { OpenAIAgentsProvider } from "@composio/openai-agents";

// Replace these placeholder values with your real credentials before running the example.
const OPENROUTER_API_KEY = "";
const COMPOSIO_API_KEY = "";
const COMPOSIO_USER_ID = "";

async function main() {
    if (
        [OPENROUTER_API_KEY, COMPOSIO_API_KEY, COMPOSIO_USER_ID].some((value) =>
            value === ""
        )
    ) {
        throw new Error("Update the API key placeholders before running this example.");
    }
    const composio = new Composio({
        apiKey: COMPOSIO_API_KEY,
        provider: new OpenAIAgentsProvider({ strict: true }),
    });
    const linearTool = await composio.tools.get(COMPOSIO_USER_ID, {
        toolkits: ["LINEAR"]
    });
    const openrouter = createOpenRouter({
        apiKey: OPENROUTER_API_KEY,
    });
    const model = aisdk(openrouter("anthropic/claude-sonnet-4"));
    const agent = new Agent({
        name: "Linear Projects Agent",
        instructions: "Use the Linear tool to list projects and summarize them for the user.",
        model,
        tools: linearTool,
    });
    console.log("Running agent with LINEAR_LIST_LINEAR_PROJECTS...");
    const result = await run(agent, "List all projects in the connected Linear workspace.");
    console.log("Final agent output:", result.finalOutput ?? "No final output.");
}

main().catch((error) => {
    console.error("Agent execution failed:", error);
});
```

1. Replace variable at the top with real values
2. Run install and dev scripts:
```
pnpm i
pnpm run dev
```
## Changes
- Compare the tool params type from schema to the actual call and map "" to {} if needed 
## Tests
- New tests validate both tools and handoffs for mapping "" to {} where appropriate


ertembiyik:
Related to https://github.com/vercel/ai/issues/6687

seratch:
Thanks for sending this, but I am unsure if this should be done on this SDK side.

walkerheathr24-rgb:
Thank you for all of the help! I wasn't able to get to it sooner due have a procedure. It is the 9th one this year so that I really appreciate all of the support on this!

seratch:
@ertembiyik I haven't checked this for a while, but doesn't the latest ai-sdk still handle this pattern in a good way? If yes, we may consider having this workaround on our side.

ertembiyik:
> @ertembiyik I haven't checked this for a while, but doesn't the latest ai-sdk still handle this pattern in a good way? If yes, we may consider having this workaround on our side.

hey! let me check it today with ai-sdk/provider version 3 and come back

ertembiyik:
upd: i have tested the issue against ai sdk v6.0.0-beta.72 and @ai-sdk/provider v3.0.0-beta.8

the issue is still reproducible, i have simplified the example to not include composio, im now using a tool from @openai/agents with an empty object

important to mention that this issue seems to be reproducible on anthropic/claude-sonnet-4 model from openrouter, though on x-ai/grok-4-fast it seems to work just fine

here is the example i used: 
[agents-bug-ai-sdk-v6.zip](https://github.com/user-attachments/files/23142200/agents-bug-ai-sdk-v6.zip)

ertembiyik:
~~but i believe you are right that this is not the place for such patch, it should be done in openrouter sdk or ai sdk itself i guess~~

did one more test, where i took pure ai sdk solution w/ openrouter without relying on agents sdk. it did fine for anthropic/claude-sonnet-4 in the exact same empty object case

here it is: 
[agents-bug-pure-ai-sdk.zip](https://github.com/user-attachments/files/23142366/agents-bug-pure-ai-sdk.zip)

with that said, seems like the agents sdk is the source of the issue

.changeset/quick-frogs-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-extensions': minor
```
seratch:
```suggestion
'@openai/agents-extensions': patch
```

seratch:
Thank you so much for taking time for further investigation. This should be valuable for other users too, so we're fine to have this change. If we see regression after releasing this change (I don't suppose at this moment though), we may revert the change.

pReya:
@ertembiyik @seratch This still seems to fail in my case (I'm not using OpenRouter but `vercel/ai` – but the underlying problem is identical). I'm using latest release (0.3.2) on both packages (`@openai/agents` and `@openai/agents-extensions`) and I still get `SyntaxError: Unexpected end of JSON input` in `runImplementation.ts` when using a tool call (`toolRun.toolCall`) that looks like this:

```json
{
  "type": "function_call",
  "callId": "toolu_01XabcWXWkJrR7iTECAJxwX",
  "name": "list_teams",
  "status": "completed",
  "arguments": ""
}
```

Seems this patch here only fixed some cases, but not all of them?!

ertembiyik:
hey, @pReya! 
could you please include a minimal reproducible example so i can take a look at it?

PR #510, opened 2025-09-25
fix(randomUUID): add fallback when crypto.randomUUID is unavailable

ziyak97:
In some environments (non-secure origins like http, older browsers, certain test runners), `crypto.randomUUID` is not available, causing runtime errors.

This change preserves the existing fast path when `crypto.randomUUID` exists, and adds a RFC4122 v4-shaped fallback for other cases:

- Detects availability via: typeof crypto !== 'undefined' && crypto.randomUUID
- Uses `crypto.randomUUID.bind(crypto)` when present
- Otherwise formats a UUID using 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(...) with Math.random, ensuring version (4) and variant (8|9|a|b) bits
- Keeps strict typing with template-literal UUID type for DX

No breaking changes.


seratch:
@ziyak97 This looks great. Could you add some unit tests to help with long-term maintenance?

ziyak97:
Done @seratch

seratch (approved):
LGTM; thank you!

PR #471, opened 2025-09-14
feat(mcp): support structuredContent via useStructuredContent; return full CallToolResult

SkinnnyJay:
## Summary

- Adds support for `structuredContent` in MCP tool results.
- Returns full `CallToolResult` from `MCPServer#callTool` (previously returned `content[]`).
- Adds `useStructuredContent` option to MCP server constructors (stdio, streamable-http, SSE), default `false` to preserve backward compatibility.
- Updates `mcpToFunctionTool` to include `structuredContent` in tool outputs when enabled.
- Closes #468.
- Parity with Python feature: openai/openai-agents-python#1150.
- Fixes TypeScript compilation errors (TS2352) in MCP server shims that prevented builds from completing.

## Motivation

Previously, MCP server `callTool` only returned `result.content`. If a tool returned only `structuredOutput`, agents saw an empty array. This PR brings parity with the Python SDK by exposing `structuredContent` and providing a toggle to control inclusion in tool outputs.

## Changes

- API: `MCPServer#callTool(...)` now returns `CallToolResult` (full object: `{ content, structuredContent? }`).
- Option: `useStructuredContent?: boolean` added to all MCP server options and stored on base classes. Default `false` to avoid duplicate data when servers include structured output inside `content`.
- Composition in `mcpToFunctionTool`:
  - Default (flag off):
    - Single `content` item → return that item object
    - Multiple `content` items → return array of items
  - With `useStructuredContent` true:
    - Single `content` item → return `[contentItem, structuredContent]`
    - Multiple `content` items → append `structuredContent` to the array
    - No `content` → return `structuredContent` object
- No stringification: We keep JS outputs as objects/arrays to avoid forcing downstream `JSON.parse`. This differs from Python’s string concatenation but preserves capability parity and improves ergonomics/perf in JS.

### Before
```ts
// Returned only content[]
const content = await server.callTool('tool', args); // [] when tool only returned structuredOutput
```

### After
```ts
// Returns full object
const result = await server.callTool('tool', args);
// result: { content: [...], structuredContent?: any }
```

```ts
// Default (useStructuredContent: false)
// Single content → object; multiple → array
```

```ts
// With useStructuredContent: true
const server = new MCPServerStdio({ command: 'my-mcp', useStructuredContent: true });
// Single content + structured → [contentItem, structuredContent]
// Only structured → structuredContent object
```

## Tests

- Added `packages/agents-core/test/mcpStructuredContent.test.ts`:
  - Default off: single content returns the item object.
  - With flag true:
[...]

PR #553, opened 2025-10-01
Revert "feat(mcp): support structuredContent via useStructuredContent; return full CallToolResult"

seratch:
We've received reports that the latest MCP server tool support does not work for certain use cases, so this pull request reverts openai/openai-agents-js#471 ; We're still interested in enhancing the internals this way, but we'll hold off making the changes again until we figure the cause out.


SkinnnyJay:
@seratch Curious about? Any examples? We could build some tests to assert these and other use cases.

> We've received reports that the latest MCP server tool support does not work for certain use cases

PR #574, opened 2025-10-10
Update guardrails.mdx

virakngauv:
Update typo that incorrectly explained that output guardrails get the same input as input guardrails (i.e. same as the agent input) instead of from the agent output.

### DIFF
--- a/docs/src/content/docs/guides/guardrails.mdx
+++ b/docs/src/content/docs/guides/guardrails.mdx
@@ -27,9 +27,9 @@ Input guardrails run in three steps:
 
 ## Output guardrails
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/guardrails.mdx
```diff
@@ -27,9 +27,9 @@ Input guardrails run in three steps:
 
 ## Output guardrails
 
-Output guardrails follow the same pattern:
+Output guardrails run in 3 steps:
```
virakngauv, line 30: Used the python docs as a guide. https://openai.github.io/openai-agents-python/guardrails/

PR #575, opened 2025-10-10
Wait for streaming to complete before ending traces in async mode

### LINKED ISSUE #570: Tracing: onTraceEnd called before all spans finished when running agent with streaming: true
### Describe the bug

When streaming is true, `onTraceEnd` is called right after the `onTraceStart`, it does not wait for the stream to be finished. This causes issues with 3rd party tracing connectors (e.g. [braintrust](https://github.com/braintrustdata/braintrust-sdk/blob/d6278e852f56fdb00ee5b9d97226be1589853166/integrations/openai-agents-js/src/index.ts#L172-L189)) that rely on `onTraceEnd` to come after all spans in the run have finished.

When `stream: false`, this works as expected: onTraceEnd is called when the whole `run` calls has finished.

The `run` call, when streaming, does not block, so this await'd fn returns pretty immediately, right after the first `onSpanStart`

https://github.com/openai/openai-agents-js/blob/34a9528e3b7e52addaae36a0c4316e52c248996e/packages/agents-core/src/tracing/context.ts#L59-L61

A workaround seems to be wrapping our own `withTrace` around the run call, but not sure how that interacts with the internal withTrace calls...

Workarounds welcome!

### Debug information

- Agents SDK version: (e.g. `v0.1.8`)
- Runtime environment (e.g. `Node.js 22.15.0`)

### Repro steps

```js
class MyCustomProcessor implements TracingProcessor {
  async onTraceStart(trace: Trace) {
    console.log(`Trace started: ${trace.name}`);
  }

  async onSpanStart(span: Span) {
    const spanData = span.spanData;
    if (spanData?.type === 'generation') {
      console.log(`  LLM Generation started for prompt: "${spanData.input}"`);
    } else if (spanData?.type === 'function') {
      console.log(`  Tool call started for function: ${spanData.name}`);
    }
  }

  async onSpanEnd(span: Span) {
    const spanData = span.spanData;
    if (spanData?.type === 'generation' && spanData.output) {
      console.log(`  LLM Generation ended with output: "${spanData.output}"`);
    }
  }

  async onTraceEnd(trace: Trace) {
    console.log(`Trace ended: ${trace.name}`);
    console.log(`Trace ID: ${trace.traceId}`);
  }

  async shutdown(timeout?: number) {}

  async forceFlush() {}
}

// use it
setTraceProcessors([new MyCustomProcessor()]);
```

### Expected behavior

My expectation is that the streaming behavior follows the stream: false behavior that `onTraceEnd` is called after the whole stream has been consumed and the whole `run` call is over

CLowbrow:
## Description

Fixes a bug where `onTraceEnd` was called immediately after `onTraceStart` when streaming is enabled, without waiting for the stream loop to complete.

## Behavior

**Before**: When `stream: true`, `onTraceEnd` was called immediately after the `run()` call returned, before spans ended.

**After**: When `stream: true`, `onTraceEnd` is called asynchronously after the stream loop completes and all spans have ended. The `run()` call still returns immediately (non-blocking).

Closes #570


seratch:
Thanks for sending this! At a glance, it looks good to me, but I will take a closer look early next week.

Issue #538, opened 2025-09-27
Realtime API response.done events missing detailed usage fields


ginzlabs:
`response.done ` events from the Realtime API do not include the full token usage breakdown shown in the Playground and documentation. Only total input and output counts are returned.

Currently, the event contains only:

```
"usage": {
  "input_tokens": 132,
  "output_tokens": 121
}
```

But the Playground/docs show a full structure:

```
"usage": {
  "total_tokens": 253,
  "input_tokens": 132,
  "output_tokens": 121,
  "input_token_details": {
    "text_tokens": 119,
    "audio_tokens": 13,
    "image_tokens": 0,
    "cached_tokens": 64,
    "cached_tokens_details": {
      "text_tokens": 64,
      "audio_tokens": 0,
      "image_tokens": 0
    }
  },
  "output_token_details": {
    "text_tokens": 30,
    "audio_tokens": 91
  }
}
```

Tested with all realtime models by inspecting raw event data. The detailed fields are never included.

**Expected behavior:**` response.done` should return the complete usage object as documented.

ginzlabs:
I tracked down the root cause — the problem is in `packages/agents-realtime/openaiRealtimeEvents.ts` with the Zod schema validation.

The schema is currently expecting `input_tokens_details `/ `output_tokens_details `(plural tokens), but the Realtime WebSocket API actually returns `input_token_details `/ `output_token_details `(singular token).

Because of this mismatch, the validation fails and strips out the detailed usage fields, which is why only the minimal input_tokens / output_tokens values are coming through in response.done.

leecjohnny:
I also face this issue, as a workaround, I'm reading the events from the webrtc datachannel directly.

PR #578, opened 2025-10-12
fix: resolve usage details validation in realtime events

Alejandro-AP00:
### Summary
Resolves  https://github.com/openai/openai-agents-js/issues/538

### Test plan
I used the realtime-demo example and console log the results, without this fix the details are empty objects.


.changeset/busy-trains-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-realtime': minor
```
seratch:
```suggestion
'@openai/agents-realtime': patch
```

Issue #580, opened 2025-10-13
"No existing trace found" error when using MCP servers with Websocket RealtimeAgent


AlexandreBouju:
### Describe the bug
From my understanding of this [issue](https://github.com/openai/openai-agents-js/issues/353) and #402, it should now be possible to provide an MCPServerStreamableHttp to a RealtimeAgent directly via the `mcpServers` property.

But, when initializing a RealtimeAgent with an MCP server, the application crashes with an unhandled promise rejection: `Error: No existing trace found`. The error occurs during the agent setup process when trying to connect MCP servers to gather tools.

### Debug information
- **Agents SDK version**:
  - `@openai/agents`: v0.1.9
  - `@openai/agents-realtime`: v0.1.8
- **Runtime environment**: Node.js v22.20.0
- **Operating System**: Linux
- **TypeScript version**: 5.9.3

## Error Stack Trace
```
Unhandled rejection Error: No existing trace found
    at withNewSpanContext (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/tracing/context.ts:182:11)
    at /home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/tracing/createSpans.ts:36:30
    at getFunctionToolsFromServer (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/mcp.ts:317:30)
    at getAllMcpTools (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/mcp.ts:426:31)
    at RealtimeAgent.getMcpTools (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/agent.ts:630:28)
    at RealtimeAgent.getAllTools (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-core/src/agent.ts:649:33)
    at RealtimeSession.#setCurrentAgent (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-realtime/src/realtimeSession.ts:322:65)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async RealtimeSession.connect (/home/selectra/Projects/voice-agents/openai-realtime-connector/node_modules/@openai/agents-realtime/src/realtimeSession.ts:848:5)
```

### Repro steps

```typescript
[...]

PR #581, opened 2025-10-15
Tracing to work for streamableHTTP

rgubler24:
This PR is related directly to issue #580

### DIFF
--- a/packages/agents-core/src/tracing/context.ts
+++ b/packages/agents-core/src/tracing/context.ts
@@ -118,13 +118,15 @@ export async function getOrCreateTrace<T>(
 
 /**
  * This function will set the current span in the execution context.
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this PR. However, the issue should be resolved in a different way.

rgubler24:
@seratch thank you for get back quickly. This is my first PR for this project. Could you explain how you would like it resolved? This is an issue that my company is having and would like to help get this fixed.

seratch:
I haven't figured the viable solution for this issue out yet, but removing the validation affects other scenarios than the realtime agent situation. So, we need to identify why realtime agents could go through the code path and resolve the cause of it.

rgubler24:
Understand. I'll take a deeper look later. Thank you.

Issue #445, opened 2025-09-08
Realtime agent stuck in a conversational loop after 0.1.0 update


gregory-buffard:
Hello OpenAI community :

We are facing an issue related to the Realtime Agent configuration using the Agents SDK where after the latest update (0.1.0), our Realtime Agent just runs into a loop, here's a conversation example :

> Agent : Good morning! Before starting, are you an individual or a professional?
> Agent : Good morning! Before starting, are you an individual or a professional?
> User : individual
> Agent : Good morning! Before starting, are you an individual or a professional?
> User : individual...
> Agent: Good, how may I help you ?
> User : I need ... { user clearly asks for a specific need }
> Agent : Good, how may I help you ?
> User : I need ... { user clearly asks for a specific need again }
> User : { agent doesn't respond so user asks for need again in a simple wording }
> User : { agent doesn't respond again so user asks for need again in a simple wording and then ends the conversation }

As you may conclude, the implementation is very poorly handled, the agent not only repeats itself for no apparent reason and also sometimes doesn't event answer.

We thought the problem might come from the prompt, which we have altered on many iterations but the issue doesn't seem to stem from here since this problem is also observed on many very different prompts. [Here's the example of the previous conversation prompt used for the agent](https://github.com/user-attachments/files/22206536/prompt.txt) (the prompt is translated from French, apologies for possible confusion in wording).

As a result, we have rewinded to the previous 0.17.0 version, which seem to work okay [**although it's not really good neither in terms of quality of speech and intelligence compared to the experience we have access to in the playground or the app strangely... (e.g., lack of emotions in the voice, "fake breathing", etc.")**].
Additionally, the issue seems to persist even on 0.17.0 only when we select the newly released `gpt-realtime` model.

We're out of options on how to fix this issue, hence why I'm turning towards the Agent SDK community.
Finally, our very simple code implementation of the agent can be found in the [openai.ts](https://github.com/user-attachments/files/22206469/openai.ts) file.

Does anyone else face a similar issue ? If so, have you managed to resolve it ? How ?
We'll enormously appreciate any kind of help on this matter since we're desperately out of possibilities...

Do not hesitate to let me know if any additional code resources are needed for better understanding of the matter.

gregory-buffard:
Update : The loop problem seems not to be occurring on the new 0.1.1 when using legacy models (i.e., GPT 4o Realtime and GPT 4o-mini Realtime) but persists on the new _GPT Realtime_ (`gpt-realtime`)

[...]

PR #591, opened 2025-10-19
feat: allow audio resampling for twilio transport layer

Timoyoungster:
This adds the option to resample the audio between twilio and realtime, in hopes that by resampling the 8kHz from twilio to 24kHz the looping behaviour from #445 can be alleviated.

I was unable to test whether this actually helps, because I gave up trying after 2 days of struggling with various import errors related to the _shims directories.


Timoyoungster:
I have now managed to test how resampling affects model behavior by implementing the resampling functions directly in the compiled/installed node package.

Unfortunately it did not improve anything, so this pr will not help with regard to #445.

Nevertheless, I think the functionality to process the audio prior to forwarding it could still be useful in some scenarios.

seratch:
Thanks for doing further investigation. We don't plan to add this resampling option at this moment, but we'll continue seeing if people need it.

PR #602, opened 2025-10-23
feat: #313 Enable tools to return image/file data to an Agent

### LINKED ISSUE #313: Ability to return images/files from tool calls
I'm working on a refactor of my company's codebase to this sdk, but one blocker on our end is that there isn't a way for tool calls to return files.

The way our current code handles this is that tools can return either a string (in which case it works the same as it currently does) or an array of messages along with a string as a structured response. The array of messages gets inserted into the conversation as user-roled messages.

Is this something you guys would be open to having?

seratch:
This pull request resolves #313 and aligns with Python SDK's changes: https://github.com/openai/openai-agents-python/pull/1898

- Image Output: https://github.com/openai/openai-agents-js/blob/feea53ae522a629d02fba64a4103cede5ff72a17/examples/ai-sdk/image-tool-output.ts
- File Output: https://github.com/openai/openai-agents-js/blob/feea53ae522a629d02fba64a4103cede5ff72a17/examples/basic/file-tool-output.ts

Note that data protocol in TS SDK is different (this SDK has "protocol"), so the changes here are not exactly the same with Python SDK. That said, the direction and functionalities should be the same.

### DIFF
--- a/examples/docs/handoffs/customizeHandoff.ts
+++ b/examples/docs/handoffs/customizeHandoff.ts
@@ -1,13 +1,17 @@
+import { z } from 'zod';
 import { Agent, handoff, RunContext } from '@openai/agents';
 
-function onHandoff(ctx: RunContext) {
-  console.log('Handoff called');
+const FooSchema = z.object({ foo: z.string() });
+
+function onHandoff(ctx: RunContext, input?: { foo: string }) {
+  console.log('Handoff called with:', input?.foo);
 }
 
 const agent = new Agent({ name: 'My agent' });
 
 const handoffObj = handoff(agent, {
   onHandoff,
+  inputType: FooSchema,
   toolNameOverride: 'custom_handoff_tool',
   toolDescriptionOverride: 'Custom description',
 });


packages/agents-core/src/types/protocol.ts
```diff
@@ -167,20 +176,99 @@ export const ToolOutputText = SharedBase.extend({
...
+    FileIdObjectSchema,
+  ])
+  .describe(
+    'Inline data (with metadata) or references pointing to file contents.',
+  );
+
+const zStringWithHints = <T extends string>(..._hints: T[]) =>
+  z.string() as unknown as z.ZodType<T | (string & {})>;
+
 export const ToolOutputImage = SharedBase.extend({
```
seratch, line 242: This has been available since the initial release, but actually there had been no way to utilize this type. That said, we're going to change the data structure to fit the reality, so I'd go with a minor version bump.

.changeset/bumpy-papayas-drum.md
```diff
@@ -0,0 +1,7 @@
+---
+'@openai/agents-extensions': minor
```
seratch, line 2: This is intentionally minor bump because this PR changes the accidentally publicly available "unused" ToolOutputImage data structure. See my comment on the protocol file.

packages/agents-core/src/types/protocol.ts
```diff
@@ -167,20 +176,99 @@ export const ToolOutputText = SharedBase.extend({
...
+  .union([ImageDataObjectSchema, ImageUrlObjectSchema, ImageFileIdObjectSchema])
+  .describe('Inline image data or references to uploaded content.');
+
+const FileDataObjectSchema = z.object({
+  data: z
+    .union([z.string(), z.instanceof(Uint8Array)])
+    .describe(
+      'Base64 encoded file data, or raw bytes that will be encoded automatically.',
+    ),
+  mediaType: z
```
seratch, line 212: this is required in reality, so intentionally non-optional

packages/agents-core/src/types/protocol.ts
```diff
@@ -167,20 +176,99 @@ export const ToolOutputText = SharedBase.extend({
...
+    .string()
+    .describe('IANA media type describing the file contents'),
+  filename: z.string().describe('Filename associated with the inline data'),
```
seratch, line 215: same as above; without this, you will get 400 error from responses api

packages/agents-core/src/utils/base64.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
seratch, line 1: adding a new dependency is an option too, but having this simple utility should be fine

packages/agents-core/src/utils/smartString.ts
```diff
@@ -1,14 +1,104 @@
+const BYTE_PREVIEW_LIMIT = 20;
+
 export function toSmartString(value: unknown): string {
   if (value === null || value === undefined) {
     return String(value);
-  } else if (typeof value === 'string') {
+  }
+
+  if (isArrayBufferLike(value)) {
+    return formatByteArray(new Uint8Array(value));
```
seratch, line 9: mainly for debug logging improvements; the payload for API calls does not have byte array data

packages/agents-core/src/runImplementation.ts
```diff
@@ -666,6 +673,22 @@ export function getToolCallOutputItem(
   toolCall: protocol.FunctionCallItem,
   output: string | unknown,
 ): FunctionCallResultItem {
+  const maybeStructuredOutputs = normalizeStructuredToolOutputs(output);
+
+  if (maybeStructuredOutputs) {
```
seratch, line 678: this is the new pattern; existing text data goes in the same way

packages/agents-core/src/runImplementation.ts
```diff
@@ -678,6 +701,408 @@ export function getToolCallOutputItem(
...
+  | ToolOutputImage
+  | ToolOutputFileContent;
+
+/**
+ * Accepts whatever the tool returned and attempts to coerce it into the structured protocol
+ * shapes we expose to downstream model adapters (input_text/input_image/input_file). Tools are
+ * allowed to return either a single structured object or an array of them; anything else falls
+ * back to the legacy string pipeline.
+ */
+function normalizeStructuredToolOutputs(
```
seratch, line 715: these internal functions may look complicated, but basically they deal with the supported patterns for better dev experience (rather than asking devs to generate data URL on their own); still it's also possible to pass data url.

packages/agents-extensions/src/aiSdk.ts
```diff
@@ -29,6 +29,7 @@ import {
   ModelSettingsToolChoice,
```
seratch, line 29: it seems many of the changes in this file should be revisited; i am working on this now

seratch: Fixed

packages/agents-core/src/utils/base64.ts
```diff
@@ -0,0 +1,50 @@
...
+    const enc2 = ((c1 & 0x3) << 4) | (c2 >> 4);
+    const enc3 = isNaN(c2) ? 64 : ((c2 & 0xf) << 2) | (c3 >> 6);
+    const enc4 = isNaN(c3) ? 64 : c3 & 0x3f;
+
+    result +=
+      chars.charAt(enc1) +
+      chars.charAt(enc2) +
+      chars.charAt(enc3) +
+      chars.charAt(enc4);
+  }
```
dkundel-openai, line 47: Are there any systems that we support that don't support `btoa` or why do we need an additional manual conversion?

packages/agents-core/src/runImplementation.ts
dkundel-openai: Is this breaking things where previously a tool might have returned an array of data that we simply stringified that would now result in a conversion into image or file types?

dkundel-openai: if so do we want to instead provide something like having people return inside an execute function of a tool something like `return result(rich_data_with_image)` so that we know when to apply this logic and when not? (Not married to the name of `result`

packages/agents-core/src/utils/base64.ts
seratch: Good call. I got help from codex and there are two things to mention:

- The manual fallback is intentional. We support React Native targets, and React Native still doesn’t provide a built‑in btoa (or atob) on globalThis, so without the custom encoder any call coming from that runtime would throw.
- Node.js historically skipped btoa; even though recent versions expose it in globalThis, relying on it is fragile and the Buffer path we already have is the safer default.

The second one means btoa still exists but it's marked as legacy (not hard deprecation though): https://nodejs.org/download/release/v20.11.1/docs/api/all.html#all_globals_btoadata and the recommended alternative is to use Buffer#toString('base64')

packages/agents-core/src/runImplementation.ts
seratch: I don't think so (also double-checked with codex too). inside this function, it calls normalizeStructuredToolOutput and the internal method checks if there is type (text, image, etc.); if the output is not with any of expected types and relevant data structure, they won't converted in the new way. for safety, i've added more test patterns covering the scenarios you've mentioned here.

xtc35874:
How to return multiple images？

PR #604, opened 2025-10-23
Upgrade dependencies for dev, docs, and examples

seratch:


examples/agent-patterns/package.json
```diff
@@ -3,8 +3,7 @@
   "name": "agent-patterns",
   "dependencies": {
     "@openai/agents": "workspace:*",
-    "chalk": "^5.4.1",
-    "zod": "^3.25.40"
```
seratch, line 7: these examples are part of pnpm workspaces, so zod will come from the workspaces; having zod again here could cause conflicts, so removing these should be the correct way to manage the dependency.

PR #611, opened 2025-10-23
Fix: Handoff creation error when using `onHandoff` without `inputType`

MuhammadHamidRaza:
#### 🧩 Problem

When creating a handoff using the `handoff()` helper, the code currently throws:

```py
UserError: You must provide either both `onHandoff` and `inputType` or neither.
```

This happens because the SDK enforces that both must be provided together.  
In the docs example, `onHandoff` was used **without** an `inputType`, which triggers this validation error.

#### ⚙️ Root Cause

The docs’ example:

```py
handoff(agent, {
  onHandoff, toolNameOverride: 'custom_handoff_tool', toolDescriptionOverride: 'Custom description',
});
```

is missing the required `inputType` parameter when `onHandoff` is defined.

#### 🧠 Solution

To fix the issue, we’ll:

-   Either **remove `onHandoff`** (if no structured input is required),  
    or
    
-   **Add `inputType`** (if the handoff will receive structured data).
    

For now, we’ll go with the simple fix — **remove `onHandoff`** and only keep custom tool details.

#### ✅ Fixed Code

```py
import { Agent, handoff } from  '@openai/agents';
const agent = new  Agent({ name: 'My agent' }); 
const handoffObj = handoff(agent, { 
toolNameOverride: 'custom_handoff_tool',
toolDescriptionOverride: 'Custom description',
});
``` 

This resolves the runtime error and allows the handoff to be created successfully.

#### 🧾 Notes

-   Future handoffs that need to process structured input should define both:
    
```py
onHandoff: async (ctx, input) => { ... }, inputType: z.object({ ... })
``` 
    
-   Verified by running with `npx tsx agent/index.ts` — no errors occur now.
    

----------

### DIFF
--- a/examples/docs/handoffs/customizeHandoff.ts
+++ b/examples/docs/handoffs/customizeHandoff.ts
@@ -1,15 +1,17 @@
 import { z } from 'zod';
 import { Agent, handoff, RunContext } from '@openai/agents';
 
-function onHandoff(ctx: RunContext, input: { foo: string }) {
-  console.log('Handoff called with:', input.foo);
+const FooSchema = z.object({ foo: z.string() });
+
+function onHandoff(ctx: RunContext, input?: { foo: string }) {
+  console.log('Handoff called with:', input?.foo);
 }
 
 const agent = new Agent({ name: 'My agent' });
 
 const handoffObj = handoff(agent, {
   onHandoff,
-  inputType: z.object({ foo: z.string() }),
+  inputType: FooSchema,
   toolNameOverride: 'custom_handoff_tool',
   toolDescriptionOverride: 'Custom description',
 });


examples/docs/handoffs/customizeHandoff.ts
```diff
@@ -1,13 +1,8 @@
...
-
-function onHandoff(ctx: RunContext) {
-  console.log('Handoff called');
-}
+import { Agent, handoff } from '@openai/agents';
 
 const agent = new Agent({ name: 'My agent' });
 
 const handoffObj = handoff(agent, {
-  onHandoff,
```
seratch, line 10: deleting this makes this code snippet meaningless. can you add `inputType: z.object({ foo: z.string() })`?

seratch (changes requested):

examples/docs/handoffs/customizeHandoff.ts
MuhammadHamidRaza: now check it

Issue #613, opened 2025-10-24
OpenAIRealtimeWebRTC `connection_change` Fails to Detect Disconnection After Connection is Established


IM594:
### Please read this first

- **Have you read the docs?**: Yes
- **Have you searched for related issues?**: Yes

---

### Describe the bug

The **`connection_change`** event on `OpenAIRealtimeWebRTC` is **broken** for connection loss detection. While this event is available immediately after we instantiate `new OpenAIRealtimeWebRTC()`, it **only provides partial coverage** of the connection lifecycle.

It correctly reports transitions *before* the session is active (e.g., `"disconnected"`-> `"connecting"` -> `"connected"`), but it **fails to fire reliably** when the established connection is lost (i.e., when transitioning from `"connected"` back to `"disconnected"`, `"failed"`, or `"closed"`).

This means the agent's connection status gets **stuck at `"connected"`** in our application state, even when the underlying WebRTC connection is dead.

To get accurate disconnection events, we are **forced to manually listen** to the `RTCPeerConnection.onconnectionstatechange`. This is a cumbersome workaround because:

1.  It defeats the purpose of the SDK's abstraction provided by `OpenAIRealtimeWebRTC`.
2.  We can only reliably attach this manual monitor *after* the SDK's (flawed) event signals `"connected"`. This creates a clumsy, **nested event system** just to track basic connection status, preventing us from setting up comprehensive monitoring immediately after instantiating the WebRTC object.

The SDK's current behavior makes robust error handling impossible without this complicated workaround.

---

### Debug information

- Agents SDK version: **v0.1.11**
- Runtime environment: **Node.js v22.15.1** and **Chrome 141.0.7390.108 (arm64)**

### Repro steps

1.  Initialize and connect a `RealtimeSession` using `OpenAIRealtimeWebRTC`.
2.  Attach a listener to the SDK event: `webRTC.on("connection_change", (state) => { console.log("SDK state:", state); })`.
3.  Observe that initial states (`"disconnected"`, `"connecting"`, `"connected"`) are correctly reported.
4.  **Implement the workaround:** Use the `webRTC.on("connection_change")` event to set up the secondary `RTCPeerConnection.onconnectionstatechange` monitor once the state is `"connected"`.
5.  **Simulate disconnection:**
    * Disable your device's network connection (e.g., toggle Wi-Fi off).
[...]

PR #620, opened 2025-10-25
fix: #613 Listen to peerConnection state in `OpenAIRealtimeWebRTC` to detect disconnects

IM594:
This pull request resolves https://github.com/openai/openai-agents-js/issues/613

To resolve this issue, I added an onconnectionstatechange listener to the RTCPeerConnection instance — this is the minimum fix. If the underlying connection state transitions to 'disconnected', 'failed', or 'closed', the close() method is now invoked.

### DIFF
--- a/packages/agents-realtime/test/openaiRealtimeWebRtc.test.ts
+++ b/packages/agents-realtime/test/openaiRealtimeWebRtc.test.ts
@@ -324,6 +324,33 @@ describe('OpenAIRealtimeWebRTC.connectionState', () => {
     expect(rtc.status).toBe('disconnected');
     expect(events).toEqual(['connecting', 'connected', 'disconnected']);
   });
+
+  it('migrates connection state handler when peer connection is replaced', async () => {
+    class CustomFakePeerConnection extends FakeRTCPeerConnection {}
+    const customPC = new CustomFakePeerConnection();
+
+    const rtc = new OpenAIRealtimeWebRTC({
+      changePeerConnection: async () => customPC as any,
+    });
+
+    const closeSpy = vi.spyOn(rtc, 'close');
+    const events: string[] = [];
+    rtc.on('connection_change', (status) => events.push(status));
+
+    await rtc.connect({ apiKey: 'ek_test' });
+
+    expect(rtc.status).toBe('connected');
+    expect(rtc.connectionState.peerConnection).toBe(customPC as any);
+    expect(closeSpy).not.toHaveBeenCalled();
+    expect(events).toEqual(['connecting', 'connected']);
+
+    customPC._simulateStateChange('failed');
+    await new Promise((resolve) => setTimeout(resolve, 0));
+
+    expect(closeSpy).toHaveBeenCalled();
+    expect(rtc.status).toBe('disconnected');
+    expect(events).toEqual(['connecting', 'connected', 'disconnected']);
+  });
 });
 
 describe('OpenAIRealtimeWebRTC.callId', () => {


IM594:
https://github.com/user-attachments/assets/ce6af3f7-e087-4a23-be19-5d74407ba845


https://github.com/user-attachments/assets/df4105cd-b28f-4627-99c0-5662825b4447

seratch:
Thanks for sending this patch! I actually also did something similar, so I think this should be fine. I will look into details and do testing on my end too.

seratch:
@codex review

seratch:
@IM594 I asked my local codex to review the changes again, and it says:

```
• Review comment:

  - [P1] Rebind connectionstate handler after changePeerConnection — packages/agents-realtime/src/openaiRealtimeWebRtc.ts:263-266
    When changePeerConnection returns a new RTCPeerConnection, the onconnectionstatechange handler added just above never gets reattached to that replacement. That means the new
  peer connection will never trigger close() on failed/disconnected states, so status stays connected and no connection_change event fires—the exact regression this patch is trying
  to solve. Anyone overriding the connection (a documented capability) will still miss disconnects after this change. Please reapply the handler to the returned peer connection (and
  ideally remove it from the original one) so the auto-disconnect logic works in that scenario.
```

And this makes sense to me. The suggested changes are:

```diff
diff --git a/packages/agents-realtime/src/openaiRealtimeWebRtc.ts b/packages/agents-realtime/src/openaiRealtimeWebRtc.ts
index 2c5c661..19d5279 100644
--- a/packages/agents-realtime/src/openaiRealtimeWebRtc.ts
+++ b/packages/agents-realtime/src/openaiRealtimeWebRtc.ts
@@ -181,17 +181,22 @@ export class OpenAIRealtimeWebRTC
         const dataChannel = peerConnection.createDataChannel('oai-events');
         let callId: string | undefined = undefined;
 
-        peerConnection.onconnectionstatechange = () => {
-          switch (peerConnection.connectionState) {
-            case 'disconnected':
-            case 'failed':
-            case 'closed':
-              this.close();
-              break;
-            // 'connected' state is handled by dataChannel.onopen. So we don't need to handle it here.
-            // 'new' and 'connecting' are intermediate states and don't require action here
-          }
+        const attachConnectionStateHandler = (
+          connection: RTCPeerConnection,
+        ) => {
+          connection.onconnectionstatechange = () => {
+            switch (connection.connectionState) {
+              case 'disconnected':
+              case 'failed':
+              case 'closed':
+                this.close();
+                break;
+              // 'connected' state is handled by dataChannel.onopen. So we don't need to handle it here.
+              // 'new' and 'connecting' are intermediate states and do not require action here.
+            }
+          };
         };
+        attachConnectionStateHandler(peerConnection);
 
         this.#state = {
           status: 'connecting',
@@ -261,8 +266,13 @@ export class OpenAIRealtimeWebRTC
         peerConnection.addTrack(stream.getAudioTracks()[0]);
 
         if (this.options.changePeerConnection) {
+          const originalPeerConnection = peerConnection;
           peerConnection =
             await this.options.changePeerConnection(peerConnection);
+          if (originalPeerConnection !== peerConnection) {
+            originalPeerConnection.onconnectionstatechange = null;
+          }
+          attachConnectionStateHandler(peerConnection);
           this.#state = { ...this.#state, peerConnection };
         }

```

What do you think?

IM594:
@seratch Great catch, and thanks for pointing this out! You're right, the handler needed to be moved. I've added a test for this.

My apologies for the oversight and any extra work it caused. I really appreciate the thorough review—it's what makes the SDK more robust

seratch (approved):

PR #641, opened 2025-11-05
fix: #639 Type issue with realtime agent handoffs

### LINKED ISSUE #639: Type issue with realtime agent handoffs
Looking at the handoffs definitions of RealtimeAgent:

https://github.com/openai/openai-agents-js/blob/14016fd4d86e202911e23ed26c233aa79e59baaa/packages/agents-realtime/src/realtimeAgent.ts#L32

I'm having to ignore agents context in order to make typescript happy, I'm getting something wrong.

```ts
handoffs: specialists as unknown as RealtimeAgent<unknown>[],
```

seratch:
This pull request resolves #639; the changes in the types should not break any existing code

### DIFF
--- a/packages/agents-realtime/src/realtimeAgent.ts
+++ b/packages/agents-realtime/src/realtimeAgent.ts
@@ -9,7 +9,7 @@ import { RealtimeContextData } from './realtimeSession';
 
 export type RealtimeAgentConfiguration<TContext = UnknownContext> = Partial<
   Omit<
-    AgentConfiguration<TContext, TextOutput>,
+    AgentConfiguration<RealtimeContextData<TContext>, TextOutput>,
     | 'model'
     | 'handoffs'
     | 'modelSettings'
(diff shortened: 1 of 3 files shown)


packages/agents-realtime/src/realtimeAgent.ts
```diff
@@ -9,7 +9,7 @@ import { RealtimeContextData } from './realtimeSession';
 
 export type RealtimeAgentConfiguration<TContext = UnknownContext> = Partial<
   Omit<
-    AgentConfiguration<TContext, TextOutput>,
+    AgentConfiguration<RealtimeContextData<TContext>, TextOutput>,
```
seratch, line 12: wrapping with RealtimeContextData just appends `history` property as optional, so this is safe for both tsc and runtime.

viniciuspalma: This was great, Thank you for resolving so fast.

Issue #638, opened 2025-11-04
Usage in context is not being updated with stream


yusuf-eren:
Example code:

```typescript
import 'dotenv/config';
import { Agent, run } from '@openai/agents';
import fs from 'fs';

async function main() {
  const agent = new Agent({
    name: 'greeting-agent',
    instructions: 'You are a helpful assistant that can greet the user.',
    tools: [],
  });
  const response = await run(
    agent,
    [
      {
        role: 'user',
        content: 'Hello, how are you?',
      },
    ],
    { stream: true }
  );
  await response.completed;
  fs.writeFileSync(
    'response-with-stream.json',
    JSON.stringify(response, null, 2)
  );
}

main();

```

output:
```json
{
   "state": {
    "context": {
      "usage": {
        "requests": 0,
        "inputTokens": 0,
        "outputTokens": 0,
        "totalTokens": 0
      },
      "approvals": {},
      "context": {}
    }
   },
}
```

If stream option is disabled its ok. but if stream is enabled, it doesnt update the context.

seratch:
Hi, thanks for writing in. Indeed, the different behavior of the result object for streaming mode could be confusing, but when you go with streaming mode, your app code needs to extract data from response_done event:
```typescript
import { Agent, run, Usage } from '@openai/agents';

async function main() {
  const agent = new Agent({
    name: 'Joker',
    instructions: 'You are a helpful assistant.',
  });
  const stream = await run(agent, 'Please tell me 5 jokes.', { stream: true });
  let usage: Usage | undefined;
  for await (const event of stream) {
    if (event.type === 'raw_model_stream_event' && event.data.type === 'response_done') {
      usage = new Usage(event.data.response.usage);
    }
  }
  console.log(usage);
}

main().catch((error) => {
  console.error(error);
  process.exit(1);
});
```
Hope this helps.

rclayton-godaddy:
@seratch I ran into this issue as well when I was building a `TracingProvider` for Langfuse.  The issue is that usage data via Chat Completions API is not coming back with the span on `onSpanEnd`.  I was debugging `@openai/agents-openai/dist/openaiChatCompletionsModel.mjs` and noticed on lines 146-164 that the response object is never updated with usage data (and all values remain `0`):

```
            const stream = await this.#fetchResponse(request, span, true);
            const response = {
                id: FAKE_ID,
                created: Math.floor(Date.now() / 1000),
                model: this.#model,
                object: 'chat.completion',
                choices: [],
[...]

PR #652, opened 2025-11-13
fix: #638 Add usage data from stream on response (Chat Completions)

rclayton-godaddy:
Streaming results may not include usage data for providers like LiteLLM, causing "generation spans" (`onSpanEnd`) to not have usage results.  Adding a simple check to provide usage data from the last stream event fixes the issue.

No usage data is on the response unless it's supplied via streaming chunks (LiteLLM doesn't do this):
image

However, the last event has the usage data:
image

This PR grabs the usage data from the `response_done` event on the stream, appending it to the response.

Note:  there is a bit of friction in the model.  The `openaiChatCompletionStreaming.ts` component converts the OAI usage from "snake case" variables to Pascal case.  However, the response type requires "snake case".  I chose to not do anything fancy, preferring a direct mapping. 

Refer to: https://github.com/openai/openai-agents-js/issues/638#issuecomment-3524532698

### DIFF
--- a/.changeset/shaky-bananas-check.md
+++ b/.changeset/shaky-bananas-check.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Export usage data from Chat Completions response for trace
--- a/packages/agents-openai/src/openaiChatCompletionsModel.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsModel.ts
@@ -214,6 +214,18 @@ export class OpenAIChatCompletionsModel implements Model {
         response,
         stream,
       )) {
+        if (
+          event.type === 'response_done' &&
+          response.usage?.total_tokens === 0
+        ) {
+          response.usage = {
+            prompt_tokens: event.response.usage.inputTokens,
+            completion_tokens: event.response.usage.outputTokens,
+            total_tokens: event.response.usage.totalTokens,
+            prompt_tokens_details: event.response.usage.inputTokensDetails,
+            completion_tokens_details: event.response.usage.outputTokensDetails,
+          };
+        }
         yield event;
       }
 
--- a/packages/agents-openai/test/openaiChatCompletionsModel.test.ts
+++ b/packages/agents-openai/test/openaiChatCompletionsModel.test.ts
@@ -482,4 +482,57 @@ describe('OpenAIChatCompletionsModel', () => {
     expect(convertChatCompletionsStreamToResponses).toHaveBeenCalled();
     expect(events).toEqual([{ type: 'first' }, { type: 'second' }]);
   });
+
+  it('populates usage from response_done event when initial usage is zero', async () => {
+    // override the original implementation to add the response_done event.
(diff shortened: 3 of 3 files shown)


rclayton-godaddy:
@seratch another issue I encountered with the `OpenAIChatCompletionsModel` is that it doesn't provide the essential `SpanData` you would want with a `Generation` span.  The `response` object is never updated.  My change adds the usage, but then I realized I'm not getting back `choices`.  I fixed this locally by casting the output as `any`, which at least allows me to report the output on the span.  However, this obviously suboptimal. The `choices` model comes from OAI Chat Completions, but `convertChatCompletionsStreamToResponses` is returning `AgentOutputItem`.

The question is, should Generation SpanData be in the OAI Agent SDK model, or in the underlying model implementation's?

```typescript
 const response = {
      id: FAKE_ID,
      created: Math.floor(Date.now() / 1000),
      model: this.#model,
      object: 'chat.completion',
      choices: [],
      usage: {
          prompt_tokens: 0,
          completion_tokens: 0,
          total_tokens: 0,
      },
  };
  
  for await (const event of convertChatCompletionsStreamToResponses(response, stream)) {
      if (
          event.type === 'response_done' &&
          response.usage?.total_tokens === 0
      ) {
          response.choices = event.response.output;
          response.usage = {
              prompt_tokens: event.response.usage.inputTokens,
              completion_tokens: event.response.usage.outputTokens,
              total_tokens: event.response.usage.totalTokens,
              prompt_tokens_details: event.response.usage.inputTokensDetails,
              completion_tokens_details: event.response.usage.outputTokensDetails,
          };
      }
      yield event;
  }
  
  if (span && response && request.tracing === true) {
      span.spanData.output = [response];
  }
  ```

**Comment by seratch** [comment]
@rclayton-godaddy I've verified the changes in this PR are sufficient for Chat Completions use cases.

I am not sure about the question in your last comment, but with the changes in this PR, the span data includes usage data as expected:
```json
{
  "created": 1763096420,
  "id": "FAKE_ID",
  "usage": {
    "prompt_tokens_details": {
      "cached_tokens": 0
    },
    "completion_tokens_details": {
      "reasoning_tokens": 0
    },
    "completion_tokens": 100,
    "total_tokens": 124,
    "prompt_tokens": 24
  },
  "model": "gpt-4.1",
  "object": "chat.completion",
  "choices": []
}
```
The tracing dashboard is optimized for Responses API, so when you use Chat Completions API, the data can be found like the above JSON data.

I am fine to merge this PR, but if you have anything to clarify, let me know.

rclayton-godaddy:
@seratch If you look at the result you posted, `choices` has no values, but should have the output from the model.  The consequence is that "span end" doesn't receive the output, so implementers of `TracingProvider` have to implement workarounds to associate the outputs to the span.

seratch:
Ah, I see. I got the point now.

rclayton-godaddy:
I would rather have the PR merged and can submit another PR for passing choices back from the streaming provider, so it's available on the span data.

seratch:
Yeah, I agree. We can improve it separately!

PR #674, opened 2025-11-19
fix: #523 transcript removal issue when being interrupted

### LINKED ISSUE #523: Latest output_audio transcripts item in session.history is removed when a user interrupts
### Please read this first

- **Have you read the docs?**  Yes
- **Have you searched for related issues?**  Yes

### Describe the bug

In a realtime twilio session any user speech triggers interrupt, even when the agent isn’t speaking, which forces truncate → retrieve and drops assistant output_audio item transcripts.

### Debug information

- Agents SDK version: v0.1.4
- Runtime environment Node.js v20.10.0

### Repro steps

Log on audio_interrupted and history_updated during a twilio call. Watch audio_interrupted always fire whenever the user speaks, and watch the previously present assistant transcript get dropped due to truncate.

### Expected behavior

The audio_interrupted event should only fire when the assistant is actually being interrupted. Not any time the user speaks.

We should have the option not to truncate even during an interrupt in case we still want to store the transcript for tracing.

Ideally partial transcripts are implemented on interrupt or truncate events.

seratch:
This pull request resolves #523

### DIFF
--- a/.changeset/khaki-bobcats-swim.md
+++ b/.changeset/khaki-bobcats-swim.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: #523 transcript removal issue when being interrupted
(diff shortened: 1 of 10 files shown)


seratch:
will do further WebSocket testing before merging this

seratch:
OK, everything looks great

TrevorJamesH:
Thank you!

PR #712, opened 2025-12-01
fix: strip leading {} from streaming tool call arguments (Bedrock)

anton-tw:
## Summary
- Fixes an issue where some providers (e.g., Bedrock) return an initial empty `{}` followed by actual arguments when streaming tool calls, resulting in malformed `{}{...}` JSON
- Detects and strips the leading `{}` prefix while preserving legitimate empty `{}` arguments

## Test plan
- [x] Added test for stripping `{}` prefix when followed by real arguments
- [x] Added test to verify legitimate empty `{}` arguments are preserved

### DIFF
--- a/.changeset/brave-regions-hug.md
+++ b/.changeset/brave-regions-hug.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: #699 Forward fetch parameter to SSEClientTransport in MCPServerSSE
--- a/.changeset/heavy-balloons-melt.md
+++ b/.changeset/heavy-balloons-melt.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents-core": patch
+---
+
+feat: Add ToolOptions to agents-core package export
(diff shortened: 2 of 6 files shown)


anton-tw:
@seratch We ran into an issue when using Bedrock as the provider - during streaming, tool call arguments would arrive as `{}` in the first chunk, followed by the actual arguments in subsequent chunks. This resulted in malformed JSON like `{}{\"key\":\"value\"}` which breaks parsing.

This PR adds a small fix that detects and strips the leading `{}` when it's followed by real arguments, while still preserving legitimate empty `{}` arguments for tools that don't require parameters.

Would appreciate your thoughts on this approach!

packages/agents-openai/src/openaiChatCompletionsStreaming.ts
```diff
@@ -138,6 +138,9 @@ export async function* convertChatCompletionsStreamToResponses(
   }
 
   for (const function_call of Object.values(state.function_calls)) {
+    if (function_call.arguments.startsWith('{}{')) {
```
seratch, line 144:
```suggestion
    // Some providers, such as Bedrock, may send two items:
    // 1) an empty argument, and 2) the actual argument data.
    // This is a workaround for that specific behavior.
    if (function_call.arguments.startsWith('{}{')) {
```

seratch (approved):

PR #725, opened 2025-12-03
fix: Add Chat Completions choices data in traces when using streaming

jonjomckay:
This PR attempts to fix the issue described at the bottom of https://github.com/openai/openai-agents-js/pull/652 - the `choices` are missing from spans for Chat Completions, for `Generation` spans.

I'm unsure if this is the best way to do this, so I'm opening this to get a discussion going!


packages/agents-openai/src/openaiChatCompletionsModel.ts
```diff
@@ -218,6 +225,55 @@ export class OpenAIChatCompletionsModel implements Model {
           event.type === 'response_done' &&
           response.usage?.total_tokens === 0
         ) {
+          response.choices = event.response.output
+            .flatMap((o): OpenAI.Chat.Completions.ChatCompletion.Choice[] => {
+              if (o.type === 'function_call') {
```
seratch, line 230: The logic here requires updates when the platform adds new types. We'd like to explore more robust way to handle the conversion here.

seratch (changes requested):
Thanks for sending this patch! 

I have a few points on the suggested changes:
- my comment on the conversion logic maintenance cost
- add unit tests covering the scenario

packages/agents-openai/src/openaiChatCompletionsModel.ts
jonjomckay: Yeah, I was worried about that too. Do you have any suggestions? This is the first time I've actually looked into this API!

jonjomckay: @seratch do you have any preferences on how to handle the conversion here?

jonjomckay:
@seratch do you have any suggestions on the above? I don't think this should have been closed yet as there's still unresolved PR comments!

jonjomckay:
@seratch it'd be really nice to get this merged - we're having to maintain a patch set against the library, otherwise we'd be missing critical information in our traces. Do you have any suggestions on your comment in https://github.com/openai/openai-agents-js/pull/725#discussion_r2587016446?

You said you'd like to explore a more robust way to handle the conversion - what were you thinking? I can try and implement it this week.

Issue #733, opened 2025-12-05
'file' message type with data URL should be supported by chat completions


danielmklein:
### Describe the bug

If you run an agent and pass it content that includes an `input_file` where `file` is base64-encoded data, like this...

```typescript
content: [
  {
    type: 'input_file',
    file: "data:application/pdf;base64,JVBER...snip...",
    providerData: {
      filename: "cool-doc.pdf"
    }
  }
],
```

you hit this error: 
https://github.com/openai/openai-agents-js/blob/1300121c5dc8d139b5d7c798a2fc861efa06dd3e/packages/agents-openai/src/openaiChatCompletionsConverter.ts#L99-L103

even though the [docs and API do support passing such an input file](https://platform.openai.com/docs/guides/pdf-files?api-mode=chat). 

It looks like this issue also existed in the python SDK but was fixed: https://github.com/openai/openai-agents-python/issues/893.

### Debug information

- Agents SDK version: (e.g. `v0.3.4)
- Node 22.21.1

### Repro steps

```typescript
import { Agent, run } from '@openai/agents';

const agent = new Agent({
  name: 'Assistant',
  instructions: 'You are a helpful assistant.',
});

const result = await run(
  agent,
  [
    {
      role: 'user',
      content: [
        {
          type: 'input_text',
          text: "what does this file say?",
        }
        {
          type: 'input_file',
          file: "data:application/pdf;base64,JVBER...snip...", // replace with a valid base64-encoded pdf
          providerData: {
            filename: "cool-doc.pdf"
          }
        }
      ],
    }
  ],
);`
```

### Expected behavior

I would expect this to successfully pass the file input data to the API without error.

PR #735, opened 2025-12-05
fix: support input_file for chat completions when possible

danielmklein:
This fixes https://github.com/openai/openai-agents-js/issues/733

Instead of throwing a blanket error in cases when the user passes an `input_file` when working with chat completions, handle cases where the completions API actually does support what was passed.

### DIFF
--- a/.changeset/sharp-towns-cheer.md
+++ b/.changeset/sharp-towns-cheer.md
@@ -1,5 +1,5 @@
 ---
-'@openai/agents-openai': minor
+'@openai/agents-openai': patch
 ---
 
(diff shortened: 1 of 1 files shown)


seratch (approved):
LGTM

.changeset/sharp-towns-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-openai': minor
```
seratch:
```suggestion
'@openai/agents-openai': patch
```

PR #746, opened 2025-12-09
fix: propagate providerData for function_calls in chat completions converter

danielmklein:
We do not currently propagate the provider data on `function_call` items in the chat completions converter -- this can be problematic in cases when a user wants to pass through provider-specific data (especially given that we construct an `assistant` message in one doesn't exist).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,9 +104,9 @@ When opening an issue, include:
 
 1. Fork the repository and create a branch with a descriptive name (e.g., `fix/missing-error`, `feat/new-tool`).
 2. Ensure your branch is up to date with `main`.
-3. Make your changes, add or update tests, and ensure that:
+3. Make your changes, add or update tests, and ensure that the following succeeds:
    ```bash
-   pnpm build && pnpm test && pnpm lint
+   pnpm build && pnpm -r build-check && pnpm test && pnpm lint
    ```
 4. If applicable, generate a changeset (`pnpm changeset`).
 5. Make sure you have [Trufflehog](https://github.com/trufflesecurity/trufflehog) installed to ensure no secrets are accidentally committed.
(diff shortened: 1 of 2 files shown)


danielmklein:
> Thanks for sending this patch. The change may be fine, but can you add some unit tests covering the affected scenarios to prevent future regressions?

@seratch great point -- added!

seratch:
The test code compilation fails. Can you fix it?

```
Error: packages/agents-openai build-check: test/openaiChatCompletionsConverter.test.ts(384,7): error TS2352: Conversion of type '{ type: "function_call_result"; id: string; callId: string; output: string; providerData: { extra: string; }; }' to type '{ type: "function_call_result"; status: "in_progress" | "completed" | "incomplete"; name: string; output: string | { type: "text"; text: string; providerData?: Record<string, any> | undefined; } | { ...; } | { ...; } | ({ ...; } | ... 1 more ... | { ...; })[]; callId: string; providerData?: Record<...> | undefined; ...' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

danielmklein:
> The test code compilation fails. Can you fix it?
> 
> ```
> Error: packages/agents-openai build-check: test/openaiChatCompletionsConverter.test.ts(384,7): error TS2352: Conversion of type '{ type: "function_call_result"; id: string; callId: string; output: string; providerData: { extra: string; }; }' to type '{ type: "function_call_result"; status: "in_progress" | "completed" | "incomplete"; name: string; output: string | { type: "text"; text: string; providerData?: Record<string, any> | undefined; } | { ...; } | { ...; } | ({ ...; } | ... 1 more ... | { ...; })[]; callId: string; providerData?: Record<...> | undefined; ...' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
> ```

@seratch sorry for wasting a bit of your time there! I've fixed that type issue and updated `CONTRIBUTING.md` to add the `build-check` call to the command in the pull request steps as well.

PR #760, opened 2025-12-12
feat: Add responses.compact-wired session feature

seratch:
This pull request adds seamless support of the newly added compaction typed items in agent convo history: https://platform.openai.com/docs/api-reference/responses/compact

## Key changes

The key changes are:
1. Support compaction type items in history
2. Add OpenAIResponsesCompactionSession, which is a decorator of any Session implementation (except OpenAIConversationsSession) and runs the compaction API call when the compaction should be done for the context efficiency

## How it works

When the compaction is done, all the items except user messages in history data will be compacted into a single encrypted content as below:

```json
[
  {
    "id": "msg_0857699edcad0c8600693bbe84218481a0abc41dd37854ea76",
    "type": "message",
    "status": "completed",
    "content": [
      {
        "type": "input_text",
        "text": "Call fetch_image_data with label \"alpha\". Then explain compaction in 1 sentence."
      }
    ],
    "role": "user"
  },
  {
    "id": "msg_0857699edcad0c8600693bbe8f1a9481a083584278da0af74f",
    "type": "message",
    "status": "completed",
    "content": [
      {
        "type": "input_text",
        "text": "Call fetch_image_data with label \"beta\". Then add a fun fact about space in 1 sentence."
      }
    ],
    "role": "user"
  },
  {
    "id": "msg_0857699edcad0c8600693bbe9c271881a0aaa3a993a337f9bf",
    "type": "message",
    "status": "completed",
    "content": [
      {
        "type": "input_text",
        "text": "Call fetch_image_data with label \"gamma\". Then add a fun fact about oceans in 1 sentence."
      }
    ],
    "role": "user"
  },
  {
    "id": "msg_0857699edcad0c8600693bbec5a47c81a092b11723771d5139",
    "type": "message",
    "status": "completed",
    "content": [
      {
        "type": "input_text",
        "text": "Call fetch_image_data with label \"delta\". Then add a fun fact about volcanoes in 1 sentence."
      }
    ],
    "role": "user"
  },
  {
    "id": "msg_0857699edcad0c8600693bbed5b19481a0961473e7127873b9",
    "type": "message",
    "status": "completed",
    "content": [
      {
        "type": "input_text",
        "text": "Call fetch_image_data with label \"epsilon\". Then add a fun fact about deserts in 1 sentence."
      }
    ],
    "role": "user"
  },
  {
    "id": "cmp_0857699edcad0c8601693bbee5e0f081a0a020c6a7f4305161",
    "type": "compaction",
    "encrypted_content": "gAAAAABpO77lnRqRPYqAvgFZT7x9m8uhUfyDC4C4Xb9__..."
  }
]
```

At least, "type": "compaction" must be supported in agent loops.

### DIFF
--- a/packages/agents-core/src/memory/session.ts
+++ b/packages/agents-core/src/memory/session.ts
@@ -51,9 +51,13 @@ export type OpenAIResponsesCompactionArgs = {
   /**
    * The `response.id` from a completed OpenAI Responses API turn, if available.
    *
-   * When undefined, compaction should be skipped.
+   * When omitted, implementations may fall back to a cached value or throw.
    */
-  responseId: string | undefined;
+  responseId?: string | undefined;
+  /**
+   * When true, compaction should run regardless of any internal thresholds or hooks.
+   */
+  force?: boolean;
 };
 
 export interface OpenAIResponsesCompactionAwareSession extends Session {
@@ -66,7 +70,7 @@ export interface OpenAIResponsesCompactionAwareSession extends Session {
    * This hook is best-effort. Implementations should consider handling transient failures and
    * deciding whether to retry or skip compaction for the current turn.
    */
-  runCompaction(args: OpenAIResponsesCompactionArgs): Promise<void> | void;
+  runCompaction(args?: OpenAIResponsesCompactionArgs): Promise<void> | void;
 }
 
 export function isOpenAIResponsesCompactionAwareSession(
--- a/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
+++ b/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
@@ -15,6 +15,23 @@ import {
 const DEFAULT_COMPACTION_THRESHOLD = 10;
 const logger = getLogger('openai-agents:openai:compaction');
 
+export type OpenAIResponsesCompactionDecisionContext = {
+  /**
+   * The `response.id` from a completed OpenAI Responses API turn, if available.
+   */
+  responseId: string | undefined;
+  /**
+   * Items considered compaction candidates (excludes user and compaction items).
+   * The array must not be mutated.
+   */
+  compactionCandidateItems: AgentInputItem[];
+  /**
+   * All stored items retrieved from the underlying session, if available.
+   * The array must not be mutated.
+   */
+  sessionItems: AgentInputItem[];
+};
+
 export type OpenAIResponsesCompactionSessionOptions = {
   /**
    * OpenAI client used to call `responses.compact`.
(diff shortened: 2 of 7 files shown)


rka-oai:
Thank you for doing this! LGTM. LMK if I can help support responses.compact support in the SDK any further. (Let's still wait for someone who's more familiar than me with Agents SDK to stamp, but looks good from a responses POV.

packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
```diff
@@ -0,0 +1,278 @@
...
+  model?: OpenAI.ResponsesModel;
+  /**
+   * Returns the number of items that should contribute to the compaction threshold.
+   *
+   * This function is used to decide when to call `responses.compact`, and it is also used to keep
+   * an incremental count as new items are appended to the underlying session.
+   *
+   * Defaults to counting every stored item except `compaction` items and user messages.
+   */
+  countCompactionItems?: (items: AgentInputItem[]) => number;
```
dkundel-openai: I think the most common way people would want to compact is based on the token size of messages not necessarily the amount of messages. That is how I think Codex does it for example. Right now my understanding is the way that you can do that is by changing the `countCompactionItems` to count tokens instead and then set the `compactionThreshold` as a token number. Right?

Is there a way we can make that more seamless?

seratch: @dkundel-openai This is a good point. 

A challenge is that our APIs do not provide a simple way to get the accumulated token consumption for an agent turn. Therefore, I believe compactionThreshold + countCompactionItems should be one reasonable way to decide to run compaction. 
I still think we can use this simple method as the default built-in one, but I agree we should redesign this  hook to be a more generic one like `shouldCompact(items): boolean`. With that, developers can do anything in the method. What do yo think?

packages/agents-core/src/memory/session.ts
```diff
@@ -43,3 +43,38 @@ export interface Session {
...
+export type OpenAIResponsesCompactionArgs = {
+  /**
+   * The `response.id` from a completed OpenAI Responses API turn, if available.
+   *
+   * When undefined, compaction should be skipped.
+   */
+  responseId: string | undefined;
+};
+
+export interface OpenAIResponsesCompactionAwareSession extends Session {
```
dkundel-openai, line 63: If this type of Session extends the regular session can't we just call compaction inside the `addItem` part of Session rather than introducing a new subtype? My understanding was that the point of Session was that it would decide what to store and how to store it.

packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
seratch: Just to clarify, I am fine with any names for the hook apart from "shouldCompact"

packages/agents-core/src/memory/session.ts
seratch: When you use `responses.compact` API with an external session store, your code has to clear all items (at least all non-user-message items) associated with the session ID first, then re-insert everything (N user messages + 1 compaction item). Because of that, it doesn’t fit well with `addItems` / `getItems` method customization.

That said, it's still feasible to do above in `addItems` method. Another benefit of the current design is that you can use this new subclass as a [decorator-design-pattern](https://en.wikipedia.org/wiki/Decorator_pattern) style wrapper around your existing session store. This enables developers to reuse this single logic without having similar logic within their code.

packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
seratch: @dkundel-openai I've revisited the design by https://github.com/openai/openai-agents-js/pull/760/changes/014b9263e242f4628455b12930d4a597238aedd7:
- removed compactionThreshold, countCompactionItems
- instead, added general callback named shouldTriggerCompaction
- added compactionCandidateItems, sessionItems, and responseId for custom shouldTriggerCompaction logic

packages/agents-core/src/memory/session.ts
seratch: I didn't change this for the above reason.

Issue #663, opened 2025-11-17
ComputerTool improvements


markb-trustifi:
### Please read this first

- **Have you read the docs?** YES
- **Have you searched for related issues?** YES

### Describe the feature

Please add the following functionality to the `computerTool`:
- Computer Tool misses `needsApproval`, so it can be approved/rejected only on the Agent level:
```
const computerAgent = new Agent({tools: [computerTool({ computer })]});
const computerTool = computerAgent.asTool({
   needsApproval: async (runContext, args) => {...}
});
```
See in `executeComputerActions`, it misses code something like:
```
const approvalItem = new RunToolApprovalItem(toolCall, agent, computerTool.name);
const requiresApproval = await computerTool.needsApproval(runContext, toolCall.action, toolCall.callId);
```
- Computer Tool doesn't pass runContext to the ComputerBase class.
The `_runComputerActionAndScreenshot` should pass it to every computer action, something like:
`computer.click(action.x, action.y, action.button, runContext)`
- The ComputerBase class instance is actually a statefull and doesn't differ between `Runner.run` executions. It can also lead to the security risks because different Agent loops can be run by different users. To reinitialize it before every runner loop, ComputerBase could have `initRun` method that will be called on "agent_start" event.

seratch:
Thanks for sharing these feedback! Yes, I am aware of these limitations. We'll consider enhancing the tool in future releases.

markb-trustifi:
Regarding point 2, it will be great if you could change the line:
https://github.com/openai/openai-agents-js/blob/5d2fc833b7d20322a2a5542b689a9b47f6a7b1f4/packages/agents-core/src/runImplementation.ts#L1721
with this code:
```js
if (computer.invoke) {
   output = await computer.invoke(runContext, toolCall);
}
else {
   output = await _runComputerActionAndScreenshot(computer, toolCall);
}
```

PR #761, opened 2025-12-12
Feat: Computer tool - add invoke method with RunContext

markb-trustifi:
This PR resolves issue #663.
Right now Computer tool should implement each UI action as a separate method. These methods also don't have RunContext parameter opposite to regular tools. It makes impossible passing user context when executing UI actions.

### The fix adds:
One common method for all UI actions with `RunContext` (contains user context) and `ComputerUseCallItem` (contains UI actions). 
The fix keeps backward compatibility with the regular methods definition.

### Invoke method implementation example:
```ts
const { computerTool, Computer, asInvokeComputer, isInvokeComputer  } = require("@openai/agents");
function makeComputer(): Computer {
	return asInvokeComputer({
		environment: 'browser',
		dimensions: [1, 1],
		invoke: async (runContext, toolCall) => {
			const action = toolCall.action;
			const ws = runContext.context.ws;
			const {promise, resolve, reject} = Promise.withResolvers();

			ws.once("message", (raw) => {
				let msg;
				try {
					msg = JSON.parse(raw);
				} catch (e) {
					return reject(new Error("Invalid message."));
				}
				if (!msg.screenshot) {
					return reject(new Error("Screenshot not found."));
				}
				resolve(msg.screenshot);
			});

			ws.send(JSON.stringify(action));

			return promise;
		}
	});
}
const comp = makeComputer();
const tool = computerTool({ computer: comp });
```

### DIFF
--- a/packages/agents-core/src/index.ts
+++ b/packages/agents-core/src/index.ts
@@ -13,7 +13,7 @@ export {
   ToolUseBehavior,
   ToolUseBehaviorFlags,
 } from './agent';
-export { Computer } from './computer';
+export { Computer, asInvokeComputer, isInvokeComputer } from './computer';
 export { ShellAction, ShellResult, ShellOutputResult, Shell } from './shell';
 export { ApplyPatchOperation, ApplyPatchResult, Editor } from './editor';
 export {


seratch:
Thanks for sharing this implementation idea. However, I don't think adding new "invoke" method is the only and best way to achieve the goal to separate different computer runs. We prefer a simpler fix like detecting simultaneous executions using the same computer instance or anything else, which does not require existing computer implementations to change their internals.

markb-trustifi:
> Thanks for sharing this implementation idea. However, I don't think adding new "invoke" method is the only and best way to achieve the goal to separate different computer runs. We prefer a simpler fix like detecting simultaneous executions using the same computer instance or anything else, which does not require existing computer implementations to change their internals.

The main goal was to add user context to agent method executions, similar to how regular tools work.
I added a new invoke method in order to avoid changing the existing method signatures.
If you can think of another way to add RunContext, I’d be happy to hear about it.
Current CUA implementation doesn't differ between multiple users or multiple screen sizes and requires creating new Agent for every run loop.

seratch:
Here is my approach for resolving this issue: https://github.com/openai/openai-agents-js/pull/771 If you have any comments, please feel free to share them!

seratch:
Thanks again for sharing this idea. Closing this PR in favor of https://github.com/openai/openai-agents-js/pull/771

PR #771, opened 2025-12-16
fix: Enable creating/disposing Computer per agent run ref: #663

seratch:
This pull request resolves the second point mentioned at #663; 

Currently the computer tool accepts only a static Computer instance and it could be shared among runs. If you want to separate computer executions for multiple requests, you have to create agent and its required tools for each request. 

This pull request enables developers to easily create/dispose a Computer per agent run.

The existing approach (still supported):
```typescript
  // If your app never runs multiple computer using agents at the same time,
  // you can create a singleton computer and use it in all your agents.
  const computer = await new LocalPlaywrightComputer().init();
  try {
    const agent = new Agent({
      name: 'Browser user',
      model: 'computer-use-preview',
      instructions: 'You are a helpful agent.',
      tools: [computerTool({ computer })],
    });
    const result = await run(agent, "What's the weather in Tokyo?");
    console.log(`\nFinal response:\n${result.finalOutput}`);
  } finally {
    await computer.dispose();
  }
```

New approach this pull request adds:
```typescript
  // If your app runs multiple computer using agents at the same time,
  // you can create a computer per request.
  const agent = new Agent({
    name: 'Browser user',
    model: 'computer-use-preview',
    instructions: 'You are a helpful agent.',
    tools: [
      computerTool({
        // initialize a new computer for each run and dispose it after the run is complete
        computer: {
          create: async ({ runContext }) => {
            console.log('Initializing computer for run context:', runContext);
            return await new LocalPlaywrightComputer().init();
          },
          dispose: async ({ runContext, computer }) => {
            console.log('Disposing of computer for run context:', runContext);
            await computer.dispose();
          },
        },
      }),
    ],
  });
  const result = await run(agent, "What's the weather in Tokyo?");
  console.log(`\nFinal response:\n${result.finalOutput}`);
```

If you don't need to dispose a Computer instance, you can pass only the create function:
```typescript
  const agent = new Agent({
    name: 'Browser user',
    model: 'computer-use-preview',
    instructions: 'You are a helpful agent.',
    tools: [
      computerTool({
        // initialize a new computer for each run
        computer: async ({ runContext }) => {
          console.log('Initializing computer for run context:', runContext);
          return await new LocalPlaywrightComputer().init();
        },
      }),
    ],
  });
```

### DIFF
--- a/packages/agents-core/test/runImplementation.test.ts
+++ b/packages/agents-core/test/runImplementation.test.ts
@@ -1913,9 +1913,9 @@ describe('executeComputerActions', () => {
     expect(result.every((r) => r instanceof ToolCallOutputItem)).toBe(true);
   });
 
-  it('initializes computer with run context using createComputer', async () => {
-    const createComputer = vi.fn(async () => makeComputer());
-    const tool = computerTool({ computer: createComputer });
+  it('initializes computer with run context using a computer initializer', async () => {
+    const initializer = vi.fn(async () => makeComputer());
+    const tool = computerTool({ computer: initializer });
     const toolCall: protocol.ComputerUseCallItem = {
       id: 'id1',
       type: 'computer_call',
@@ -1947,7 +1947,7 @@ describe('executeComputerActions', () => {
       ctxB,
     );
 
-    expect(createComputer).toHaveBeenCalledTimes(2);
+    expect(initializer).toHaveBeenCalledTimes(2);
     const compA = await resolveComputer({ tool, runContext: ctxA });
     const compB = await resolveComputer({ tool, runContext: ctxB });
     expect(compA).not.toBe(compB);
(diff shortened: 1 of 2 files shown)


packages/agents-core/src/tool.ts
```diff
@@ -131,13 +131,90 @@ export type FunctionTool<
...
+};
+
+/**
+ * A function that initializes a computer for the current run.
+ */
+type BivariantComputerCreate<
+  Context = UnknownContext,
+  TComputer extends Computer = Computer,
+> = {
+  bivarianceHack: (
```
dkundel-openai, line 151: Minor thing but do we need this to be called bivarianceHack? Not sure it's very explainatory

seratch: Yeah, it could be confusing, so I've added code comment clarifying what it is: https://github.com/openai/openai-agents-js/pull/771/changes/bdb6253d66dcf46f80e53da242b36789f7d00d26

The naming seems to be a common pattern in TS community when using [the function-parameter-bivariance technique](https://www.typescriptlang.org/docs/handbook/type-compatibility.html#function-parameter-bivariance). Thus, going with the name should be fine.

PR #847, opened 2026-01-12
feat: #645  support function call history issue

### LINKED ISSUE #645: Add support for adding function call items to conversation history via `updateHistory`
### Please read this first

- **Have you read the docs?** Yes - [Agents SDK docs](https://openai.github.io/openai-agents-js/)
- **Have you searched for related issues?** Yes

### Describe the feature

**Feature:** Add support for manually creating and adding function call items to conversation history via `updateHistory`.

**Current Limitation:**

Currently, when using `updateHistory()` to modify the Realtime API conversation history, function call items are ignored with a warning (see [code here](https://github.com/openai/openai-agents-js/blob/main/packages/agents-realtime/src/openaiRealtimeBase.ts#L886)):

```typescript
} else if (addition.type === 'function_call') {
  logger.warn(
    'Function calls cannot be manually added or updated at the moment. Ignoring.',
  );
}
```

This means that while `history_updated` events include tool calls in the returned history array (`RealtimeItem[]` which includes `RealtimeToolCallItem`), there's no way to manually add or restore tool calls when managing conversation history.

Skip06:
This PR adds explicit support for the function_call type within the resetHistory loop. It translates the SDK's local history item into the required conversation.item.create event format for the OpenAI Realtime API.

Key Changes:

Logic: Added an else if block in packages/agents-realtime/src/openaiRealtimeBase.ts to identify and process function_call items.

Mapping: Correctly maps itemId to id and ensures the call_id, name, and arguments are preserved in the payload sent to the server.

Verification: Added a unit test in packages/agents-realtime/test/openaiRealtimeBase.test.ts to confirm that the correct event is emitted when a tool call is present in the history

Resolves #645


Skip06:
@seratch I tried to work the issue #645 . 
I added an else if block in openaiRealtimeBase.ts to identify and process "function_call" and then I  added a unit test in openaiRealtimeBase.test.ts

tsconfig.json
```diff
@@ -45,7 +45,7 @@
     // "resolvePackageJsonExports": true,                /* Use the package.json 'exports' field when resolving package imports. */
     // "resolvePackageJsonImports": true,                /* Use the package.json 'imports' field when resolving imports. */
     // "customConditions": [],                           /* Conditions to set in addition to the resolver-specific defaults when resolving imports. */
-    "noUncheckedSideEffectImports": true,             /* Check side effect imports. */
+    // "noUncheckedSideEffectImports": true,             /* Check side effect imports. */
```
seratch, line 48: I don't think this change is necessary

PR #973, opened 2026-02-14
fix(agents-core): drop trailing reasoning items without output

pushkarkumarvats:
Trailing **reasoning** items without a following assistant output were being
persisted during streaming, which violates the API invariant and can cause
400 errors on subsequent turns.

This change trims orphaned trailing reasoning items during output extraction
and adds tests covering multiple edge cases.

### DIFF
--- a/packages/agents-core/test/runner/items.extractOutput.test.ts
+++ b/packages/agents-core/test/runner/items.extractOutput.test.ts
@@ -23,7 +23,9 @@ describe('extractOutputItemsFromRunItems', () => {
       ),
       new RunMessageOutputItem(
         {
-          type: 'assistant_message',
+          type: 'message',
+          role: 'assistant',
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the PR and for digging into this issue. After deeper validation, we're going to close this PR without merging.

This change mitigates a symptom by trimming trailing reasoning items, but it does not address the root cause of the 400 errors. The failure is specifically tied to reasoning items with IDs being sent without their required following item. A blanket trim can avoid some failures, but it can also drop valid reasoning context.

We need to run thorough testing first, but for most cases, removing id property in reasoning items could be a more robust workaround. That said, it is still a workaround, so we won't change the default behavior to automatically strip all the IDs before sending to Responses API. We may add an option to turn this behavior on in future releases rather than simply removing the last reasoning item.

PR #984, opened 2026-02-18
fix: #479 avoid Bun browser startup crash when shim modules are temporarily unresolved

### LINKED ISSUE #479: Bun bundled application fails to run in browser (reading `loadEnv`)
### Describe the bug

When using bun to bundle a browser-client app, if the `@openai/realtime-agents` package is imported it will simply break at runtime.

```
frontend TypeError: Cannot read properties of null (reading 'loadEnv')
      at loadEnv3 (/Users/chris/Projects/bun-openai-bug/node_modules/@openai/agents-core/dist/config.mjs:8:12)
      at isEnabled (/Users/chris/Projects/bun-openai-bug/node_modules/@openai/agents-core/dist/config.mjs:17:17)
      at get dontLogModelData (/Users/chris/Projects/bun-openai-bug/node_modules/@openai/agents-core/dist/config.mjs:41:16)
      at /Users/chris/Projects/bun-openai-bug/node_modules/@openai/agents-core/dist/logger.mjs:7:34
    from browser tab http://localhost:3000/
```

### Debug information

- Agents SDK version: `"@openai/agents-realtime": "^0.1.3"`
- Runtime environment: `Bun v1.2.22` (Chrome browser)

### Repro steps

Repro: https://github.com/chrisui/bun-openai-bug

```
git clone git@github.com:chrisui/bun-openai-bug.git
bun install
bun index.html
```

goto localhost:3000 and you'll see the error

### Expected behavior

No runtime error.

Compare to it working with `bunx vite dev index.html`

seratch:
This pull request fixes #479 a Bun browser startup crash where `@openai/agents-core` could throw `TypeError: Cannot read properties of null (reading 'loadEnv')` during module initialization. The fix makes `config.loadEnv` and browser-environment checks resilient when shim exports are temporarily unavailable, and updates `logger` so logging flags are evaluated lazily instead of at module load time. A regression test was added to ensure logger initialization does not eagerly evaluate logging flags. This change is a backwards-compatible runtime hardening update for `@openai/agents-core`.

### DIFF
--- a/.changeset/tiny-gorillas-wave.md
+++ b/.changeset/tiny-gorillas-wave.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: #479 avoid Bun browser startup crash when shim modules are temporarily unresolved
--- a/packages/agents-core/src/config.ts
+++ b/packages/agents-core/src/config.ts
@@ -1,9 +1,25 @@
 // Use function instead of exporting the value to prevent
 // circular dependency resolution issues caused by other exports in '@openai/agents-core/_shims'
-import {
-  loadEnv as _loadEnv,
-  isBrowserEnvironment,
-} from '@openai/agents-core/_shims';
+import * as _shims from '@openai/agents-core/_shims';
+
+function fallbackIsBrowserEnvironment(): boolean {
+  return (
+    typeof window !== 'undefined' &&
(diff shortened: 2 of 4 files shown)


seratch:
Confirmed this works with the repro steps in https://github.com/openai/openai-agents-js/issues/479

PR #993, opened 2026-02-19
fix(agents-core): rehydrate RunState interruption items on resume

seratch:
This pull request fixes a runtime bug in resumed runs where `RunState.fromString()` left `next_step_interruption` items as plain JSON objects instead of `RunToolApprovalItem` instances. The bug caused post-resume behavior mismatches, including missing class getters (for example `.name`) and inconsistent interruption handling expectations.

The change explicitly types `getInterruptions()` as `RunToolApprovalItem[]` in `packages/agents-core/src/runState.ts`, and rehydrates interruption entries during deserialization so resumed state matches in-memory state. It supports both the current serialized item format and legacy raw interruption shapes.

The update also expands unit coverage in `packages/agents-core/test/runState.test.ts` for:
- function-tool interruption rehydration,
- hosted-tool interruption rehydration plus approval flow,
- legacy interruption payload compatibility.

### DIFF
--- a/.changeset/four-olives-grin.md
+++ b/.changeset/four-olives-grin.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: rehydrate RunState interruptions and type getInterruptions
--- a/packages/agents-core/src/runState.ts
+++ b/packages/agents-core/src/runState.ts
@@ -557,11 +557,14 @@ export class RunState<TContext, TAgent extends Agent<any, any>> {
   /**
    * Returns all interruptions if the current step is an interruption otherwise returns an empty array.
    */
-  getInterruptions() {
+  getInterruptions(): RunToolApprovalItem[] {
     if (this._currentStep?.type !== 'next_step_interruption') {
       return [];
     }
-    return this._currentStep.data.interruptions;
+    const interruptions = this._currentStep.data.interruptions;
+    return Array.isArray(interruptions)
+      ? (interruptions as RunToolApprovalItem[])
+      : [];
   }
 
   private getPendingAgentToolRunKey(toolName: string, callId: string): string {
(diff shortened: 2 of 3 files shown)


dhruvbird:
@seratch Thanks for putting up the fix! 

I haven't yet tested this change, but wanted to check if this works for interruptions that happen on a tool within an agent as tool. For example, 

main-agent > agent-as-tool > tool-with-interruption. 

The reason I ask is because I looked at `buildAgentMap` in [runState.ts](https://github.com/openai/openai-agents-js/blob/main/packages/agents-core/src/runState.ts#L976-L1003) and it seems like the `agentMap` is only for handoff agents and not agents used as tools. 

Specifically in the case above, does `RunToolApprovalItem` constructor need to get instance of main-agent or agent-as-tool when handling interruption for tool-with-interruption, which is a tool on agent-as-tool? 

@seratch to clarify, previously the agent name was: 

```
"agent": {"name": "agent-as-tool" }
```

and now it's: 

```
"agent": {"name": "main-agent" }
```

which means that the agent instance that is being picked up from agentMap is that of the main agent.

seratch:
@dhruvbird Thanks for calling this out. You were right. We'll resolve this issue shortly.

dhruvbird:
> @dhruvbird Thanks for calling this out. You were right. We'll resolve this issue shortly.

@seratch 

One of my team mates tested your change and confirmed that it does in fact work for interrupts on tools that are on an agent as tool. However, the agent name is wrong. Surprisingly it does work, so practially there is no fix needed, but it still begs the question of why it works. 

cc @suwenzhu

seratch:
The reason the HITL logic still works despite this bug is that the approval flow and the subsequent tool execution rely only on the tool name and call ID. As long as the tool call resolves to the correct agent to run, the incorrect agent reference in the interruption object does not affect execution. That said, I’m working on a fix, and the incorrect agent reference will be resolved soon.

PR #996, opened 2026-02-20
fix(agents-core): preserve nested agent-tool approval attribution on restore

seratch:
This pull request fixes a nested interruption resume bug where a tool approval raised inside an agent-as-tool flow could be restored as if it belonged to the outer agent after RunState serialization/deserialization. This is just a data reference bug in interruption data object, so the HITL logic works correctly regardless of it.

The change updates the agents-core runtime to keep an internal reference from agent-tools back to their source agent and extends run-state agent map reconstruction to traverse agent-tools in addition to handoffs. This preserves the correct RunToolApprovalItem.agent for flows like main-agent > agent-as-tool > tool-with-interruption, and adds a regression assertion to lock that behavior in after RunState.fromString(...).


seratch:
@codex review again

Issue #261, opened 2025-07-28
ModelBehaviorError: Invalid output type when using @openrouter/ai-sdk-provider with outputType


rajnandan1:
### Describe the bug

ModelBehaviorError: Invalid output type when using openrouter with aisdk and outputType

```bash
ModelBehaviorError: Invalid output type
    at executeToolsAndSideEffects (/Users/rajnandan1/Code/node_modules/@openai/agents-core/src/runImplementation.ts:614:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async <anonymous> (/Users/rajnandan1/Code/node_modules/@openai/agents-core/src/run.ts:408:32)
    at async <anonymous> (/Users/rajnandan1/Code/node_modules/@openai/agents-core/src/tracing/context.ts:60:20)
    at async <anonymous> (/Users/rajnandan1/Code/aisdk.ts:29:17) {
  state: undefined
}
```

### Debug information

- Agents SDK version: (e.g. `v0.0.13`)
- Runtime environment (e.g. `Node.js v24.1.0`)
- "zod": "^3.25.67",
- "@openrouter/ai-sdk-provider": "^0.7.3",
- "@ai-sdk/openai": "^1.3.23",
- "@openai/agents-extensions": "^0.0.13",

### Repro steps

```js
import { Agent } from '@openai/agents';
import { aisdk } from '@openai/agents-extensions';
import { createOpenRouter } from '@openrouter/ai-sdk-provider';
import { Runner } from '@openai/agents';
import { z } from 'zod';
import dotenv from 'dotenv';
dotenv.config();
const StartAgentResponseSchema2 = z.object({
	content: z.string().describe('Markdown formatted content ')
});

const modelName = 'anthropic/claude-sonnet-4';
const openrouter = createOpenRouter({
	apiKey: process.env.OPEN_ROUTER_KEY || ''
});
const model = aisdk(openrouter(modelName));

const runner = new Runner({
	model: model,
	tracingDisabled: true
});
const agent = new Agent({
	name: 'Assistant',
	instructions: 'You are a helpful assistant',
	outputType: StartAgentResponseSchema2
});

try {
	const result = await runner.run(agent, 'Write a haiku about  programming.');
	console.log(result.finalOutput);
} catch (e) {
	console.log(e);
}
```

### Expected behavior

```json
{
  'content': 'Function calls itself  \nEndless echoes in the code  \nI see me again'
}
```

[...]

PR #1034, opened 2026-03-01
fix: #261 robustly parse JSON output when using openrouter/aisdk

sliortega295-ops:
This PR fixes a bug where third-party models (via OpenRouter or ai-sdk) return JSON responses wrapped in markdown code blocks (```json ... ```), which causes `ModelBehaviorError: Invalid output type` during parsing in `turnResolution.ts`.

I've added logic to defensively clean up the response and extract the JSON payload, preventing validation errors for otherwise valid outputs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Run build
         run: pnpm build
       - name: Install, build, and upload your site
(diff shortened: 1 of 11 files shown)


seratch:
@sliortega295-ops Thanks for sending this patch! I think https://github.com/openai/openai-agents-js/pull/1036 should be more flexible and robust customization option for ai-sdk users. Thus, let me close this one in favor of #1036

PR #1057, opened 2026-03-05
feat: add message parameter to reject() on RunState

KostaGPT:
## Summary

This PR adds an optional `message` parameter to `RunState.reject()`, giving callers a simpler, per-call way to control the rejection text the agent sees. When provided, `message` takes precedence over the `toolErrorFormatter` callback and the SDK default string (`"Tool execution was not approved."`).
## Motivation

A common UX pattern is "ignore and redirect": the user sees a pending tool call (e.g., "Send email with subject X?") and instead of approving or declining, sends new instructions ("Change the subject to Y and resend"). The agent needs to see both the rejection *and* the user's message to act correctly.

Previously, the decision to reject happened at the call site (e.g., in response to a "Type yes to approve, no to reject" prompt), but the rejection *message* had to be constructed elsewhere — in a global `toolErrorFormatter` callback. This meant reading the user's input from run context as a side-channel. The decision and its explanation were split across two different places in the code.

## What changed

**New API on `RunState.reject()`:**

```ts
// message is used as the rejection text sent to the model.
// If not provided, toolErrorFormatter (if configured) or the SDK default is used.
state.reject(interruption, { message: 'Not allowed in this environment.' });
```

**Precedence:** `message` > `toolErrorFormatter` > SDK default (`"Tool execution was not approved."`)

At the `reject()` call site, the caller already has access to everything the formatter receives as arguments — the interruption (which carries `toolName` and `callId`), the agent reference, and the run context.

## Test plan

- Added unit tests for `reject()` with message, serialization round-trip preservation, and per-callId message isolation
- Added test verifying `message` takes precedence over `toolErrorFormatter`
- All existing formatter tests preserved and passing
- All 1354 tests pass, full build succeeds

### DIFF
--- a/packages/agents-core/src/runContext.ts
+++ b/packages/agents-core/src/runContext.ts
@@ -126,6 +126,12 @@ export class RunContext<TContext = UnknownContext> {
     return this.#approvals.get(toolName)?.messages?.[callId];
   }
 
+  #getCallId(approvalItem: RunToolApprovalItem): string {
+    return 'callId' in approvalItem.rawItem
+      ? approvalItem.rawItem.callId // function tools
+      : approvalItem.rawItem.id!; // hosted tools
+  }
+
   /**
    * Check if a tool call has been approved.
    *
@@ -200,12 +206,7 @@ export class RunContext<TContext = UnknownContext> {
       rejected: [],
     };
     if (Array.isArray(approvalEntry.approved)) {
-      // function tool has call_id, hosted tool call has id
-      const callId =
-        'callId' in approvalItem.rawItem
-          ? approvalItem.rawItem.callId // function tools
-          : approvalItem.rawItem.id!; // hosted tools
-      approvalEntry.approved.push(callId);
+      approvalEntry.approved.push(this.#getCallId(approvalItem));
     }
     this.#approvals.set(toolName, approvalEntry);
   }
@@ -225,14 +226,12 @@ export class RunContext<TContext = UnknownContext> {
     const toolName =
       approvalItem.toolName ?? (approvalItem.rawItem as any).name;
     if (alwaysReject) {
-      const callId =
-        'callId' in approvalItem.rawItem
-          ? approvalItem.rawItem.callId
-          : approvalItem.rawItem.id!;
       this.#approvals.set(toolName, {
         approved: false,
         rejected: true,
-        ...(message ? { messages: { [callId]: message } } : {}),
+        ...(message
+          ? { messages: { [this.#getCallId(approvalItem)]: message } }
+          : {}),
       });
       return;
     }
@@ -243,11 +242,7 @@ export class RunContext<TContext = UnknownContext> {
     };
 
     if (Array.isArray(approvalEntry.rejected)) {
-      // function tool has call_id, hosted tool call has id
-      const callId =
-        'callId' in approvalItem.rawItem
-          ? approvalItem.rawItem.callId // function tools
-          : approvalItem.rawItem.id!; // hosted tools
+      const callId = this.#getCallId(approvalItem);
       approvalEntry.rejected.push(callId);
       if (message) {
         approvalEntry.messages = approvalEntry.messages ?? {};


seratch:
@KostaGPT Thanks for sharing this. I like the option addition, but we're unable to accept ToolErrorFormatter deletion as it is a breaking change to other users.

KostaGPT:
Thanks for the quick feedback! Wasn't sure about that part — went with the bolder option to see if that was acceptable. I'll keep `toolErrorFormatter` in place and scope this PR to just adding the `message` parameter on `reject()`.

A couple of questions to make sure I get the coexistence right:

1. **Precedence**: When both `message` and `toolErrorFormatter` are provided, should `message` take priority (since it's more specific, being per-call vs global)?
2. **Deprecation**: Would you like `toolErrorFormatter` marked as `@deprecated` pointing users toward `message`, or should it remain as a fully supported alternative?

seratch:
>Precedence: When both message and toolErrorFormatter are provided, should message take priority (since it's more specific, being per-call vs global)?

Yeah, this should make sense. toolErrorFormatter can act as the universal default one.

>Deprecation: Would you like toolErrorFormatter marked as @deprecated pointing users toward message, or should it remain as a fully supported alternative?

We don't want to mark toolErrorFormatter as deprecated this time. For other users, the simple customization it provides could be still useful.

KostaGPT:
Got it, thanks.

>For other users, the simple customization it provides could be still useful.

FWIW, for me as a user, being able to supply the rejection message right at the .reject() call seems to be the simplest and most direct way to do this, even if all I want to do is always specify the same custom constant string. I found the ergonomics of `toolErrorFormatter` a bit awkward.

examples/tools/computer-use-hitl.ts
```diff
@@ -1,29 +1,9 @@
 import { chromium, Browser, Page } from 'playwright';
 import { createInterface } from 'node:readline/promises';
```
KostaGPT, line 2: @seratch I've kept the new way in the example, let me know if you want me to roll this back too.

seratch: This is fine! It's simpler and I agree new users would like to learn this first.

.changeset/reject-message-parameter.md
```diff
@@ -0,0 +1,9 @@
+---
+'@openai/agents-core': minor
+'@openai/agents-realtime': minor
```
seratch: When there is no breaking change, this project uses patch bump

```suggestion
'@openai/agents-core': patch
'@openai/agents-realtime': patch
```

seratch (changes requested):

examples/tools/computer-use-hitl.ts
KostaGPT: 🤘

seratch:
Here are a few findings from my local codex reviews:

- `alwaysReject` currently drops the new `message` option. In packages/agents-core/src/runContext.ts#L218, the `alwaysReject` branch returns before storing any custom rejection message, so `state.reject(item, { alwaysReject: true, message: "..." })` still falls back to `toolErrorFormatter` or the default SDK text. Suggested solution: either persist a tool-level sticky rejection message and make lookup fall back to it for future calls, or explicitly disallow/document the `alwaysReject + message` combination.

- The serialized `RunState` schema changed without a version bump. packages/agents-core/src/runState.ts#L39 still says the schema is `1.6`, but packages/agents-core/src/runState.ts#L314 adds `context.approvals[*].messages`. That means version-skew scenarios can silently lose the new data. Suggested solution: bump the schema version to `1.7`, update the version history/support list, and add a compatibility test that proves old payloads still load while new payloads preserve rejection messages.

- Empty-string messages are ignored. In packages/agents-core/src/runContext.ts#L247, the code uses `if (message)`, so `message: ""` is treated as “not provided.” Suggested solution: decide the intended contract and encode it explicitly: use `message !== undefined` if empty strings are valid, or reject/document empty strings as unsupported.

seratch:
Also, I already appreciate sharing the idea! If you would like me to take over this PR and complete it, I will work on it later on.

KostaGPT:
> Here are a few findings from my local codex reviews:

Thanks, just pushed a change to cover the `alwaysReject` case.

What would you like to do on the schema version and empty string scenario?

KostaGPT:
> Also, I already appreciate sharing the idea! If you would like me to take over this PR and complete it, I will work on it later on.

Sure, that sounds great, thank you! I'll be looking forward to the release :)

KostaGPT:
I couldn't help but do a small additional refactor in f18afd3, feel free to revert.

I leave the rest to you, and thanks again!

seratch:
Closing this in favor of https://github.com/openai/openai-agents-js/pull/1058

KostaGPT:
> Closing this in favor of #1058

Thanks for the quick turnaround!

PR #1071, opened 2026-03-07
fix: #1070 preserve MCP image mimeType in tool outputs

### LINKED ISSUE #1070: normalizeStructuredToolOutput ignores MCP mimeType when building data URLs for image content
### Description

When an MCP server returns image content following the MCP protocol spec (`{type: 'image', data: '<base64>', mimeType: 'image/jpeg'}`), the SDK fails to construct a valid data URL. The raw base64 string ends up in the `image_url` field of the OpenAI Responses API request, which rejects it with:

```
400 Invalid 'input[N].output[0].image_url'. Expected a valid URL, but got a value with an invalid format.
```

### Root Cause

In `packages/agents-core/src/runner/toolExecution.ts`, `normalizeStructuredToolOutput` reads `value.mediaType` to derive the MIME type for constructing the data URL:

```typescript
const fallbackImageMediaType = isNonEmptyString(value.mediaType)
    ? value.mediaType
    : undefined;
```

However, the **MCP protocol spec** uses `mimeType` (not `mediaType`) for image content. Since `value.mediaType` is always `undefined` for standard MCP responses, `fallbackImageMediaType` is never set, and the fallback path produces a raw base64 string instead of a proper `data:image/jpeg;base64,...` URL:

```typescript
imageString = fallbackImageMediaType
    ? toInlineImageString(value.data, fallbackImageMediaType)  // ← never reached
    : value.data;  // ← raw base64, not a valid URL
```

This raw base64 propagates through `convertStructuredToolOutputToInputItem` → `convertStructuredOutputToRequestItem`, where it gets placed into `image_url`, which the OpenAI Responses API rejects.

> **Note:** adding `mediaType` on the MCP server side is not a workaround because `@modelcontextprotocol/sdk` uses a Zod schema that strips unknown keys on the client side, so the field never reaches this code.

### Proposed Fix

```typescript
// packages/agents-core/src/runner/toolExecution.ts

const fallbackImageMediaType = isNonEmptyString(value.mediaType)
    ? value.mediaType
    : isNonEmptyString(value.mimeType)   // ← also check MCP's field name
        ? value.mimeType
        : undefined;
```

### Reproduction

1. Set up an MCP server that returns `ImageContent` (e.g. `{type: 'image', data: '<base64>', mimeType: 'image/jpeg'}`) from a tool call.
2. Use `MCPServerStreamableHttp` with an `Agent` and call `run()`.
3. Observe the `400 Invalid '...image_url'` error from the OpenAI Responses API.

### Version

`@openai/agents-core` 0.5.2

seratch:
This pull request resolves #1070.

This pull request fixes MCP image outputs that include `mimeType` but not `mediaType`, which previously caused the runtime to forward raw base64 into Responses `image_url` fields and trigger a 400 validation error. It updates both the core tool-output normalization path and the OpenAI Responses adapter so MCP-style image payloads are converted into proper `data:<mime>;base64,...` URLs whether the image data is returned at the top level or inside nested image objects.

### DIFF
--- a/packages/agents-openai/src/openaiResponsesModel.ts
+++ b/packages/agents-openai/src/openaiResponsesModel.ts
@@ -586,7 +586,8 @@ function convertLegacyToolOutputContent(
       structured.image = output.image;
     } else if (isRecord(output.image)) {
       const imageObj = output.image as Record<string, any>;
-      const inlineMediaType = getImageInlineMediaType(imageObj);
+      const inlineMediaType =
+        getImageInlineMediaType(imageObj) ?? topLevelInlineMediaType;
       if (typeof imageObj.url === 'string' && imageObj.url.length > 0) {
         structured.image = imageObj.url;
       } else if (
--- a/packages/agents-openai/test/openaiResponsesModel.helpers.test.ts
+++ b/packages/agents-openai/test/openaiResponsesModel.helpers.test.ts
@@ -1280,6 +1280,34 @@ describe('getInputItems', () => {
     });
   });
 
+  it('falls back to top-level mimeType for nested image data', () => {
+    const base64 = Buffer.from('top-level-mimetype').toString('base64');
+    const items = getInputItems([
+      {
+        type: 'function_call_result',
+        callId: 'c10',
+        output: {
+          type: 'image',
+          image: {
+            data: base64,
+          },
+          mimeType: 'image/jpeg',
+        },
+      },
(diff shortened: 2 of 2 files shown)


seratch:
I've quickly checked the necessity of adding the same to Python SDK, but the SDK doesn't have this issue.

seratch:
@codex review again

Issue #894, opened 2026-01-20
Define retry policy when generation request timeouts


OrGivati:
We are using gemini using ai-sdk(`@ai-sdk/google-vertex:4.0.15`) and also some of the OpenAI models.
From time to time, some generation request take too long - +300 second(see image).

<img width="486" height="289" alt="Image" src="https://github.com/user-attachments/assets/e61540f1-adc9-4bbf-a4a5-c9dc655addf6" />

My questions:
1. can I control the timeout of the generation request? 
2. can I define retry policy for the agent to apply in case generation fails on timeout? or other exceptions?  

if we will be able to set to the `run` options(`SharedRunOptions`) retry policy on generation request based on the errors type thrown by the provider, it can help improve the reliability of the agent run. 
Both by providing custom timeout I can fail request early and with the retry policy I can recover in case of failures. 

### Debug information

- Agents SDK version: (e.g. `v0.4.0`)
- Runtime environment (e.g. `Node.js 22.19.0`)

seratch:
You can enforce timeouts via `SharedRunOptions.signal`. The ai-sdk adapter forwards it to its `abortSignal`. With node 22, you can do:

```typescript
const result = await run(agent, input, {
  signal: AbortSignal.timeout(30_000),
});
```

If you need broader support, use `AbortController` + `setTimeout` and pass its signal.

There is no built‑in run‑level retry policy in agents-core package right now. It seems ai-sdk supports `maxRetries` in call options, and the adapter already spreads `modelSettings.providerData` into ai-sdk call options — so you can set:

```typescript
const agent = new Agent({
  // ...
  modelSettings: {
    providerData: { maxRetries: 2 },
  },
});
```

For future ideas of the customizable retry policy, we may consider adding new options for it, but I'd like to learn what's the most useful options for many developers, an how ai-sdk could support something similar.

OrGivati:
Thanks!

I will use the provider data to define the timeout and retries. 

I want to clarify, the retry policy and timeouts I want to define are not for the entire agent run, but specific request to the model during the run which take abnormally longer than usual.
Since the agent can run for minutes - losing the entire run because 1 request that failed on timeout / any other known exception is a shame. 
Define retry policy in the SDK level can help a lot to improve the reliability of the application. Although, I get it might be depend on the model provider and its specific exceptions. 
The SDK might help ensure that I won't apply retries on every generation request and might just backoff entirely from the run.

seratch:
[...]

PR #1079, opened 2026-03-09
fix(responses): retry transient terminated transport errors

### LINKED ISSUE #855: TypeError: terminated
We use `openai-agents-js` (default **Responses API**) to run our Multi-Agent Systems (MAS). During MAS runs, we intermittently hit the following error:

```json
{
  "error": "TypeError: terminated"
}
```

We’ve seen this issue for a long time, but the frequency has **increased significantly over the last 2 days**. It happens in **both streamed and non-streamed** MAS runs.

We use the **`gpt-4.1`** and **`gpt-5`** model families across our MASes.

Full error (from our logs):

```txt
"message": "terminated",
2026-01-14 10:07:06.941 CET
"stack": "TypeError: terminated
    at Fetch.onAborted (node:internal/deps/undici/undici:11322:53)
    at Fetch.emit (node:events:524:28)
    at Fetch.terminate (node:internal/deps/undici/undici:10480:14)
    at Object.onError (node:internal/deps/undici/undici:11443:38)
    at Request.onError (node:internal/deps/undici/undici:2210:31)
    at Object.errorRequest (node:internal/deps/undici/undici:1707:17)
    at TLSSocket.<anonymous> (node:internal/deps/undici/undici:6433:16)
    at TLSSocket.emit (node:events:536:35)
    at node:net:343:12
    at TCP.done (node:_tls_wrap:648:7)"
```

Because the SDK uses the Responses API by default, this may be related to the following thread:

https://community.openai.com/t/200-ok-from-v1-responses-but-body-times-out-und-err-body-timeout/1371083

This issue may also be related:

https://github.com/openai/openai-agents-js/issues/744

And this is the corresponding “twin” issue in the `OpenAI Developer Community`:

https://community.openai.com/t/agents-sdk-typeerror-terminated/1371661

If there are any known incidents, rate limits, networking changes, or recommended client settings for these model families that could explain the recent increase, we’d appreciate any guidance. We’d also like to understand whether this error is caused by our system (e.g., timeouts/aborts/network/proxy behavior) or if it’s originating from the OpenAI backend.

SergioChan:
### Summary

- retry `responses.create` once when the request fails with transient transport errors like `TypeError: terminated`
- only retry when the request signal is not aborted, so caller-initiated cancellations still fail immediately
- add a focused unit test that verifies the one-time retry path

### Test plan

- Attempted to run:
  - `pnpm test -- packages/agents-openai/test/openaiResponsesModel.test.ts -t "retries once when responses.create fails with a transient terminated error"`
- This environment does not have workspace dependencies installed (`node_modules` missing), so the command fails before tests execute.
- Manual validation: reviewed the retry guard to ensure retry only happens for transient transport errors and never on caller-aborted signals.

### Issue number

Closes #855

### Checks

- [ ] I've made sure tests pass
- [ ] I've added a changeset using `pnpm changeset` to indicate my changes

### DIFF
--- a/.changeset/lucky-needles-glow.md
+++ b/.changeset/lucky-needles-glow.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents-openai": patch
+---
+
+fix(responses): retry transient terminated transport errors


SergioChan:
Thanks for the signal on changeset validation — I pushed a follow-up commit adding a changeset for @openai/agents-openai (patch). CI should re-run now.

seratch:
Thanks for sending this patch! As I mentioned at https://github.com/openai/openai-agents-js/issues/894#issuecomment-3781973802, I've been thinking of the enhancement to enable developers to pass customizable retry handlers and its configuration such as max-tries etc. I haven't started working on it yet, but we'll consider adding the option in future releases instead of adding fixed retry code within the model code.

SergioChan:
Thanks for the context — makes sense. If you prefer to keep retry behavior configurable and centralized, I’m happy to close this draft and revisit once that extension point is available.

If useful, I can also rework this PR to a smaller non-behavioral change (e.g., improved surfaced error context for terminated transport failures) that won’t lock in retry policy.

seratch:
I've come up with https://github.com/openai/openai-agents-js/pull/1083 for this need. Thank you again for your efforts here.

PR #1090, opened 2026-03-12
fix(agents-core): type raw model stream events precisely

### LINKED ISSUE #279: Change StreamEventGenericItem.event Type from any to OpenAI ResponseStreamEvent
Currently, in the `raw_model_stream_event` type, when the `data.type` field is `"model"`, we expose an `event` field whose type is declared as `any` (courtesy of `StreamEventGenericItem`). This means consumers lose all type-safety and autocompletion on the shape of the event payload.

**Steps to Reproduce:**

1. In the OpenAI client, import or interact with the streaming API.
2. Inspect the TypeScript definition of `raw_model_stream_event`.
3. Note that `data.event` is typed as `any` under the hood.
```
export declare const StreamEventGenericItem: z.ZodObject<{
    /**
     * Additional optional provider specific data. Used for custom functionality or model provider
     * specific fields.
     */
    providerData: z.ZodOptional<z.ZodRecord<z.ZodString, z.ZodAny>>;
} & {
    type: z.ZodLiteral<"model">;
    event: z.ZodAny;
}, "strip", z.ZodTypeAny, {
    type: "model";
    providerData?: Record<string, any> | undefined;
    event?: any;
}, {
    type: "model";
    providerData?: Record<string, any> | undefined;
    event?: any;
}>;
```

**Expected Behavior:**

* The `event` property should be strongly typed to match what the API actually returns for model-stream events.
* We should leverage the existing `ResponseStreamEvent` type defined in `openai/resources/responses/responses.mjs` rather than falling back to `any`.

**Suggested Solution:**

* Update the definition of `StreamEventGenericItem<T>` (or wherever `StreamEventGenericItem` is used) so that when `data.type === 'model'`, the `event` field is typed as `ResponseStreamEvent`.
* Ensure that `ResponseStreamEvent` is properly imported from `openai/resources/responses/responses.mjs`.
* Add any needed unit tests or type assertions to verify that the change provides correct type-checking and does not break existing uses.

**Additional Context:**
The change will improve DX by restoring full type-safety and IntelliSense when handling streaming responses from model endpoints. This aligns with the rest of the client library’s emphasis on precise typings.

fahe1em1:
## Summary
- type `StreamEventGenericItem.event` as the OpenAI `ResponseStreamEvent` union instead of `any`
- keep the runtime schema unchanged while tightening the exported TypeScript surface for SDK consumers
- add a focused type assertion test so the contract stays locked in

## Root Cause
The current protocol schema uses `z.any()` for raw model stream events, and the inferred public TypeScript type inherits that `any`, which strips IntelliSense and type-safety from `raw_model_stream_event.data.event`.

## Why This Is Small And Safe
This patch only tightens the exported TypeScript type. It does not change runtime parsing, event shapes, or streaming behavior, so existing consumers keep the same runtime semantics while gaining more precise typings.

## Testing
- `pnpm --filter @openai/agents-core build-check`
- `pnpm exec prettier --check packages/agents-core/src/types/protocol.ts packages/agents-core/test/run.test.ts`

Closes #279.

### DIFF
--- a/packages/agents-core/src/types/protocol.ts
+++ b/packages/agents-core/src/types/protocol.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import type { ResponseStreamEvent as OpenAIResponseStreamEvent } from 'openai/resources/responses/responses';
 
 // ----------------------------
 // Shared base types
@@ -941,7 +942,10 @@ export const StreamEventGenericItem = SharedBase.extend({
   type: z.literal('model'),
   event: z.any().describe('The event from the model'),
 });
-export type StreamEventGenericItem = z.infer<typeof StreamEventGenericItem>;
+export type StreamEventGenericItem = SharedBase & {
+  type: 'model';
+  event: OpenAIResponseStreamEvent;
+};
 
 export const StreamEvent = z.discriminatedUnion('type', [
   StreamEventTextStream,
--- a/packages/agents-core/test/run.test.ts
+++ b/packages/agents-core/test/run.test.ts
@@ -10,6 +10,7 @@ import {
   vi,
 } from 'vitest';
 import { z } from 'zod';
+import type { ResponseStreamEvent as OpenAIResponseStreamEvent } from 'openai/resources/responses/responses';
 import {
   Agent,
   InputGuardrailTripwireTriggered,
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for sending this patch. Closing this in favor of https://github.com/openai/openai-agents-js/pull/1091

PR #1100, opened 2026-03-14
fix: use V3 file-data/file-url format for tool result images in aisdk adapter

Ema93sh:
## Problem

When using `aisdk()` with a V3 provider (e.g. Vercel AI Gateway), tool results containing images fail with a `400 Invalid input` error. This happens because `convertStructuredOutputsToAiSdkOutput` always emits V2 `{type:'media'}` content parts, even when the model's `specificationVersion` is `v3`.

The `getSpecVersion()` helper already exists and is used elsewhere (e.g. `toolToLanguageV2Tool`), but was not wired into the tool result conversion path.

## Fix

Thread the spec version into `convertToAiSdkOutput` → `convertStructuredOutputsToAiSdkOutput` and emit the correct content part format per version:

| | Base64 image | URL image |
|---|---|---|
| **V2** | `{type:'media', data, mediaType}` | `{type:'media', data, mediaType:'image/*'}` |
| **V3** | `{type:'file-data', data, mediaType}` | `{type:'file-url', url}` |

## Test plan

- Existing V2 test passes unchanged
- New V3 test validates `file-data`/`file-url` output
- All 1677 tests pass
- `pnpm build && pnpm -r build-check && pnpm test && pnpm lint` all green

### DIFF
--- a/.changeset/olive-glasses-cry.md
+++ b/.changeset/olive-glasses-cry.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+Fix tool result image format for AI SDK V3 models. convertStructuredOutputsToAiSdkOutput now uses file-data/file-url content parts for V3 models instead of the V2-only media format, fixing 400 errors with V3 providers like Vercel AI Gateway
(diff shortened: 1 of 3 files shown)


seratch:
At the code level, this does look good. However, I ran a few live API tests and found that it failed with some providers. So while I think it makes sense to migrate eventually, I do not think we should make this change right away.

seratch:
For the above reason, let me close this PR for now. That said, we can revisit this change in the future.

PR #1113, opened 2026-03-21
fix: omit acknowledged_safety_checks property in computer_call items

seratch:
This pull request applies the same workaround with https://github.com/openai/openai-agents-python/issues/2741 to computer using tool calls. The runtime behavioral change is on the server-side, so if the server behavior were reverted quickly, this client-side workaround may not be necessary. I am communicating with the relevant team. I will decide whether this SDK should have this later on.


seratch:
This issue is now fixed on the server side.

PR #1121, opened 2026-03-24
fix: export HANDOFF_REJECTED_MESSAGE constant for parallel handoff rejection

ntbuduma:
## Summary
- Extract the hardcoded "Multiple handoffs detected, ignoring this one." string into a named `HANDOFF_REJECTED_MESSAGE` constant in `handoff.ts`
- Export it from `@openai/agents-core` so consumers can reliably detect when a parallel handoff was rejected without matching on a magic string

## Test plan
- [ ✅ ] Verify `pnpm build` succeeds
- [ ✅ ] Verify `pnpm test` passes with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/export-handoff-rejected-message.md
+++ b/.changeset/export-handoff-rejected-message.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this patch! However, we don't plan to just make the string constant public this way.

PR #1129, opened 2026-03-25
fix(agents-openai): omit empty computer safety checks on replay

seratch:
This pull request fixes a Responses API replay bug for computer tool turns. Replayed `computer_call` items were always sending empty `pending_safety_checks`, which caused follow-up computer tool requests to fail with `400 pending_safety_checks is not supported for the "computer" tool.`

The change updates replay input reconstruction to omit empty `pending_safety_checks` and empty `acknowledged_safety_checks` instead of forcing empty arrays into the payload. It also adds regression coverage for empty computer safety-check fields and includes a patch changeset for `@openai/agents-openai`. This keeps non-empty safety-check data intact while matching the live API behavior required for `computer-use` and `computer-use-hitl` flows.

### DIFF
--- a/.agents/skills/examples-auto-run/scripts/run.sh
+++ b/.agents/skills/examples-auto-run/scripts/run.sh
@@ -184,6 +184,24 @@ for (const item of [...set]) {
 NODE
   }
 
+  load_auto_input() {
+    local name="$1"
+    node --input-type=module - "$ROOT" "$name" <<'NODE'
+import path from 'node:path';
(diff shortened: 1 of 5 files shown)


.agents/skills/examples-auto-run/scripts/run.sh
```diff
@@ -184,6 +184,24 @@ for (const item of [...set]) {
 NODE
   }
 
+  load_auto_input() {
```
seratch, line 187: this is for stabler executions

PR #1130, opened 2026-03-25
feat(agents-extensions): add ToolOutputTrimmer for token-efficient long conversations

mvanhorn:
## Summary

Ports the Python SDK's `ToolOutputTrimmer` to `@openai/agents-extensions`.

Long-running agent conversations accumulate large tool outputs (search results, code execution output) that consume tokens but lose relevance over time. `ToolOutputTrimmer` is a `callModelInputFilter` that replaces bulky tool outputs from older turns with compact previews.

## Usage

```typescript
import { ToolOutputTrimmer } from '@openai/agents-extensions';

const trimmer = new ToolOutputTrimmer({
  recentTurns: 2,       // keep last 2 user messages at full fidelity
  maxOutputChars: 500,   // trim outputs above this threshold
  previewChars: 200,     // preserve this many chars as preview
  trimmableTools: new Set(['search', 'execute_code']),  // optional filter
});

const result = await run(agent, input, {
  callModelInputFilter: trimmer.filter,
});
```

## How it works

1. Walks backward through items counting user messages to find the old/recent boundary
2. For old `function_call_result` items with output exceeding `maxOutputChars`, replaces output with a summary + preview
3. For old `tool_search_output` items, strips verbose descriptions and JSON schema prose from tool definitions
4. Never mutates original items. Respects namespace-qualified tool names in `trimmableTools`.

## Changes

- `packages/agents-extensions/src/toolOutputTrimmer.ts` - new file, the trimmer implementation
- `packages/agents-extensions/src/index.ts` - exports `ToolOutputTrimmer` and `ToolOutputTrimmerOptions`
- `packages/agents-extensions/test/toolOutputTrimmer.test.ts` - 17 tests covering boundary detection, trimming, filtering, immutability, and validation
- `examples/docs/extensions/toolOutputTrimmer.ts` - usage example

## Evidence

Direct parity port from the Python SDK's [`tool_output_trimmer.py`](https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/tool_output_trimmer.py). The JS SDK already has the `callModelInputFilter` infrastructure in `RunConfig`, so this adds a ready-to-use filter that users would otherwise write from scratch.

## Testing

17 unit tests covering:
- Boundary detection (old vs recent turns)
- Output trimming (function_call_result, tool_search_output)
- Tool name filtering (bare names, namespace-qualified names)
- Immutability (original items never mutated)
- Input validation (recentTurns, maxOutputChars, previewChars)
- Edge cases (empty input, single turn, short outputs)

All 1770 tests pass (1753 existing + 17 new).

Post-build dogfooding skipped (score 3/10). Tested via automated test suite only.

This contribution was developed with AI assistance (Claude Code).


.changeset/tool-output-trimmer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-extensions': minor
```
seratch: can you use patch for this?

examples/docs/extensions/toolOutputTrimmer.ts
```diff
@@ -0,0 +1,19 @@
+import { Agent, run } from '@openai/agents';
```
seratch: this is not used; can you remove this?

seratch (changes requested):

mvanhorn:
Fixed in 4e01b2d: changeset downgraded to `patch`, removed unused example file.

.changeset/tool-output-trimmer.md
mvanhorn: Done in 4e01b2d.

examples/docs/extensions/toolOutputTrimmer.ts
mvanhorn: Removed in 4e01b2d.

travisbreaks (commented):
Port from the Python SDK. A few things to verify:

1. **Multi-byte character handling**: Does `maxOutputChars` count bytes or characters? If characters (which it should be, given the name), string slicing with `.slice(0, maxOutputChars)` is correct. But if the preview is later measured by token count, a truncation at a character boundary could split a multi-byte sequence in a way that inflates token count. Probably fine in practice, but worth a note.

2. **"Older" vs "recent" turn identification**: The heuristic for deciding which turns are "older" and eligible for trimming vs "recent" and preserved matters a lot for agent reasoning. If the trimmer aggressively trims tool outputs from a turn that the agent is about to reference in its chain-of-thought, reasoning quality degrades silently. Does the Python implementation have a lookback window or does it trim purely by position?

3. **Nested tool calls**: If a tool call returns structured output that includes references to other tool calls (e.g., an orchestrator pattern), trimming the outer output could leave dangling references. Does the test suite cover this scenario?

4. **Preview generation**: The truncated preview should ideally preserve the beginning and end of the output (head + tail) rather than just the head, since error messages and return values often appear at the end. Does the implementation do this, or is it a pure prefix truncation?

Good to see the changeset type addressed per seratch's feedback.

mvanhorn:
`maxOutputChars` counts JavaScript characters via `.slice()`, not bytes -- so multi-byte characters like emoji are counted as their `.length` (which for most BMP characters is 1, but surrogate pairs like emoji count as 2). This matches the Python SDK's behavior with `len()` on strings.

For token counting, truncation at a character boundary is safe -- the tokenizer handles any valid UTF-8 string. A mid-codepoint split can't happen because `.slice()` operates on UTF-16 code units, not bytes.

PR #1164, opened 2026-04-14
docs(readme): fix subpackage install commands

wsk-builds:
## Summary

This pull request fixes incorrect installation snippets in the published subpackage READMEs for `@openai/agents-core`, `@openai/agents-openai`, and `@openai/agents-realtime`.

## Background

Each of these READMEs currently tells users to install `@openai/agents`, which is the umbrella package rather than the package documented on that page. That is misleading for consumers who intentionally want the lower-level packages directly, and it also omits `zod` from the install snippets.

## Changes

- update `packages/agents-core/README.md` to install `@openai/agents-core zod`
- update `packages/agents-openai/README.md` to install `@openai/agents-openai zod`
- update `packages/agents-realtime/README.md` to install `@openai/agents-realtime zod`
- add a patch changeset covering the three affected packages

## Impact

- aligns each subpackage README with the actual published package name
- reduces installation confusion for users consuming lower-level packages directly
- ensures the next release notes capture this package-level documentation correction

## Validation

- ran `pnpm changeset:validate-prompt`
- did not run the full verification stack because this is a docs-only README change

### DIFF
--- a/.changeset/green-steaks-mix.md
+++ b/.changeset/green-steaks-mix.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-core': patch
+'@openai/agents-openai': patch
+'@openai/agents-realtime': patch
+---
+
+fix: correct subpackage README install commands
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for sending this. However, this is intentional, so let use close this PR.

PR #1167, opened 2026-04-14
docs(examples): fix docs snippet commands

wsk-builds:
## Summary
- replace the nonexistent `pnpm -F docs start:*` commands in `examples/docs/README.md`
- point each README entry at the real snippet file path in `examples/docs/**`
- document `pnpm exec tsx examples/docs/...` as the supported way to run these snippets from the repo root

## Why
`examples/docs/package.json` only defines `build-check`, so the old README commands were not runnable. This change uses commands that are directly verifiable against files that actually exist in the repository, which keeps the README accurate without adding a large set of maintenance-only package scripts.

## Validation
- verified every snippet path linked from `examples/docs/README.md` exists
- verified `pnpm exec tsx --version` resolves successfully (`tsx v4.21.0`, `node v22.22.2`)

### DIFF
--- a/examples/docs/README.md
+++ b/examples/docs/README.md
@@ -1,80 +1,80 @@
 # Documentation Snippets
 
-This directory contains small scripts used throughout the documentation. Run them with `pnpm` using the commands shown below.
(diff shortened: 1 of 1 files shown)


seratch:
Thanks but I see the necessity of these changes.

PR #1172, opened 2026-04-15
fix(agents-core): restore discriminated union tool schemas

### LINKED ISSUE #1138: Support for discrminated union type dropped
### Please read this first

- [X] **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/)
- [X] **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug

Using z.discriminatedUnion() inside tool parameters results in a 400 error from the API.

```
Running agent with z.discriminatedUnion() tool...
ERROR (expected — this is the bug):
400 Invalid schema for function 'create_event': In context=('properties', 'recurrence'), 'oneOf' is not permitted.
```

I noticed this when upgrading from v0.3.9, where it was working previously.

### Debug information

- Agents SDK version: v0.8.1
- Runtime environment: Node.js 22.x

### Repro steps

```
import "dotenv/config";
import { tool, run, Agent } from "@openai/agents";
import { z } from "zod";

const myTool = tool({
  name: "create_event",
  description: "Create a calendar event",
  parameters: z.object({
    name: z.string(),
    recurrence: z.discriminatedUnion("type", [
      z.object({ type: z.literal("once"), date: z.string() }),
      z.object({ type: z.literal("weekly"), dayOfWeek: z.number() }),
    ]),
  }),
  execute: async (input) => {
    return `Created: ${input.name}`;
  },
});

const agent = new Agent({
  name: "scheduler",
  instructions: "You schedule events.",
  tools: [myTool],
});

console.log("Running agent with z.discriminatedUnion() tool...");
try {
  const result = await run(agent, "Create a weekly meeting on Mondays");
  console.log("Result:", result.finalOutput);
} catch (err) {
  console.error("ERROR (expected — this is the bug):");
  console.error(err.message ?? err);
  console.log("\nWorkaround: replace z.discriminatedUnion() with z.union()");
}
```

### Expected behavior

This appears to be an error with the conversion from Zod schema to JSON Schema. Expected behavior is that the conversion occurs successfully and we can use discriminatedUnion in Zod schema.

wsk-builds:
This pull request resolves #1138.

Zod discriminated unions can still fail tool schema generation when compatibility fallbacks see legacy internals or when upstream helper output includes unsupported `oneOf` branches. In those cases the API rejects the function schema with a 400 before the tool can run.

This change:
- teaches `zodJsonSchemaCompat` to treat legacy discriminated-union internals like other unions so the fallback path still produces a valid schema
- normalizes helper-generated `oneOf` branches into `anyOf` before returning tool and structured-output schemas to OpenAI
- adds regression coverage for the compat converter and the tool schema generation path

Behavior impact:
- tool parameters built from `z.discriminatedUnion()` now stay within the OpenAI-supported schema subset
- the fix is intentionally narrow and only rewrites `oneOf` into `anyOf` for generated schemas

### DIFF
--- a/.changeset/green-maps-laugh.md
+++ b/.changeset/green-maps-laugh.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
-fix: restore discriminated union tool schema compatibility (#1138)
+fix: handle legacy discriminated union fallback schemas
--- a/.changeset/green-maps-laugh.md
+++ b/.changeset/green-maps-laugh.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: restore discriminated union tool schema compatibility (#1138)
--- a/packages/agents-core/src/utils/tools.ts
+++ b/packages/agents-core/src/utils/tools.ts
@@ -8,6 +8,7 @@ import {
   zodJsonSchemaCompat,
   hasJsonSchemaObjectShape,
   mergeJsonSchemaDescriptions,
+  normalizeOpenAiJsonSchema,
 } from './zodJsonSchemaCompat';
 import type { ZodObjectLike } from './zodCompat';
 import { asZodType } from './zodCompat';
@@ -60,7 +61,8 @@ export type FunctionToolName = string & { __brand?: 'ToolName' } & {
 function buildJsonSchemaFromZod(
   inputType: ZodObjectLike,
 ): JsonObjectSchema<any> | undefined {
-  return zodJsonSchemaCompat(inputType);
+  const schema = zodJsonSchemaCompat(inputType);
+  return schema ? normalizeOpenAiJsonSchema(schema) : undefined;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


seratch (changes requested):
Thanks for taking another pass at this. I agree this PR is materially different from #1139 in one important way: it does not bring back the earlier "flatten unions into nullable strict objects" strategy, so the runtime-shape and token-cost concerns from that thread do not directly apply here.

That said, I am still not convinced this should merge as-is.

- On current `main` with current dependencies, the common Zod v3/v4 `discriminatedUnion()` and `union()` cases appear to already serialize to `anyOf` through `openai/helpers/zod`, for both tool parameters and structured outputs. That matches the live API observations called out in #1139.
- The fallback-side change to handle legacy `discriminatedunion` internals seems reasonable on its own.
- The broader `normalizeOpenAiJsonSchema()` walk is harder to justify. Rewriting `oneOf` into `anyOf` recursively is not semantics-preserving in general, and this PR applies that rewrite to structured outputs as well, not just tool parameters.
- The added tests show that the rewrite happens, but they do not yet demonstrate a minimal repro on current `main` where this rewrite is actually required.

Before merging, I would want to see one of these:

1. a minimal repro on current `main` / current dependency versions that still emits `oneOf` and fails against the API, or
2. the PR narrowed to the smallest demonstrated fix, for example just the legacy fallback `discriminatedunion` handling if that is the real gap.

As written, this feels closer to a broad schema-normalization policy than a tightly justified bug fix, so I would prefer to hold off on merging until the need and scope are clearer.

seratch (approved):
This is a quite small improvement now, but there is no reason to avoid having it

PR #1175, opened 2026-04-16
fix(realtime): replay assistant audio history from transcripts

### LINKED ISSUE #963: session.updateHistory doesn't support audio turns
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/) Yes
- **Have you searched for related issues?** Others may have faced similar issues. Yes

### Describe the bug

I'm trying to store a realtime conversation while it's in progress, and restore it later. When I access `session.history`, I get RealtimeItems that look like this:

```ts
[{
      itemId: 'item_123',
      type: 'message',
      role: 'assistant',
      status: 'completed',
      content: [
        {
          type: 'output_audio',
          transcript: 'Thanks for calling. How can I help?',
          audio: null,
        },
      ],
    }]
```

When I try to load this with `session.updateHistory`, I get the following error:

```
 Error [ERR_UNHANDLED_ERROR]: Unhandled error. ({
   type: 'error',
   error: {
     type: 'error',
     event_id: 'event_D8sDiiDisPUHKG1YVD3Cl',
     error: {
       type: 'invalid_request_error',
       code: 'missing_required_parameter',
       message: "Missing required parameter: 'item.content[0].text'.",
       param: 'item.content[0].text',
       event_id: null
     }
   }
 })
```

### Debug information

- Agents SDK version: `v0.4.8`
- Runtime environment Node.js v22.20.0

### Repro steps

```ts
    const myItem = {
      itemId: 'item_123',
      type: 'message' as const,
      role: 'assistant' as const,
      status: 'completed' as const,
      content: [
        {
          type: 'output_audio' as const,
          transcript: 'Thanks for calling. How can I help?',
          audio: null,
        },
      ],
    };
    session.updateHistory([myItem]);
```

### Expected behavior

I'd like to be able to restore a conversation with the exact data structures in `session.history` (function calls aside)

In the meantime, if anyone knows what format `updateHistory` accepts, that would be a great workaround

wsk-builds:
## Summary
- fix `session.updateHistory()` replay for assistant audio turns by converting transcript-backed `output_audio` history into replayable assistant text
- skip assistant audio turns that do not have transcripts instead of sending invalid `conversation.item.create` payloads
- add regression tests and document that assistant audio replay is transcript-backed rather than raw audio restoration

## Why
This pull request resolves #963.

`RealtimeSession.history` can contain assistant `output_audio` items, but feeding that history back into `session.updateHistory()` currently fails because the Realtime API does not accept assistant audio messages in `conversation.item.create`. This change preserves round-tripping for assistant turns when a transcript is available and narrows the unsupported case to transcript-less audio replay.

## What Changed
- added `realtimeMessageToConversationItemCreateMessage(...)` to normalize replayed history items into a shape accepted by `conversation.item.create`
- updated `OpenAIRealtimeBase.resetHistory()` to replay assistant audio history as `output_text` using `output_audio.transcript`
- log and skip assistant audio items that do not have transcripts instead of emitting an invalid request
- added regression coverage in the realtime utils/base tests for transcript-backed replay and transcript-missing skips
- updated the voice agents guide to explain that assistant audio history replay depends on transcripts and does not restore audio blobs

## Testing
- `pnpm vitest run packages/agents-realtime/test/utils.test.ts packages/agents-realtime/test/openaiRealtimeBase.test.ts`
- `pnpm changeset:validate-prompt`

## Risks / Follow-ups
- assistant audio turns without transcripts still cannot be replayed because the upstream Realtime API does not currently support creating assistant audio history directly

### DIFF
--- a/.changeset/audio-history-replay.md
+++ b/.changeset/audio-history-replay.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: replay assistant audio history from transcripts
--- a/docs/src/content/docs/guides/voice-agents/build.mdx
+++ b/docs/src/content/docs/guides/voice-agents/build.mdx
@@ -254,14 +254,14 @@ Set `debounceTextLength: -1` when you only want to evaluate the fully generated 
 
 `RealtimeSession` automatically maintains a local `history` snapshot that tracks user messages, assistant output, tool calls, and truncation state. You can render it in the UI, inspect it inside tools, or update it when you need to correct or remove items.
 
-As the conversation changes the session emits `history_updated`. If you need to request history changes, use `updateHistory()`. It asks the transport to diff the current history and send the necessary delete/create events; the local `session.history` view updates as the corresponding conversation events come back.
+As the conversation changes the session emits `history_updated`. If you need to request history changes, use `updateHistory()`. It asks the transport to diff the current history and send the necessary delete/create events; the local `session.history` view updates as the corresponding conversation events come back. Because the Realtime API does not currently accept assistant audio items in `conversation.item.create`, replaying assistant audio history converts transcript-backed `output_audio` entries into assistant text and skips assistant audio turns that do not have transcripts.
 
 <Code lang="typescript" code={updateHistoryExample} />
 
 #### Limitations
 
 1. You cannot currently edit function tool calls after the fact.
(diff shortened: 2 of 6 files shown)


seratch:
Have you verified if this really works with the production API endpoint? My understanding is that the enhancement couldn't be done only by this client-side SDK efforts.

PR #1184, opened 2026-04-19
fix(twilio): #1173 send clear message on interrupt when audio is fully flushed

### LINKED ISSUE #1173: Twilio Extension Fails to interrupt playback if audio has already been fully sent.
### Please read this first

- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug

During a longer response (think e.g. a legal disclaimer with +10s of playback), it seems the `session.interrupt()` (which then calls the `transport.interrupt`) does not actually interrupt Twilio playback.

It seems that the TwilioRealtimeTransport only implements [_interrupt](https://github.com/openai/openai-agents-js/blob/d17da194ec18f2fb88aa2fea1dc3054a0f185e11/packages/agents-extensions/src/TwilioRealtimeTransport.ts#L188). However, the super class [interrupt](https://github.com/openai/openai-agents-js/blob/d17da194ec18f2fb88aa2fea1dc3054a0f185e11/packages/agents-realtime/src/openaiRealtimeWebsocket.ts#L500) silently aborts if it has already submitted all the audio, never calling the `_interrupt`.

The variables seem to be reset [here](https://github.com/openai/openai-agents-js/blob/d17da194ec18f2fb88aa2fea1dc3054a0f185e11/packages/agents-realtime/src/openaiRealtimeWebsocket.ts#L112), which in turn gets called on [response.output_audio.done](https://github.com/openai/openai-agents-js/blob/d17da194ec18f2fb88aa2fea1dc3054a0f185e11/packages/agents-realtime/src/openaiRealtimeBase.ts#L263).

However, Twilio uses marks to indicate when playback has finished, so if the generation + submission to Twilio takes 1s, but actual response is 10s in length, you only have a 1s window to interrupt correctly clearing the response stream.

### Debug information

- Agents SDK version: 0.8.3
- Runtime environment (e.g. `Node.js 22.16.0`)

### Repro steps

Give the agent an instruction like
```
Read the following disclaimer:
All the information on this website is published in good faith and for general information purpose only. Website Name does not make any warranties about the completeness, reliability and accuracy of this information. Any action you take upon the information you find on this website (Website.com), is strictly at your own risk. will not be liable for any losses and/or damages in connection with the use of our website.
```
Create your Twilio sessions, and then register an event listener like so:
```
this.session.on('transport_event', async (transportEvent) => {
  if (transportEvent.type === 'input_audio_buffer.speech_started') {
    try {
            this.session.interrupt();
     } catch (interruptErr) {
            console.warn(`interrupt() failed (race condition):`, interruptErr?.message || interruptErr);
     }
  }
}
```
Then after connecting start with
```
this.session.transport.sendEvent({ type: 'response.create' });
```
so the agent will talk first.

Let the agent talk for a bit, then interrupt the agent. Notice that the agent keeps on talking if the interruption does not happen during generation or during transfer of audio to Twilio.

Now, add this code after (or before) the interrupt code above:
```
if (this.twilioWebSocket) {
  this.twilioWebSocket.send(JSON.stringify({ event: 'clear', streamSid: this.payload?.streamSid }));
}
```

Notice the playback now gets interrupted.

### Expected behavior

The expectation is that sending an interrupt to the Twilio Realtime Session will interrupt audio playback.

Likely fix will involve adding the top level interrupt method to TwilioRealtimeTransport which would do the clear like so:
```
interrupt(cancelOngoingResponse: boolean = true) {
    // ALWAYS clear the Twilio buffer immediately when interrupted,
    // even if OpenAI has already finished generating the response.
    this.#twilioWebSocket.send(
      JSON.stringify({
        event: 'clear',
        streamSid: this.#streamSid,
      }),
    );
    super.interrupt(cancelOngoingResponse);
  }
```
and removing the clear from the _interrupt (though think there is no harm clearing twice).

xodn348:
## Problem

Fixes #1173.

When all audio chunks have already been flushed to Twilio before an interruption is triggered (e.g. the user speaks immediately after the AI finishes talking), the parent `interrupt()` method in `OpenAIRealtimeWebSocket` returns early because the audio queue is empty. This means `_interrupt()` is never called, so no Twilio `clear` message is sent — leaving Twilio actively playing audio that should have been stopped.

## Root Cause

`openaiRealtimeWebsocket.ts` `interrupt()` at line 500 guards with an early return when no audio is queued:

```ts
// parent class — simplified
interrupt(cancelOngoingResponse = true) {
  if (this.audioQueue.length === 0) return; // ← early exit, _interrupt() never fires
  this._interrupt(...);
}
```

`TwilioRealtimeTransportLayer._interrupt()` already sends the `clear` message, but it is only reached when audio is still buffered.

## Fix

Override the public `interrupt()` method in `TwilioRealtimeTransportLayer` to **always** send the Twilio `clear` message before delegating to `super.interrupt()`:

```ts
interrupt(cancelOngoingResponse: boolean = true): void {
  this.#twilioWebSocket.send(
    JSON.stringify({
      event: 'clear',
      streamSid: this.#streamSid,
    }),
  );
  super.interrupt(cancelOngoingResponse);
}
```

This guarantees active Twilio playback is stopped regardless of whether the audio queue is empty or not. The existing `_interrupt()` override is preserved unchanged and continues to handle mid-stream interruptions with the correct elapsed-time calculation.

## Testing

- LSP diagnostics on the modified file: **0 errors**
- Security scan (`codesure`): **0 findings**
- Build-check errors in the fork are pre-existing (missing `node_modules`) and unrelated to this change

### DIFF
--- a/packages/agents-extensions/src/TwilioRealtimeTransport.ts
+++ b/packages/agents-extensions/src/TwilioRealtimeTransport.ts
@@ -203,6 +203,16 @@ export class TwilioRealtimeTransportLayer extends OpenAIRealtimeWebSocket {
     super.updateSessionConfig(newConfig);
   }
 
+  interrupt(cancelOngoingResponse: boolean = true): void {
+    this.#twilioWebSocket.send(
+      JSON.stringify({
+        event: 'clear',
+        streamSid: this.#streamSid,
+      }),
+    );
+    super.interrupt(cancelOngoingResponse);
+  }
+
   _interrupt(_elapsedTime: number, cancelOngoingResponse: boolean = true) {
     const elapsedTime = this.#lastPlayedChunkCount + 50; /* 50ms buffer */
     this.#logger.debug(


seratch:
I haven't checked this in depth, but can you resolve the CI errors and pass Codex's code reviews first?

packages/agents-extensions/src/TwilioRealtimeTransport.ts
wsk-builds, line 1: Thanks again for working on this. I re-checked the current diff and I think the direction is right: the bug in #1173 happens when the base `interrupt()` returns before `_interrupt()` can send Twilio `clear`, so handling the public `interrupt()` path makes sense.

Before merge, I think a few things should be tightened:

1. Please add a regression test for the public `transport.interrupt()` path, especially the case where OpenAI-side playback state has already been reset / no longer has enough state for the base `interrupt()` path to reach `_interrupt()`. The existing `_interrupt()` tests are useful, but they do not cover the exact failure mode fixed here.

2. This currently sends Twilio `clear` in `interrupt()`, then can send another `clear` from `_interrupt()` if `super.interrupt()` continues into the normal interruption path. It would be cleaner to either avoid the duplicate send or explicitly document/verify that duplicate `clear` messages are harmless for Twilio Media Streams.

3. `#streamSid` is nullable until the Twilio `start` event is received. With the current implementation, calling `interrupt()` before `start` can send `{ streamSid: null }`. I think this should either be guarded, logged/skipped, or covered by a test if that payload is intentionally acceptable.

4. Since this changes behavior in `packages/agents-extensions`, this should include a patch changeset for `@openai/agents-extensions`.

Suggested local checks after the updates:

```bash
pnpm i
pnpm -F @openai/agents-extensions build-check
CI=1 pnpm test -- packages/agents-extensions/test/TwilioRealtimeTransport.test.ts
```

Overall, I think this is solving the right problem, but the public interrupt-path test and the clear edge cases would make the fix much safer to merge.

tmmvn:
FYI: We have found a further issue with the interrupt system. While adding this clear will flush any interrupted audio, as OpenAI Realtime thinks it already spoke everything, there are no truncate events, which will cause the model to hallucinate. It seems the Python agents SDK has a more comprehensive wrapper that for some reason has not made its way to the javascript agents SDK.

In addition, it seems the JavaScript SDK is missing parity with the Python SDK in that the wrapper in the Python SDK for example prevents sending incomplete frames to Twilio. Twilio ulaw expects a fixed chunk buffer, otherwise the audio will have jitter, especially in the start.

To fix all the issues, we ended up overriding the Twilio Transport with our own enhanced version. However, as it seems the websocket Twilio integration in the JS library was not ready for prime-time, we ended up switching back to SIP transport (we had some connection latency issues we were able to sort).

Here is the working fixed transport layer if you want to bring any of the enhancements over. Let me know if you want me to also post this to the issue ticket, or open a new one. I probably won't have time to implement the fixes to the `TwilioRealtimeTransportLayer` and create a PR with tests anytime soon.

```
import { TwilioRealtimeTransportLayer } from '@openai/agents-extensions';
import { utils } from '@openai/agents/realtime';

const GATED_EVENT_TYPES = new Set([
  'response.content_part.done',
  'conversation.item.done',
  'response.output_item.done',
  'response.done',
]);

// g711_ulaw: 8000 samples/sec, 1 byte per sample => 8 bytes per ms
const BYTES_PER_MS_G711 = 8;

export class BufferedTwilioTransport extends TwilioRealtimeTransportLayer {
  constructor(options) {
    // Wrap the Twilio WebSocket with a minimal Proxy that only intercepts
    // done: marks sent by the parent's audio_done handler in connect().
    // Audio is handled by _onAudio() override instead.
    const originalWs = options.twilioWebSocket;
    const self = { instance: null };
    const wrappedWs = new Proxy(originalWs, {
      get(target, prop) {
        if (prop === 'send') {
          return (data) => {
            try {
              const msg = typeof data === 'string' ? JSON.parse(data) : data;
              if (msg.event === 'mark' && msg.mark && msg.mark.name && msg.mark.name.startsWith('done:') && self.instance) {
                // Queue the done: mark at the tail of our audio buffer so
                // it reaches Twilio only after all buffered audio has been sent.
                self.instance._eventQueue.push({
                  byteOffset: self.instance._audioBuffer.length,
                  data: msg,
                });
                return;
              }
            } catch (e) { /* parse error — pass through */ }
            originalWs.send(data);
          };
        }
        const value = Reflect.get(target, prop);
        return typeof value === 'function' ? value.bind(target) : value;
      }
    });

    super({ ...options, twilioWebSocket: wrappedWs });

    // Wire up the self reference so the Proxy closure can access our live state
    self.instance = this;

    this.originalWs = originalWs;
    this.streamSid = options.streamSid;
    this._audioBuffer = Buffer.alloc(0);
    this._eventQueue = [];

    this._playoutInterval = null;
    this._isPlayoutActive = false;
    this._targetBufferBytes = 160; // 160 bytes (1 chunk) for ongoing buffering
    this._initialBufferBytes = 160 * 30; // 4800 bytes (30 chunks worth for initial buffer)
    this._chunkSize = 160;
    this._flushTimeout = null;
    this._isInitial = true;
    // ms-based tracking (mirrors Python SDK's RealtimePlaybackTracker)
    this._generatedAudioMs = 0;
    this._playedAudioMs = 0;
    this._currentItem = null;

    // Playback tail state tracking
    this._isInPlaybackTail = false;
    this._pendingItemId = null;
    this._pendingResponseId = null;
    this._gatedEvents = [];
    this._gateTimeoutId = null;
    this._audioGenerationEndTime = null;

    // Listen for Twilio done: marks to detect playback completion.
    // Registered here (before the RealtimeSession registers its * listener)
    // so our listener fires FIRST — allowing us to synchronously replay
    // gated events before the session sees the done: mark.
    this.on('*', (event) => {
      if (
        event.type === 'twilio_message' &&
        event.message?.event === 'mark' &&
        event.message?.mark?.name?.startsWith('done:')
      ) {
        if (this._isInPlaybackTail) {
          // Validate that this done: mark is for the item we're actually
          // waiting on. Stale marks from a previously-interrupted response
          // can echo back from Twilio and must be ignored.
          const markItemId = event.message.mark.name.substring(5); // strip 'done:'
          if (markItemId === this._pendingItemId) {
            this._onPlaybackComplete();
          } else {
            console.log(
              `[BufferedTwilioTransport] Ignoring stale done: mark for item ${markItemId} (waiting for ${this._pendingItemId})`,
            );
          }
        }
      }
    });
  }

  // ─── Override _onAudio directly instead of Proxy ──────────────────────
  // TwilioRealtimeTransportLayer._onAudio() sends media+mark to the Twilio
  // WebSocket. We override it to intercept audio into our pacing buffer
  // instead, and forward marks as byte-offset entries in the event queue.
  _onAudio(audioEvent) {
    const payload = utils.arrayBufferToBase64(audioEvent.data);
    const chunk = Buffer.from(payload, 'base64');
    this._audioBuffer = Buffer.concat([this._audioBuffer, chunk]);
    const chunkMs = chunk.length / BYTES_PER_MS_G711;
    this._generatedAudioMs += chunkMs;
    // Prevent buffer overflow by starting playout if buffer exceeds 10 chunks (1600 bytes)
    if (!this._isPlayoutActive && this._audioBuffer.length > 160 * 10) {
      this._startPlayout();
    }
    this._startPlayout();

    // Queue the per-chunk mark at the correct byte offset so it is sent
    // to Twilio only after the corresponding audio has been forwarded.
    // We reuse the parent's mark naming convention (itemId:cumulativeMs)
    // so that TwilioRealtimeTransportLayer's mark handler can still update
    // its internal lastPlayedChunkCount via the existing case branch.
    const markMsg = {
      event: 'mark',
      streamSid: this.streamSid,
      mark: {
        name: `${this.currentItemId}:${Math.round(this._generatedAudioMs)}`,
      },
    };
    this._eventQueue.push({
      byteOffset: this._audioBuffer.length,
      data: markMsg,
    });

    // Emit the audio event so any external listeners still receive it
    this.emit('audio', audioEvent);
  }

  _clearBuffer() {
    this._audioBuffer = Buffer.alloc(0);
    this._eventQueue = [];
    this._isPlayoutActive = false;
    if (this._playoutInterval) {
      clearTimeout(this._playoutInterval);
      this._playoutInterval = null;
    }
    if (this._gateTimeoutId) {
      clearTimeout(this._gateTimeoutId);
      this._gateTimeoutId = null;
    }
    if (this._flushTimeout) {
      clearTimeout(this._flushTimeout);
      this._flushTimeout = null;
    }
    this._generatedAudioMs = 0;
    this._playedAudioMs = 0;
    this._currentItem = null;
    this._isInitial = true;
  }

  _interrupt(elapsedTime, cancelOngoingResponse = true) {
    super._interrupt(elapsedTime, cancelOngoingResponse);
  }

  interrupt(cancelOngoingResponse = true) {
    const wasInPlaybackTail = this._isInPlaybackTail;

    // Snapshot ms-based tracking before clearing
    const playedMs = this._playedAudioMs;
    const generatedMs = this._generatedAudioMs;

    this._clearBuffer();
    this.originalWs.send(
      JSON.stringify({
        event: 'clear',
        streamSid: this.streamSid,
      }),
    );

    if (wasInPlaybackTail) {
      // ── Smart truncation skip (from Python SDK) ──────────────────
      // If the user has heard (nearly) all of the audio AND the response
      // is already complete, there is nothing to truncate — the model's
      // conversation history is already accurate.
      const fullyPlayed = generatedMs > 0 && playedMs >= generatedMs - 20; // 20ms tolerance
      if (fullyPlayed) {
        console.log(
          `[BufferedTwilioTransport] Skipping truncation for item ${this._pendingItemId}: ` +
          `audio fully played (${Math.round(playedMs)}ms / ${Math.round(generatedMs)}ms)`,
        );
        // Release gated events normally — nothing was cut short
        this._onPlaybackComplete();
        return;
      }

      // ── Skip truncation if no audio was played ────────────────────
      // If the user interrupted before any audio was played, don't truncate
      // at 0ms (which would remove the entire transcript). Leave it as-is.
      if (playedMs <= 0) {
        console.log(
          `[BufferedTwilioTransport] Skipping truncation for item ${this._pendingItemId}: ` +
          `no audio played yet (${Math.round(playedMs)}ms / ${Math.round(generatedMs)}ms)`,
        );
        // Release gated events normally — nothing was cut short
        this._onPlaybackComplete();
        return;
      }

      // Process response.output_item.done for the interrupted item to
      // populate transcript in the session history
      for (const evt of this._gatedEvents) {
        try {
          const raw = typeof evt.data === 'string' ? evt.data : Buffer.from(evt.data).toString();
          const parsed = JSON.parse(raw);
          if (parsed.type === 'response.output_item.done' && parsed.item_id === this._pendingItemId) {
            super._onMessage(evt);
            this._currentItem = parsed.item;
          }
        } catch (e) {
          // Ignore parsing errors
        }
      }

      // ── Server-side truncation via conversation.item.truncate ─────
      // Use the ms-based elapsed time (how much Twilio actually played)
      // and let the server handle transcript truncation, rather than
      // doing a client-side proportional substring which is inaccurate
      // because speech speed varies within a single response.
      const truncateMs = Math.max(0, Math.floor(playedMs));
      console.log(
        `[BufferedTwilioTransport] Interrupting during playback tail for item ${this._pendingItemId}: ` +
        `truncating at ${truncateMs}ms (played ${Math.round(playedMs)}ms / generated ${Math.round(generatedMs)}ms)`,
      );

      // Find the correct content_index for the audio content
      let audioContentIndex = 0;
      if (this._currentItem && this._currentItem.content) {
        audioContentIndex = this._currentItem.content.findIndex(c => c.type === 'audio');
        if (audioContentIndex === -1) audioContentIndex = 0; // fallback if not found
      }

      // Discard any remaining gated events — the response is being interrupted
      this._gatedEvents = [];
      this._isInPlaybackTail = false;
      if (this._gateTimeoutId) {
        clearTimeout(this._gateTimeoutId);
        this._gateTimeoutId = null;
      }

      // Emit audio_interrupted so listeners know playback stopped
      this.emit('audio_interrupted');

      // Send conversation.item.truncate with actual ms played.
      // Do this directly rather than calling super.interrupt() because
      // the base class has already reset its state (currentItemId etc.)
      // and would be a no-op. The response is already complete so we
      // don't need to cancel it.
      this.sendEvent({
        type: 'conversation.item.truncate',
        item_id: this._pendingItemId,
        content_index: audioContentIndex,
        audio_end_ms: truncateMs,
      });

      this._pendingItemId = null;
      this._pendingResponseId = null;
      this._audioGenerationEndTime = null;

      // Let the base class reset its internal state
      super._afterAudioDoneEvent();
    } else {
      super.interrupt(cancelOngoingResponse);
    }
  }

  _afterAudioDoneEvent() {
    // Save the current item/response IDs before the base class would clear them
    if (!this._pendingItemId) {
      this._pendingItemId = this.currentItemId;
    }

    this._isInPlaybackTail = true;
    this._audioGenerationEndTime = Date.now();

    // Safety timeout — if we never get the Twilio done: mark, release after 30s
    this._gateTimeoutId = setTimeout(() => {
      if (this._isInPlaybackTail) {
        console.warn(
          `[BufferedTwilioTransport] Safety timeout: releasing gated events for item ${this._pendingItemId} after 30s`,
        );
        this._onPlaybackComplete();
      }
    }, 30000);

    // DO NOT call super._afterAudioDoneEvent() here — that would reset
    // currentItemId and make interrupt() a no-op during the playback tail.
  }

  _onMessage(event) {
    try {
      const raw =
        typeof event.data === 'string'
          ? event.data
          : Buffer.from(event.data).toString();
      const parsed = JSON.parse(raw);

      // Capture the response ID from the audio-done event (fires before
      // _afterAudioDoneEvent puts us into playback-tail state).
      if (parsed.type === 'response.output_audio.done') {
        this._pendingResponseId = parsed.response_id;
        this._pendingItemId = parsed.item_id;
      }

      if (this._isInPlaybackTail && GATED_EVENT_TYPES.has(parsed.type)) {
        // Let function_call output items through immediately so tool
        // execution isn't blocked by audio playback.
        if (
          parsed.type === 'response.output_item.done' &&
          parsed.item?.type === 'function_call'
        ) {
          super._onMessage(event);
          return;
        }

        // Let content parts through immediately so transcripts build up
        // even during playback tail, allowing partial transcripts on interrupt.
        if (parsed.type === 'response.content_part.done') {
          super._onMessage(event);
          return;
        }

        // Gate this event until Twilio confirms playback completion
        this._gatedEvents.push(event);
        return;
      }
    } catch (e) {
      // Parsing failed — fall through to default handling
    }

    super._onMessage(event);
  }

  _onPlaybackComplete() {
    if (!this._isInPlaybackTail) return;

    const events = this._gatedEvents;
    this._gatedEvents = [];
    this._isInPlaybackTail = false;

    if (this._gateTimeoutId) {
      clearTimeout(this._gateTimeoutId);
      this._gateTimeoutId = null;
    }

    console.log(
      `[BufferedTwilioTransport] Playback complete for item ${this._pendingItemId}, playback tail was ${Date.now() - this._audioGenerationEndTime}ms`,
    );

    this._pendingItemId = null;
    this._pendingResponseId = null;
    this._audioGenerationEndTime = null;
    this._isInitial = true; // Reset for fresh buffer on next response
    this._generatedAudioMs = 0; // Reset ms tracking for next response
    this._playedAudioMs = 0;

    // NOW let the base class reset its internal state (currentItemId etc.)
    super._afterAudioDoneEvent();

    // Replay gated events so the session processes them in order
    for (const evt of events) {
      super._onMessage(evt);
    }
  }

  _startPlayout() {
    if (this._playoutInterval) return;

    const pacingIntervalMs = 19.5;
    let nextTickTime = Number(process.hrtime.bigint()) / 1e6 + pacingIntervalMs;

    const playoutStep = () => {
      const now = Number(process.hrtime.bigint()) / 1e6;
      let timeToNextTick = nextTickTime - now;

      const currentTarget = this._isInitial ? this._initialBufferBytes : this._targetBufferBytes;
      if (!this._isPlayoutActive && this._audioBuffer.length < currentTarget && this._audioBuffer.length > 0) {
        if (!this._flushTimeout) {
          this._flushTimeout = setTimeout(() => {
            this._isPlayoutActive = true;
            this._playoutInterval = setTimeout(playoutStep, 0);
          }, 100);
        }
        nextTickTime = now + pacingIntervalMs;
        this._playoutInterval = setTimeout(playoutStep, pacingIntervalMs);
        return;
      }
      if (this._flushTimeout) {
        clearTimeout(this._flushTimeout);
        this._flushTimeout = null;
      }
      this._isPlayoutActive = true;
      if (this._isInitial) this._isInitial = false;

      if (this._audioBuffer.length >= this._chunkSize) {
        const chunk = this._audioBuffer.subarray(0, this._chunkSize);
        this._audioBuffer = this._audioBuffer.subarray(this._chunkSize);
        this._playedAudioMs += chunk.length / BYTES_PER_MS_G711;

        const mediaMsg = {
          event: 'media',
          streamSid: this.streamSid,
          media: { payload: chunk.toString('base64') }
        };
        this.originalWs.send(JSON.stringify(mediaMsg));

        while (this._eventQueue.length > 0) {
          const mark = this._eventQueue[0];
          mark.byteOffset -= this._chunkSize;
          if (mark.byteOffset <= 0) {
            this.originalWs.send(JSON.stringify(mark.data));
            this._eventQueue.shift();
          } else {
            break;
          }
        }

        nextTickTime += pacingIntervalMs;
        timeToNextTick = nextTickTime - Number(process.hrtime.bigint()) / 1e6;
        if (timeToNextTick < 0) timeToNextTick = 0;

        this._playoutInterval = setTimeout(playoutStep, timeToNextTick);
      } else if (this._audioBuffer.length > 0) {
        const padding = Buffer.alloc(this._chunkSize - this._audioBuffer.length, 0xff);
        this._audioBuffer = Buffer.concat([this._audioBuffer, padding]);
        this._playedAudioMs += padding.length / BYTES_PER_MS_G711;

        nextTickTime += pacingIntervalMs;
        timeToNextTick = nextTickTime - Number(process.hrtime.bigint()) / 1e6;
        if (timeToNextTick < 0) timeToNextTick = 0;
        this._playoutInterval = setTimeout(playoutStep, timeToNextTick);
      } else {
        this._isPlayoutActive = false;
        this._playoutInterval = null;

        while (this._eventQueue.length > 0) {
          this.originalWs.send(JSON.stringify(this._eventQueue.shift().data));
        }
      }
    };

    this._playoutInterval = setTimeout(playoutStep, pacingIntervalMs);
  }
}
```
EDIT: Fix typos

PR #1191, opened 2026-04-21
fix(core): replay aborted managed tool outputs

MukundaKatta:
## Summary
- queue synthetic function_call_result items when a streamed managed-conversation run aborts after OpenAI Responses function_call events have already been emitted
- replay those synthetic outputs on the next conversationId turn so the server-side conversation store stays balanced
- add a streaming regression test that aborts after a raw function_call event and verifies the next run sends the synthetic output

## Testing
- git diff --check
- npm test -- --run packages/agents-core/test/run.stream.test.ts (blocked locally because vitest is not installed in this clone)

### DIFF
--- a/.changeset/replay-aborted-tool-outputs.md
+++ b/.changeset/replay-aborted-tool-outputs.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Replay managed tool outputs that completed before an abort, so aborted runs don't lose already-finished tool results on the next turn.
--- a/packages/agents-core/src/run.ts
+++ b/packages/agents-core/src/run.ts
@@ -71,6 +71,10 @@ import {
 import { resolveTurnAfterModelResponse } from './runner/turnResolution';
 import { prepareTurn } from './runner/turnPreparation';
 import {
+  clearManagedConversationSupplementalItems,
+  queueManagedConversationSupplementalItems,
+} from './runner/turnPreparation';
+import {
   applyTurnResult,
   handleInterruptedOutcome,
   resumeInterruptedTurn,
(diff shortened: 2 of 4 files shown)


wsk-builds (commented):
Thanks for working on this. I checked out the current draft locally and ran:

- `pnpm i`
- `pnpm -F @openai/agents-core dist:check`
- `CI=1 pnpm test -- packages/agents-core/test/run.stream.test.ts`

All completed successfully; the Vitest command ended up running the full suite in this checkout and reported 111 test files / 1783 tests passing.

I agree this is an important edge case for `conversationId` + streaming aborts: if a Responses function call has already been persisted server-side, the next turn needs some way to avoid leaving the managed conversation in an unbalanced state.

I think the main risk is the current ownership/lifetime of the pending abort items. `pendingManagedConversationAbortItems` is a module-level `Map` keyed only by `conversationId`, so the cleanup state can outlive a `Runner`/request and can also be shared across unrelated callers in the same process. That makes a few cases hard to reason about: concurrent runs using the same conversation id, aborted runs that are never resumed, and long-lived server processes where pending entries may remain indefinitely. I think this state should either be tied to `RunState` / the runner/session lifecycle, or the PR should explicitly document why process-global conversation-id state is safe here.

I also think the synthesized output semantics need another look. The PR turns a raw `function_call` into a `function_call_result` with `status: "completed"` and output text `"aborted"`. A user abort is not really a completed tool execution, and this overlaps with the status/history concerns discussed in #1104 / #1110. If `completed` is required for server-managed history compatibility, it would be good to call that out and add a test/probe explaining why `incomplete` or another representation is not viable. Otherwise the model-visible history may incorrectly imply the tool succeeded.

For tests, the current regression test covers the basic replay path, but I think this needs a broader matrix before merge:

- multiple function calls emitted before abort
- abort before a usable `call_id` is available
- `previousResponseId` mode remains a no-op
- pending items are cleared after a successful replay
- pending items do not get lost or duplicated if the next run fails
- concurrent runs with the same `conversationId`
- interaction with existing managed conversation supplemental handoff items

One smaller thing: the changeset package looks right as `@openai/agents-core`, but the summary should probably use the repo's Conventional Commit style, e.g. `fix(core): replay aborted managed conversation tool outputs`.

Overall, I think the PR is pointed at a real bug, but I'd prefer to see the pending-item ownership and synthesized-result semantics clarified before merge.

PR #1243, opened 2026-05-06
fix: preserve duplicate-name agent identity in RunState serialization

seratch:
This pull request fixes RunState resume for agent graphs that contain distinct agents with the same name, including sandbox-agent interruption and snapshot recovery flows.

It updates the serialized RunState schema to store stable optional agent identities, preserves legacy name-only duplicate rejection for older payloads, and applies identity-aware serialization to current agents, run items, guardrail results, processed responses, interruptions, and tool-use tracking. It also updates the HITL documentation and adds a patch changeset for `@openai/agents-core` and the bundled `@openai/agents` package.

### DIFF
--- a/.changeset/runstate-agent-identity.md
+++ b/.changeset/runstate-agent-identity.md
@@ -1,0 +1,6 @@
+---
+'@openai/agents-core': patch
+'@openai/agents': patch
+---
+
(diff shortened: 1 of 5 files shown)


seratch:
@codex review

PR #1265, opened 2026-05-07
fix: restore session history when responses compaction replacement fails

seratch:
This pull request fixes a data-loss edge case in `OpenAIResponsesCompactionSession` when replacing stored history after `responses.compact`.

The previous flow cleared the underlying session before writing compacted output, so a failing `addItems()` call could leave the session empty or partially replaced. The new flow snapshots the previous session items, restores them when replacement fails, and preserves the original replacement error when restore itself fails.

This mirrors the safety fix from the Python SDK: https://github.com/openai/openai-agents-python/pull/3117

### DIFF
--- a/.changeset/quiet-lamps-restore.md
+++ b/.changeset/quiet-lamps-restore.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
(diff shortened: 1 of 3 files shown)


seratch:
@codex review

PR #1276, opened 2026-05-08
fix(sandbox): align Blaxel parity behavior

mstolarzblaxelai:
## Summary

- Align Blaxel sandbox behavior with the shared sandbox error and lifecycle contracts.
- Wrap Blaxel file read failures in typed workspace read errors.
- Throw `SandboxInvalidManifestPathError` for out-of-root workspace paths.
- Convert Blaxel command timeouts from milliseconds to seconds at the provider boundary.
- Preserve owned `pauseOnExit` identity so `delete()` can pause and `resume()` can reattach.

## Why these are grouped

These fixes cover separate parity bugs, but they share the same Blaxel sandbox provider boundary and regression test surface. Keeping them together makes the provider behavior easier to review end to end while keeping the resolver-preservation fix separate in #1275.

## Validation

- `pnpm exec vitest run packages/agents-core/test/sandboxWorkspacePaths.test.ts packages/agents-extensions/test/sandbox/blaxel.test.ts`
- `pnpm -F @openai/agents-extensions build-check`
- `pnpm changeset:validate-lite`
- `pnpm build`
- `pnpm -r build-check`
- `pnpm lint`
- `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1 TMPDIR=/private/tmp CI=1 pnpm test`
- `git diff --check origin/main...HEAD`


packages/agents-extensions/src/sandbox/blaxel/sandbox.ts
```diff
@@ -1319,6 +1362,42 @@ function unknownErrorMessage(error: unknown): string {
...
+  );
+}
+
+function toBlaxelTimeoutSeconds(
+  timeoutMs: number | undefined,
+): number | undefined {
+  if (timeoutMs === undefined) {
+    return undefined;
+  }
+  return Math.ceil(timeoutMs / 1000);
```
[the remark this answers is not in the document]
mstolarzblaxelai: Thanks — I ran a live Blaxel backend probe to check this before changing code.

Result: the backend currently treats `process.exec({ waitForCompletion: true, timeout })` as seconds, matching the generated `@blaxel/core` type.

Evidence from a temporary sandbox in the `calibrator` workspace:
- `sleep 0.1` with `timeout: 10` completed successfully in ~450ms.
- `sleep 2` with `timeout: 1` failed after ~1.2s with `process timed out after 1 seconds`.
- The probe sandbox was deleted afterward and `bl get sandboxes --workspace calibrator -o json` returned `[]`.

So keeping the conversion from OpenAI SDK `timeoutMs` to Blaxel seconds is intentional here. The public docs example that shows `timeout: 60000 // 60 seconds` appears stale/conflicting with the backend and generated SDK contract.

pnpm-workspace.yaml
```diff
@@ -16,5 +16,7 @@ onlyBuiltDependencies:
 
 publishBranch: main
 minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+  - '@blaxel/core'
```
seratch: If upgrading to 0.2.81 is fine this time, can you remove this exclusion?

seratch (changes requested):

pnpm-workspace.yaml
mstolarzblaxelai: The exclusion was only needed while updating the lockfile, so it does not need to stay in the repo.

I’ll remove it and keep the locked @blaxel/core@0.2.82 upgrade.

seratch:
Once Codex PR review says 👍, we will mereg this PR. This repo automatically runs a Codex review for each commit in a PR.

pnpm-workspace.yaml
seratch: Thanks for quickly resolving this!

seratch (approved):

Issue #927, opened 2026-01-31
No configuration option exists for rate-limiting LLM API calls


Devendraxp:
I read the complete documentation and there is no option available for limit per minute model call. As I use it with free providers like groq cloud or even with openai, my agent hit per minute limit and with a error 429, it stops working.

seratch:
Thanks for sharing the idea! We do not have immediate plans to add a built-in rate limiter or customization options like that. We generally recommend defining an upper limit on your app's agent runs within a given time window. This is usually easier to manage outside of this SDK. Either way, even if this SDK provided a feature to detect bursty model traffic, you would still need to monitor overall traffic trends on your side (e.g., storing the data in your database and share the data across your app servers).

Usage data already provides relevant signals (e.g., number of model calls and token usage) that you can use for your rate-limiting logic, so you can leverage those metrics to control the pace of processing. If this SDK should (and can) provide more granular information to support this, we are happy to enhance it for that purpose.

Devendraxp:
When we run **result = await run(agent, 'xyzxyz..');,** we lose all control over our agent. It hits the LLM with back-to-back API requests until it reaches its rate limit, at which point it stops working. Even after waiting 1–2 minutes, it will not restart automatically.

Somewhere in the documentation, I read that it should **wait incrementally 1s, 2s, 4s, 8s, etc**. but in my case, it doesn’t. I asked my teammates and my local WhatsApp community, but no one had any idea why.

Additionally, any custom limits applied externally seem to be useless.

If I’m missing something, please let me know,it would help many developers.

seratch:
Enabling to retry the same model call in an exponential-backoff fashion could be indeed useful for the use case while it may not save all the errors if your budget for the model provider is not sufficient. As a workaround, utilizing ai-sdk's auto-retry or signal handler may be helpful: https://github.com/openai/openai-agents-js/issues/894#issuecomment-3772783383 In the long run, this SDK may introduce a built-in retry mechanism, but I cannot tell when it may be avaialble.

Devendraxp:
Thanks for the context. However, the team should make the **_run()_** function more configurable. Once invoked, we lose all control, everything is handed over to the agent. If it hits a rate limit, it simply crashes instead of pausing and retrying after a minute.
This makes the Agent SDK impractical for even moderately complex tasks like browser automation, where multiple steps and API calls are inevitable. A built-in retry mechanism or rate-limiting configuration isn't a "nice-to-have", it's essential for production use.

[...]

PR #1326, opened 2026-05-15
docs: clarify usage budgeting for model calls

EfeDurmaz16:
Summary
- Clarify that model retries are not a rate limiter or spending control.
- Extend the usage example with a small request/token budget check using `result.state.usage`.
- Point readers from the retry docs to usage-based application-level budgeting.

Context
- Addresses #927 without adding a new SDK-level rate limiter API.

Verification
- `pnpm -F @openai/agents-core build`
- `pnpm -F @openai/agents-openai build`
- `pnpm -F @openai/agents-realtime build`
- `pnpm -F @openai/agents build`
- `pnpm -F @openai/agents-extensions build`
- `pnpm -F docs-code build-check`
- `pnpm docs:scripts:check`
- `pnpm docs:build`
- `pnpm exec prettier --check examples/docs/results/usage.ts docs/src/content/docs/guides/results.mdx docs/src/content/docs/guides/models.mdx`
- `git diff --check`

### DIFF
--- a/docs/src/content/docs/guides/models.mdx
+++ b/docs/src/content/docs/guides/models.mdx
@@ -200,6 +200,8 @@ The SDK exports ready-made helpers on `retryPolicies`:
 
 When you compose policies, `providerSuggested()` is the safest first building block because it preserves provider vetoes and replay-safety approvals when the provider can distinguish them.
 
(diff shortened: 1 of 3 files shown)


seratch:
While we don't need these changes, thanks for the suggestion!

PR #1357, opened 2026-05-21
fix: mark rejected tool call results as incomplete

### LINKED ISSUE #1104: Rejected tool calls use status: 'completed' in function_call_result, causing model hallucinations

LeSingh1:
Fixes #1104.

`buildApprovalRejectionResult` produced `function_call_result` items with `status: 'completed'`, identical to successful tool runs. The only signal that the tool was rejected was the text content of the output, which contradicts the `status` field. As reported in #1104, this can cause models to claim a rejected tool actually executed (e.g. approve one email, reject another, model says both were sent).

The protocol already allows `'in_progress' | 'completed' | 'incomplete'` for `FunctionCallResultItem.status`, so this changes rejected results to `status: 'incomplete'`. This gives the model a consistent structural signal alongside the rejection text.

`getToolCallOutputItem` gains an optional `status` parameter that defaults to `'completed'`, so other call sites are unchanged.

Tests added:
- `getToolCallOutputItem` accepts and uses an explicit `status`.
- Function tool approval rejections produce a `function_call_result` with `status: 'incomplete'`.

`pnpm build`, `pnpm lint`, `pnpm -r build-check`, and the full vitest suite (2826 passed) all pass locally.

### DIFF
--- a/.changeset/rejected-tool-call-incomplete-status.md
+++ b/.changeset/rejected-tool-call-incomplete-status.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: mark rejected tool call results as incomplete
+
+Approval rejections previously emitted `function_call_result` items with `status: 'completed'`, identical to successful tool runs. The structural status contradicted the rejection text and could cause models to claim a rejected tool actually executed. Rejected results now use `status: 'incomplete'`.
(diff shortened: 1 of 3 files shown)


seratch:
We won't fix this issue for the reason mentioned https://github.com/openai/openai-agents-js/issues/1104#issuecomment-4096498620

Issue #1353, opened 2026-05-21
Firefox WebRTC Realtime sessions disconnect mid-turn; likely ICE consent / USE-CANDIDATE interop


crabbixOCE:
### Describe the bug

Firefox WebRTC sessions against OpenAI Realtime can connect successfully, exchange one or more turns, then disconnect mid-session/mid-turn. Chrome and Edge do not reproduce the issue in the same app/environment. Switching the same Firefox flow to the Realtime WebSocket transport prevents the connection from closing.

This appears to match Mozilla's active WebRTC bug report for OpenAI Realtime:

- https://bugzilla.mozilla.org/show_bug.cgi?id=2037009

Mozilla's investigation in that thread points toward an ICE consent interop issue: Firefox consent checks do not include `USE-CANDIDATE`, and the OpenAI media server appears to stop responding to them. A local Firefox build that adds `USE-CANDIDATE` to consent checks reportedly makes the OpenAI Realtime repro work, but that is not something application JavaScript can control. If the root cause is in Realtime media infrastructure rather than this SDK, could this issue be used to track/escalate/document it from the SDK side?

### Environment

- Package: `@openai/agents-realtime`
- Version observed locally: `0.10.1`
- Browser affected: Firefox
- Browsers not affected in the same environment: Chrome, Edge
- Transport affected: `OpenAIRealtimeWebRTC`
- Workaround that works: Realtime WebSocket transport

### Observed behavior

The session starts and audio/transcript events flow normally. After one or two turns, the peer connection transitions to disconnected/closed while the conversation is still active. In one case the assistant transcript continued to finalize text after audio went silent, but the WebRTC connection had already closed.

Example app-side logs:

```text
[OpenaiRealtimeVoiceService] WebRTC peer connection state
{ reason: "iceconnectionstatechange", connectionState: "disconnected", iceConnectionState: "disconnected", iceGatheringState: "complete", signalingState: "stable" }

[OpenaiRealtimeVoiceService] WebRTC peer connection state
{ reason: "connectionstatechange", connectionState: "closed", iceConnectionState: "closed", iceGatheringState: "complete", signalingState: "closed" }
```

`about:webrtc` showed a selected/nominated candidate pair with bytes flowing before the failure. Mozilla's linked bug shows the more specific Firefox-side signature:

```text
STUN-CLIENT(consent): Timed out
```

### Things already tried / ruled out

- Wake Lock / permissions-policy errors were present but unrelated; the call can start and run.
[...]

PR #1363, opened 2026-05-22
fix: delay WebRTC close for transient peer disconnects

seratch:
This pull request addresses the SDK-side WebRTC disconnect handling discussed in https://github.com/openai/openai-agents-js/issues/1353.

It updates `OpenAIRealtimeWebRTC` so `RTCPeerConnection.connectionState === "disconnected"` no longer immediately tears down the transport. Instead, the SDK keeps the active connection open for a short grace period, clears the pending close if the peer connection returns to `connected`, and still closes immediately for `failed` or `closed`.

The change adds regression coverage for transient disconnect recovery and for the grace-period close path in `packages/agents-realtime/test/openaiRealtimeWebRtc.test.ts`. It also adds `.changeset/sweet-webrtc-disconnect.md` with a patch changeset for `@openai/agents-realtime`.

### DIFF
--- a/.changeset/sweet-webrtc-disconnect.md
+++ b/.changeset/sweet-webrtc-disconnect.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: delay WebRTC close for transient peer disconnects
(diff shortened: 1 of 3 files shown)


lhweiler03-sc (commented):
Boom

PR #1368, opened 2026-05-26
fix: set model metadata on generation span in streaming + chat completions

imxinde:
## Summary

`OpenAIChatCompletionsModel.getStreamedResponse()` creates a generation span but never populates `spanData.model` or `spanData.model_config`. Downstream tracing consumers (OpenInference, OpenTelemetry exporters) then report the model name as `"unknown"`, which breaks model-level metrics and filtering in observability backends.

The non-streaming `getResponse()` path already sets these fields correctly (via `withGenerationSpan` callback). This PR brings the streaming path to parity.

## Changes

- **`packages/agents-openai/src/openaiChatCompletionsModel.ts`** — Set `span.spanData.model` and `span.spanData.model_config` in the `getStreamedResponse()` method, matching `getResponse()`.
- **`packages/agents-openai/test/openaiChatCompletionsModel.scenario.test.ts`** — Add test verifying the generation span contains model metadata when streaming with tracing enabled.

## Context

The Python SDK does not have this issue — its `generation_span(model=...)` context manager accepts `model` as a creation-time parameter that applies uniformly to both streaming and non-streaming paths.

In the JS SDK, the streaming path uses manual span lifecycle (`createGenerationSpan()` + `span.start()` / `span.end()`) instead of `withGenerationSpan(callback)`, and the metadata assignments were not carried over when this pattern was introduced in the initial commit.

### DIFF
--- a/.changeset/fix-streaming-generation-span-model.md
+++ b/.changeset/fix-streaming-generation-span-model.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: populate model and model_config on generation span in streaming mode
+
+`getStreamedResponse()` in `OpenAIChatCompletionsModel` was not setting `span.spanData.model` or `span.spanData.model_config` on the generation span, causing downstream tracing exporters to report the model as "unknown". The non-streaming `getResponse()` path already set these fields correctly.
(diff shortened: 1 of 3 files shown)


seratch (approved):
Confirmed this works well at runtime

Issue #466, opened 2025-09-12
Realtime SDK + Azure: unsupported fields; initial config ignored


AlexRRR:
### Summary
When using the Agents Realtime SDK with Azure OpenAI Realtime (WebRTC), the SDK emits OpenAI-specific session fields that Azure rejects (e.g., `session.type`, `output_modalities`). Also, with WebRTC the session is created during SDP, so initial agent config (instructions/tools) is ignored until the data channel opens; subsequent SDK-driven `session.update` calls include fields Azure doesn’t accept, preventing config from applying cleanly.

### Notes

I recognize these incompatibilities stem primarily from Azure OpenAI’s Realtime implementation differing from OpenAI’s GA interface (for example, rejecting `session.type`, `output_modalities`, and disallowing `["audio"]`-only `modalities`). Still, I wanted to open this issue so the community has a clear record of the symptoms, error messages, and a practical workaround. Even if the “fix” ultimately belongs on the Azure side (which btw I have no idea where to report), a provider-aware path or extensibility hook in the SDK would help developers avoid common pitfalls and ship faster across providers.

### Environment
- @openai/agents-realtime: 0.1.0
- @openai/agents: 0.1.0
- openai: 5.19.1
- Next.js: 15.5.2
- React: 19.1.1
- Node.js: 20.x
- Client: Chrome (WebRTC, ephemeral client key)
- Provider: Azure OpenAI Realtime (region WebRTC endpoint)
- gpt-realtime

### Repro steps
1. Browser, WebRTC:
   ```ts
   import { RealtimeAgent, RealtimeSession } from "@openai/agents-realtime";

   const agent = new RealtimeAgent({ name: "Assistant", instructions: "..." });
   const session = new RealtimeSession(agent, {
     model: "gpt-realtime",
     // baseUrl points to Azure WebRTC endpoint
     // e.g. https://<region>.realtimeapi-preview.ai.azure.com/v1/realtimertc
   });

   await session.connect({
     apiKey: "<ephemeral>",
     url: "https://<region>.realtimeapi-preview.ai.azure.com/v1/realtimertc",
   });
   ```
2. Observe transport events (`session.created`, then `session.update`).

### Observed behavior
- `session.created` shows Azure defaults (expected for WebRTC).
- SDK sends `session.update` with OpenAI-only fields:
  - Azure error: Unknown parameter: `session.type`.
  - Azure error: Unknown parameter: `session.output_modalities`.
[...]

PR #1374, opened 2026-05-31
feat(realtime): add session payload transform hook

PeterTianbuhan:
## What changed

Adds a `transformSessionPayload` transport option for SDK-managed Realtime `session.update` payloads. The hook is applied to both regular session config updates and tracing updates, while `buildSessionPayload()` continues to return the OpenAI Realtime API shape for existing helper use cases.

This also exports the related payload transformer types, adds regression coverage for config and tracing updates, documents an Azure OpenAI Realtime adaptation example, and includes a patch changeset for `@openai/agents-realtime`.

## Why

Azure OpenAI Realtime-compatible endpoints can reject OpenAI GA-specific session fields such as `type` or `output_modalities`. A generic transform hook lets integrations adapt the outgoing payload shape without hardcoding Azure-specific behavior into the SDK transports.

Refs #466.

## Testing

- `env PATH=/Users/peterlee/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH PNPM_CONFIG_MANAGE_PACKAGE_MANAGER_VERSIONS=false bash .agents/skills/code-change-verification/scripts/run.sh`

### DIFF
--- a/.changeset/short-pianos-reflect.md
+++ b/.changeset/short-pianos-reflect.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+feat: add Realtime session payload transform hook (#466)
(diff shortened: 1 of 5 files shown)


PeterTianbuhan:
Closing this stale PR for now. It still represents a useful compatibility idea, but it has fallen behind main and would need a fresh rebase plus renewed validation before it is worth reviewing. Happy to revisit later with a narrower refreshed version if this hook is still useful.

PR #1380, opened 2026-06-05
feat(mcp): expose serializable full tool results

seratch:
This pull request adds a serializable `callToolResult()` API for local MCP servers so integrations such as Temporal can preserve `content`, `_meta`, `structuredContent`, and `isError` across JSON serialization boundaries.

The existing `callTool()` content-array contract remains unchanged. Built-in stdio, SSE, and Streamable HTTP transports expose the new full-result method. Local MCP servers can also opt in to using JSON-stringified `structuredContent` as model-visible output with `useStructuredContent: true`.


seratch:
@codex review

PR #1381, opened 2026-06-05
fix: update Modal sandbox provider for Modal SDK 0.7.6 APIs

thomelane:
The Modal sandbox provider was written against older Modal SDK APIs. This updates it to the APIs in `modal` 0.7.6 and bumps the dependency.

## Changes

- Use the client-scoped `modal.secrets` and `modal.cloudBucketMounts` APIs instead of the module-level `Secret` / `CloudBucketMount` exports, and drop the compatibility adapters for the old shapes.
- Use `sandbox.filesystem.readBytes` / `writeBytes` for sandbox file IO instead of `sandbox.open()` file handles.
- Pass `command: ['sleep', 'infinity']` as a sandbox create param instead of setting it on the image via `image.cmd()`.
- Reject tar workspace persistence when the manifest has mount entries under the workspace root, since the archived tree would capture/overwrite mounted content (`SandboxUnsupportedFeatureError`).
- Bump `modal` to 0.7.6 and update tests for the new API surface.

All 55 tests in `test/sandbox/modal.test.ts` pass.


alfozan (approved):

seratch (approved):
Thanks for updating the code properly. LGTM

PR #1414, opened 2026-06-19
fix(realtime): handle malformed realtime messages

fallintoplace:
## Summary

- Guard realtime event parsing so malformed or non-JSON frames are ignored instead of throwing from transport message listeners.
- Reuse the parse result returned from `_onMessage()` in WebSocket and WebRTC transport-specific sequencer paths, avoiding duplicate parsing.
- Add coverage for malformed JSON, binary non-JSON frames, generic unknown events, and response.create sequencer behavior.

## Testing

### DIFF
--- a/.changeset/fresh-mugs-parse.md
+++ b/.changeset/fresh-mugs-parse.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: handle malformed realtime messages without duplicate parsing
--- a/.changeset/fresh-mugs-parse.md
+++ b/.changeset/fresh-mugs-parse.md
@@ -2,4 +2,4 @@
 '@openai/agents-realtime': patch
 ---
 
-fix: handle malformed realtime messages without duplicate parsing
+fix: ignore malformed realtime messages without throwing an error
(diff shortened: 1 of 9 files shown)


seratch (changes requested):
Thanks for your interest here This PR changes the exported `OpenAIRealtimeBase` protected `_onMessage` declaration from returning `void` to returning a parse result. An external subclass with a `void`-returning override compiles against v0.11.8 and the base commit, but fails with TS2416 against this branch, so this cannot ship as a patch as written. Please preserve the released `_onMessage(...): void` override contract and hook behavior. The simplest fix is to keep the existing duplicate parsing and only make `parseRealtimeEvent` tolerant. Alternatively, use an internal helper that does not change or bypass the protected hook. Please also add a type-level regression covering a `void`-returning subclass override.

fallintoplace:
@seratch Thank you for your comment. I have implemented how you want. Could you please check again later?

.changeset/fresh-mugs-parse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: handle malformed realtime messages without duplicate parsing
```
seratch:
```suggestion
fix: ignore malformed realtime messages without throwing an error
```

seratch (approved):

PR #1427, opened 2026-06-24
fix: reject invalid handoff input filters

fallintoplace:
## Summary
- Validate the selected handoff input filter before invoking the handoff so non-callable filters fail as configuration errors.
- Preserve the existing span error metadata while avoiding partial handoff side effects.
- Add regression coverage that confirms invalid filters do not invoke handoff callbacks or emit handoff events.

## Tests
- pnpm vitest run packages/agents-core/test/runner/toolExecution.test.ts
- pnpm -F @openai/agents-core build-check
- bash .agents/skills/code-change-verification/scripts/run.sh

### DIFF
--- a/.changeset/invalid-handoff-input-filter.md
+++ b/.changeset/invalid-handoff-input-filter.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: reject invalid handoff input filters before invoking handoffs
--- a/packages/agents-core/src/handoff.ts
+++ b/packages/agents-core/src/handoff.ts
@@ -434,7 +434,7 @@ export function handoff<
     handoff.toolDescription = config.toolDescriptionOverride;
   }
 
-  if (config.inputFilter) {
+  if (config.inputFilter != null) {
     handoff.inputFilter = config.inputFilter;
   }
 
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Thanks for sharing this patch. The underlying issue is valid, and moving validation to the effective-filter selection point before onInvokeHandoff and handoff event emission is the right direction.

Before merge, please treat only null and undefined as absent. The current truthiness guard still accepts false or "" from JavaScript or any, then performs the handoff. Please add a falsy non-function regression case and assert that the callback and both runner and agent handoff events do not fire.

Also, please either remove the manual span setError or preserve and test its structured data; withHandoffSpan currently overwrites it when it catches the thrown UserError. Once these focused changes are covered, the PR should be ready for another review.

fallintoplace:
@seratch Please take another look when you have time. Thank you for your support.

seratch (changes requested):
Thanks for updating the PR. The executor-side changes now address the earlier review: falsy non-functions are rejected before onInvokeHandoff and handoff events, and the structured span error data is preserved.

One public configuration path still bypasses the guard. handoff() only copies config.inputFilter under a truthiness check, so handoff(agent, { inputFilter: false as any }) or "" is discarded before executeHandoffCalls sees it. The new tests avoid this path by assigning h.inputFilter directly. Please preserve any non-null configured value, or validate it in handoff(), and add a regression through the public handoff(..., { inputFilter: ... }) path proving that the callback and events do not fire. After that, this should be ready for another review.

fallintoplace:
@seratch Thank you for your review. I have amended the PR.

seratch (approved):

PR #1436, opened 2026-07-01
fix(core): stop resending resolved tool results across consecutive streamed approvals

### LINKED ISSUE #1435: Consecutive tool call approvals fail in a streamed run with `previousResponseId`
### Please read this first

- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug

When a streamed run is started with `previousResponseId` and then resumed more than one tool-approval interruption, the run fails with: `400 No tool call found for function call output with call_id <id>`

### Debug information

- Agents SDK version: v0.11.6
- Runtime environment: Node.js v24.12.0

### Repro steps

1. Start a streamed run with previousResponseId set.
2. Agent calls an approval-gated tool → interruption. Approve.
3. Agent does more work, then calls another approval-gated tool → second interruption. Approve.
4. SDK sends back second approval including the first tool's `function_call_output` again, but `previous_response_id` now points past its `function_call` -> 400 error

```js
import { Agent, run, tool } from '@openai/agents';
import { z } from 'zod';

const ping = tool({
  name: 'ping',
  description: 'Ping a host.',
  parameters: z.object({ host: z.string() }),
  needsApproval: true,
  execute: async ({ host }) => `pong ${host}`,
});

const agent = new Agent({
  name: 'repro',
  model: 'gpt-5.4',
  instructions:
    'Call ping for "example.com", then AFTER it returns call ping again for "google.com". One at a time.',
  tools: [ping],
});

// Seed a first response so the failing run starts with previousResponseId set.
const seed = await run(agent, 'Say hi.');

let stream = await run(agent, 'Do the two pings now.', {
  stream: true,
  previousResponseId: seed.lastResponseId,
});

for (;;) {
  await stream.completed;
  const pending = stream.state.getInterruptions();
  if (pending.length === 0) break;
  for (const item of pending) stream.state.approve(item);
  stream = await run(agent, stream.state, { stream: true });
  // 2nd iteration throws: 400 No tool call found for function call output ...
}
```

### Expected behavior

Second approval passes without failure.

anneheartrecord:
## Summary
- Fixes #1435: a streamed run started with `previousResponseId` fails with `400 No tool call found for function call output with call_id ...` once it is resumed through more than one tool-approval interruption.
- Root cause: on each resume `ServerConversationTracker.primeFromState` rebuilds its "already sent" set from the model responses, but it only marked server-emitted items as sent. A client-generated `function_call_output` from an earlier approval was never marked, so it was replayed on every subsequent turn. By then `previousResponseId` had advanced past that call, so re-submitting its output was rejected.
- `primeFromState` now marks a tool result as sent when its `function_call` belongs to an already-resolved (non-latest) response. Only the most recent response's calls stay pending, so a resume sends just the delta. The pending set is derived from the last response regardless of whether it carries a `responseId`, so `conversationId`-managed turns behave the same.

## Tests
- `pnpm exec vitest run packages/agents-core/test/run.stream.test.ts` (adds coverage for consecutive approvals with `previousResponseId`, the `conversationId` variant, and consecutive rejections)
- `pnpm -F @openai/agents-core exec eslint src/runner/conversation.ts`

### DIFF
--- a/.changeset/tidy-approvals-resume.md
+++ b/.changeset/tidy-approvals-resume.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-core': patch
+---
+
+fix(core): stop resending resolved tool results across consecutive streamed approvals
+
+When a streamed run started with `previousResponseId` was resumed through more than one tool-approval interruption, the server conversation tracker re-primed each earlier approval's `function_call_output` on every resume. Because `previousResponseId` had already advanced past those calls, the follow-up request failed with `400 No tool call found for function call output with call_id ...`. `primeFromState` now marks tool results whose function call belongs to an already resolved response as sent, so only the latest turn's pending results are replayed.
--- a/packages/agents-core/src/runner/conversation.ts
+++ b/packages/agents-core/src/runner/conversation.ts
@@ -231,16 +231,33 @@ export class ServerConversationTracker {
     const hasResponses = modelResponses.length > 0;
 
     const serverItemKeys = new Set<string>();
+    // Every function call the server has emitted so far, and the subset that
+    // belongs to the most recent (still-pending) response. Results for calls
+    // from earlier responses were already delivered on a prior resume and must
+    // not be resent. We derive the pending set from the last response rather
+    // than gating on `responseId`, so a `conversationId`-managed turn whose
+    // response omits an id still keeps its own calls marked pending.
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for your interest here. The issue was resolved by https://github.com/openai/openai-agents-js/pull/1437 in the way we'd like to pursue.

PR #1453, opened 2026-07-09
fix(openai): support openai 6.46.0 usage types

HAYDEN-OAI:
## Summary

- Type the Chat Completions usage adapter as Agents SDK `UsageInput` instead of the generated Responses API `ResponseUsage`.
- Bump the workspace's direct `openai` dependencies to `^6.46.0` across the published packages and examples, and regenerate package metadata.
- Add a version-specific `minimumReleaseAgeExclude` for `openai@6.46.0` so the just-published launch SDK can pass the repository's normal frozen install policy.
- Add patch changesets for `@openai/agents`, `@openai/agents-core`, and `@openai/agents-openai`.

## Why

[openai/openai-node#1959](https://github.com/openai/openai-node/pull/1959) shipped `openai@6.46.0` with the required `cache_write_tokens` field on `ResponseUsage.InputTokensDetails`. The downstream Agents SDK regression check failed with TS2741 because `toResponseUsage()` synthesizes an Agents usage object containing only `cached_tokens`, while its return annotation incorrectly required the full Responses API output schema.

The value is immediately passed to `new Usage(...)`, whose supported input boundary is `UsageInput`. Using that type removes the accidental coupling to generated API output fields and lets required Responses outputs remain required.

The dependency floor and workspace examples now all use 6.46 so TypeScript does not load nominally incompatible `OpenAI` classes from 6.44 and 6.46 in the same build. The cooldown exception is scoped to exactly `openai@6.46.0`, not the package generally.

## Impact

There is no runtime behavior change in the usage adapter. Published Agents packages now require `openai@^6.46.0`, and the workspace explicitly validates against the GPT-5.6 launch SDK.

## Validation

- `pnpm changeset:validate-prompt` (patch bumps validated)
- `pnpm install --frozen-lockfile` (normal supply-chain policy passed)
- Reproduced the original downstream check with the built SDK from openai/openai-node#1959 linked into a scratch checkout:
  - `pnpm build`
  - `pnpm lint`
  - `pnpm docs:scripts:check`

### DIFF
--- a/.changeset/bright-caches-write.md
+++ b/.changeset/bright-caches-write.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: decouple Chat Completions usage from Responses types
--- a/.changeset/bright-caches-write.md
+++ b/.changeset/bright-caches-write.md
@@ -1,5 +1,7 @@
 ---
+'@openai/agents': patch
+'@openai/agents-core': patch
 '@openai/agents-openai': patch
 ---
 
-fix: decouple Chat Completions usage from Responses types
+fix: support openai-node v6.46.0 usage types
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -3,7 +3,7 @@
   "name": "basic",
   "dependencies": {
     "@openai/agents": "workspace:*",
-    "openai": "^6.42.0",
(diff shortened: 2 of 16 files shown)


apcha-oai (approved):

h8bnstalked88-boop (commented):
Install

PR #1456, opened 2026-07-10
feat: add configuration validation utilities

maslinedwin:
Provides config validation, model name validation, and default config creation for TypeScript agents.


seratch:
Thanks, but we don't need this change for now.

PR #1457, opened 2026-07-10
feat: add state management helpers

maslinedwin:
Adds AgentState interface and utilities for managing agent lifecycle and execution state.


seratch:
Thanks, but we don't need this change for now.

PR #1471, opened 2026-07-13
fix(openai): keep refusal and audio when chat content is an empty string

winklemad:
### What & why

The non-streaming Chat Completions converter (`openaiChatCompletionsModel.ts`) computes `hasContent` and only treats an empty-string `content` as "no content" for the `tool_calls` case:

```ts
const hasContent =
  message.content !== undefined &&
  message.content !== null &&
  !(message.tool_calls && message.content === '');
```

So when a provider (e.g. Azure, which sends `''` instead of `null` — the quirk this line already handles for tool calls) returns an assistant message with `content: ''` **and** a `refusal` (or `audio`), `hasContent` stays `true`. The converter emits an empty `output_text` item and silently demotes the refusal/audio into `providerData`, so the item `type` is `output_text` instead of `refusal`/`audio` and the refusal signal is lost downstream.

The streaming converter (`convertChatCompletionsStreamToResponses`) only builds text content when the delta content is truthy, so it produces the correct item for the same logical response — the two converter paths disagree on identical input:

`{ choices: [{ message: { content: '', refusal: 'no' } }] }`
- non-streaming → `content: [{ type: 'output_text', text: '', providerData: { refusal: 'no' } }]` ❌
- streaming → `content: [{ type: 'refusal', refusal: 'no' }]` ✅

### Fix

Extend the empty-content guard so an empty string doesn't preempt a `refusal` or `audio` payload either (not just `tool_calls`), matching the streaming converter.

### Tests

Added two cases to `openaiChatCompletionsModel.test.ts` — empty `content` with a refusal, and with audio — asserting the correct `refusal`/`audio` items (both fail before the change). Existing behaviour is unchanged (content `''` alone still yields an empty message; refusal/audio with absent content unaffected); the full `@openai/agents-openai` suite passes, and tsc/eslint/prettier are clean.

Found by reading the code; no linked issue.

### DIFF
--- a/.changeset/keep-refusal-audio-empty-content.md
+++ b/.changeset/keep-refusal-audio-empty-content.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Keep refusal and audio outputs when a Chat Completions message has an empty-string content, matching the streaming converter instead of emitting an empty text item.
--- a/packages/agents-openai/src/openaiChatCompletionsModel.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsModel.ts
@@ -138,8 +138,12 @@ export class OpenAIChatCompletionsModel implements Model {
       const hasContent =
         message.content !== undefined &&
         message.content !== null &&
-        // Azure OpenAI returns empty string instead of null for tool calls, causing parser rejection
-        !(message.tool_calls && message.content === '');
+        // Some providers (e.g. Azure) send an empty string instead of null, so an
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for investigating this. This PR covers the same synthetic refusal case as #1495 and additionally extends it to audio, but there is no linked user report or real provider response demonstrating either payload shape.

Because empty text and missing text have distinct semantics, we should not change the shared converter based only on constructed fixtures. I am going to close this PR as well. We can revisit the behavior if an actual provider response is supplied with the provider, model, API version, and redacted raw payload.

PR #1474, opened 2026-07-14
fix: preserve stream usage reported on non-final chunks

sohumt123:
### Summary

When converting a streamed Chat Completions response, token usage is silently lost whenever a provider reports usage on a non-final chunk.

In `convertChatCompletionsStreamToResponses` (`packages/agents-openai/src/openaiChatCompletionsStreaming.ts`), the streaming loop ran on every chunk:

```ts
// This is always set by the OpenAI API, but not by others e.g. LiteLLM
usage = (chunk as any).usage || undefined;
```

Because this overwrites `usage` unconditionally, the **last-seen** chunk is treated as authoritative. That is fine for the standard OpenAI path, where usage arrives on the final chunk. But some OpenAI-compatible providers or gateways may emit a later chunk without usage after reporting usage on an **earlier** chunk. On that final iteration `usage` is reset to `undefined`, so the terminal `response_done` event serializes usage via `usage?.prompt_tokens ?? 0` (and friends) and reports `inputTokens`, `outputTokens` and `totalTokens` all as `0` — silently corrupting cost/usage accounting.

The fix only overwrites `usage` when the current chunk actually carries usage data, so a value reported anywhere in the stream is retained:

```ts
if ((chunk as any).usage) {
  usage = (chunk as any).usage;
}
```

The normal OpenAI path is unchanged: when usage is on the final chunk, that chunk's usage still wins.

This is the direct JS counterpart of the merged Python fix openai/openai-agents-python#2126 ("Fix: usage from earlier stream chunks when later chunks have none"), which was never ported to the JS SDK.

### Test plan

Added and updated tests in `packages/agents-openai/test/openaiChatCompletionsStreaming.test.ts`:

- **New regression test** — `preserves usage reported on an earlier chunk when the final chunk has no usage`: streams a first chunk carrying `{prompt_tokens: 100, completion_tokens: 5, total_tokens: 105}`, followed by a terminal chunk with `finish_reason: 'stop'` and no usage. It asserts the final `response_done` usage is `{inputTokens: 100, outputTokens: 5, totalTokens: 105}` and that the raw `ChatCompletion.usage` matches. Before the fix this failed with all-zero usage; after the fix it passes.
- **Updated existing assertion** — in `converts chunks to protocol events`, the stream places usage `{1, 2, 3}` on a middle chunk with a trailing tool-call chunk that has no usage. The old assertion `expect(final.response.usage.totalTokens).toBe(0)` had codified the bug (mid-stream usage was discarded); it now asserts `.toBe(3)`, matching the corrected behavior. Same rationale as openai/openai-agents-python#2126.

The normal-path tests that report usage on the **last** chunk (`emits protocol events for streamed chat completions`, asserting `totalTokens: 7`, and `filters multiple choices by default`, asserting `totalTokens: 3`) are intentionally left unchanged and still pass as regression guards.

Verification (from repo root):

- `pnpm build && pnpm -r build-check` — pass
- `CI=1 pnpm test` — 3126 passed, 2 skipped
- `pnpm lint` — clean
- `.agents/skills/code-change-verification/scripts/run.sh` — all commands passed
- Added a changeset (`.changeset/fix-stream-usage-earlier-chunk.md`, patch bump for `@openai/agents-openai`)

### Issue number

No existing issue; the reproduction and root cause are described above.

### Checks

### DIFF
--- a/.changeset/fix-stream-usage-earlier-chunk.md
+++ b/.changeset/fix-stream-usage-earlier-chunk.md
@@ -4,4 +4,4 @@
 
 fix: preserve token usage reported on earlier Chat Completions stream chunks
 
-When streaming Chat Completions, usage was overwritten on every chunk, so a trailing chunk without usage reset the previously reported totals to zero. Some providers (e.g. xAI/Grok via LiteLLM) attach usage to an earlier chunk rather than the final one, which caused `response_done` to report `inputTokens`/`outputTokens`/`totalTokens` as 0. Usage is now retained when a later chunk omits it, while the normal OpenAI path (usage on the final chunk) is unchanged.
+When streaming Chat Completions, usage was overwritten on every chunk, so a trailing chunk without usage reset the previously reported totals to zero. Some OpenAI-compatible providers or gateways may emit a later chunk without usage after reporting usage on an earlier chunk, which caused `response_done` to report `inputTokens`/`outputTokens`/`totalTokens` as 0. Usage is now retained when a later chunk omits it, while the normal OpenAI path (usage on the final chunk) is unchanged.
--- a/.changeset/fix-stream-usage-earlier-chunk.md
+++ b/.changeset/fix-stream-usage-earlier-chunk.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: preserve token usage reported on earlier Chat Completions stream chunks
+
+When streaming Chat Completions, usage was overwritten on every chunk, so a trailing chunk without usage reset the previously reported totals to zero. Some providers (e.g. xAI/Grok via LiteLLM) attach usage to an earlier chunk rather than the final one, which caused `response_done` to report `inputTokens`/`outputTokens`/`totalTokens` as 0. Usage is now retained when a later chunk omits it, while the normal OpenAI path (usage on the final chunk) is unchanged.
--- a/packages/agents-openai/src/openaiChatCompletionsStreaming.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsStreaming.ts
@@ -59,8 +59,13 @@ export async function* convertChatCompletionsStreamToResponses(
       },
     };
 
-    // This is always set by the OpenAI API, but not by others e.g. LiteLLM
-    usage = (chunk as any).usage || undefined;
+    // This is always set by the OpenAI API, but not by others e.g. LiteLLM.
+    // Usage is not always reported on the final chunk (some providers such as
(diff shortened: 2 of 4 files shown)


seratch:
This PR does not yet establish that a supported JavaScript path produces this chunk ordering. The linked Python fix exercises `LitellmModel`, while the OpenAI specification and current LiteLLM documentation place non-null usage on the final data chunk.

Please provide a minimal JavaScript reproduction using `OpenAIChatCompletionsModel` or `OpenAIProvider` against the affected provider or gateway. It should include the provider and version, the raw chunk sequence showing non-null usage followed by a usage-less chunk, and the resulting zero usage in `response_done`.

If that exact path is reproducible, retaining the last non-null usage here looks like a reasonable narrow fix. Until then, I do not think we should merge compatibility behavior based only on the adjacent Python integration and a constructed stream.

seratch (changes requested):
Thanks for adding the regression test. I ran this through the actual JS converter and confirmed the failure with LiteLLM 1.80.13: a few adapters can emit a usage-bearing chunk followed by a chunk without usage, causing the current implementation to report zero usage. The proposed last-non-null behavior fixes that and does not affect the normal OpenAI streaming path.

One wording request: current LiteLLM 1.92.0 no longer reproduced this behavior, so describing it as current xAI/Grok behavior is too broad. Could you update the source comment, changeset, and PR description to something like:

> Some OpenAI-compatible providers or gateways may emit a later chunk without usage after reporting usage.

The implementation and test otherwise look good to merge.

sohumt123:
Thanks for verifying it, @seratch ! Updated the wording in all three places, source comment, changeset, and PR description,  to "some OpenAI-compatible providers or gateways may emit a later chunk without usage after reporting usage," rather than singling out xAI/Grok. Should be good to merge now.

seratch (changes requested):
Thanks, the requested wording has been updated in the implementation comment, changeset, and PR description. The same provider-specific wording still remains in two test comments, though: openaiChatCompletionsModel.scenario.test.ts still names xAI/Grok and LiteLLM, and openaiChatCompletionsStreaming.test.ts still says that xAI/Grok via LiteLLM currently behaves this way.

Please generalize those comments as well, since our latest LiteLLM verification did not reproduce that behavior. The implementation, regression coverage, and CI otherwise look ready to merge.

sohumt123:
Done -- generalized the wording in both test comments to match the implementation comment ("some OpenAI-compatible providers or gateways may emit a later chunk without usage after reporting usage"), and grepped the full diff to confirm nothing else remains. Thanks @seratch!

seratch (approved):

PR #1485, opened 2026-07-16
fix: correlate streamed text with output items

### LINKED ISSUE #1484: No way to correlate output_text_delta events with the message_output_created item they belong to
We're building a durable transcript (persistence + replay, not just rendering) on top of the streamed events and hit the following: `output_text_delta` only carries `{ type, delta }`,

```ts
// packages/agents-core/src/types/protocol.ts
export const StreamEventTextStream = SharedBase.extend({
  type: z.literal('output_text_delta'),
  delta: z.string(),
});
```

so when the `message_output_created` run item arrives later (it can land after `tool_called` / `agent_tool_start` for the same turn), there's no way to tell whether it's the completed form of the text we already streamed or a genuinely new message. We currently reconcile by matching on text content, which works but is fragile.

The ai-sdk-ui bridge in agents-extensions seems to have the same problem: `uiMessageStream.ts` generates its own synthetic text id per block and skips `message_output_created` entirely whenever any text streamed during that response (the `responseHasText` flag). That's fine for rendering, but any consumer that actually needs the completed item has to either drop it or content-match it.

What's frustrating is that the id exists at every layer except the normalized event:

- the Responses adapter spreads the raw event into `providerData`, so `providerData.item_id` is technically there, but that's provider-specific and undocumented
- the `aisdk()` wrapper drops it completely: AI SDK `text-delta` parts carry an `id` (plus `text-start`/`text-end`), but the wrapper yields `{ type: 'output_text_delta', delta }` with no `providerData` at all
- the Python SDK passes raw events through, so `ResponseTextDeltaEvent.item_id` is just available there

Would you take a change that adds an optional id to the delta events and guarantees the completed item carries the same one? Something like:

```ts
export const StreamEventTextStream = SharedBase.extend({
  type: z.literal('output_text_delta'),
  delta: z.string(),
  itemId: z.string().optional(),
});
```

with the Responses adapter filling it from `item_id`, the `aisdk()` wrapper forwarding the AI SDK part id (and setting the same id on the message item it builds for the final output), and ideally the same for reasoning deltas. Optional field, so existing consumers are unaffected. It would also let the ui bridge replace blocks in place instead of the `responseHasText` suppression.

For precedent: #100 and #412 asked for the same correlation ids for tool calls and were fixed, and openai/codex#5546 fixed this exact class of bug in codex (deltas and completions using different generated ids, so clients couldn't correlate them).

Seen on 0.13.3/0.13.4 with both the OpenAI Responses model and `aisdk()`-wrapped models. Happy to put up a PR if this direction sounds reasonable.

akim136:
This pull request resolves #1484.

Streamed `output_text_delta` events currently drop the stable output-item identifier exposed by both OpenAI Responses and AI SDK providers. Consumers that persist streamed transcripts therefore cannot reliably correlate deltas with the completed assistant message and must fall back to matching text content.

This change:

- adds an optional `itemId` to normalized text delta events;
- maps Responses `item_id` into that field while preserving it in `providerData`;
- forwards AI SDK text-part IDs and reuses a consistent ID on the completed message item; and
- adds regression coverage for both provider paths.

The field is optional, so providers that do not expose an identifier retain their existing event shape. If an AI SDK stream reports inconsistent text IDs, the completed combined message remains un-IDed rather than claiming an incorrect correlation.

Validation:

- `pnpm vitest run packages/agents-openai/test/openaiResponsesModel.test.ts packages/agents-extensions/test/ai-sdk/index.test.ts`

### DIFF
--- a/.changeset/curvy-deltas-correlate.md
+++ b/.changeset/curvy-deltas-correlate.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-core': patch
+'@openai/agents-openai': patch
+'@openai/agents-extensions': patch
+---
+
+fix: expose output item IDs on streamed text deltas (#1484)
--- a/packages/agents-core/src/types/protocol.ts
+++ b/packages/agents-core/src/types/protocol.ts
@@ -910,6 +910,10 @@ export type UsageData = z.infer<typeof UsageData>;
 export const StreamEventTextStream = SharedBase.extend({
   type: z.literal('output_text_delta'),
   /**
+   * The ID of the output item this delta belongs to, when provided by the model.
+   */
+  itemId: z.string().optional(),
+  /**
    * The delta text that was streamed by the modelto the user.
    */
   delta: z.string(),
--- a/packages/agents-extensions/src/ai-sdk/index.ts
+++ b/packages/agents-extensions/src/ai-sdk/index.ts
@@ -2010,6 +2010,7 @@ export class AiSdkModel implements Model {
         SerializedTool | SerializedHandoff
       >();
       let textOutput: protocol.OutputText | undefined;
+      let textItemId: string | undefined;
 
       // State for tracking reasoning blocks (for Anthropic extended thinking):
       // Track reasoning deltas so we can preserve Anthropic signatures even when text is redacted.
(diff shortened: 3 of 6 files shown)


akim136:
@seratch could you provide directional feedback on this before I mark ready for review?

For #1484, this draft adds an optional itemId to normalized output_text_delta events and propagates the provider’s stable item ID through the Responses and AI SDK adapters. Is that public API shape aligned with the SDK’s intended direction, or would you prefer correlation to remain in providerData or use another mechanism? Happy to adjust before requesting formal review.

kmalkan:
Thanks for picking this up — could reasoning deltas get the same treatment while you're in here (the aisdk wrapper already tracks reasoning blocks by `part.id`), so transcript consumers can drop content-matching for reasoning too?

seratch:
The issue was resolved by https://github.com/openai/openai-agents-js/pull/1486

PR #1492, opened 2026-07-17
fix(extensions): preserve AI SDK v3 image tool outputs

seratch:
This pull request fixes AI SDK v3 image tool outputs that were encoded using the v2 media format, causing providers to treat image URLs as text instead of image content. It emits `image-url` and `image-data` parts for v3 models while preserving v2 behavior, adds regression coverage for extensionless image URLs, and updates the example to use `gpt-5.6-luna` through the OpenAI Responses provider.


seratch:
@codex review

PR #1495, opened 2026-07-18
fix: preserve refusals with empty-string content

saneGuy:
## Summary

The non-streaming Chat Completions converter can drop a model refusal when a provider returns `message.content === ""` alongside `message.refusal`.

In that case `hasContent` is true, so the converter emits a blank assistant `output_text` item and never reaches the refusal branch. The streaming converter surfaces the refusal for the same payload, so streaming and non-streaming disagree.

This is most likely to show up with Chat Completions-compatible providers or gateways that normalize refusal content to an empty string instead of `null`.

## Fix

Treat `refusal + empty-string content` as no content, so the existing refusal branch runs.

This mirrors the nearby Azure guard for `tool_calls + ""`, where an empty content string should not prevent the more specific branch from handling the message.

## Tests

Added a regression test for non-streaming Chat Completions refusals with `content: ""`.

- `@openai/agents-openai` test suite passes
- eslint passes

### DIFF
--- a/.changeset/chat-refusal-empty-content.md
+++ b/.changeset/chat-refusal-empty-content.md
@@ -4,8 +4,4 @@
 
 fix: preserve chat completion refusal when content is an empty string
 
-Some providers (e.g. Azure OpenAI) return an empty-string `content` alongside a
-refusal instead of `null`. The non-streaming Chat Completions converter treated
-that empty string as real content and emitted a blank assistant `output_text`
-message, silently dropping the refusal. It now falls through to the refusal
-branch, matching the streaming converter which already surfaces the refusal.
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the contribution. I traced this through the non-streaming and streaming Chat Completions paths, but the PR only demonstrates the behavior with a manually constructed response. We do not have an actual OpenAI, Azure, or compatible-provider response containing both `content: ""` and a non-empty `refusal`.

The current implementation intentionally distinguishes present empty text from absent text, and already surfaces refusals when content is null or absent. Changing that precedence based only on a synthetic payload is not justified. This also overlaps with #1471, which has the same evidence gap.

I am going to close this PR for now. If this occurs in practice, please provide the provider, model, API version, and a redacted raw response payload showing the exact combination. We can then revisit the narrowest appropriate fix.

saneGuy:
thanks @seratch, fair bar and a clean trace. you're right this came from a synthetic payload, not a captured provider response. i'll leave it closed and come back with provider/model/api-version + a redacted raw payload if i hit the real content:'' + refusal combination in practice. appreciate you tracing both paths.

PR #1498, opened 2026-07-20
fix(mcp): clear wrapper tool cache on Stdio/SSE connect/close/invalidate

hobostay:
## Problem

When `cacheToolsList: true` is enabled, the public `MCPServerStdio` and `MCPServerSSE` wrapper classes cache the tool list in their own `_cachedTools` field (separate from the underlying shim's cache). Unlike `MCPServerStreamableHttp`, neither wrapper ever cleared that local cache, so they **served a stale tool list indefinitely** after any of:

- `invalidateToolsCache()` — the documented way to refresh tools
- `close()` followed by `listTools()`
- a reconnect via `connect()`

Only `MCPServerStreamableHttp` had a `clearLocalToolsCache()` helper wired into these methods. Stdio/SSE were left out.

## Fix

Mirror the existing `MCPServerStreamableHttp` behavior:

- Add a private `clearLocalToolsCache()` helper to both `MCPServerStdio` and `MCPServerSSE` that resets `_cachedTools`.
- Call it in `connect()`, `close()`, and `invalidateToolsCache()`.
- Make those three methods `async` so the local cache is cleared and the forwarded call to the underlying shim both settle before returning (matching `MCPServerStreamableHttp`).

The `invalidateToolsCache()` path still forwards to the underlying shim's cache invalidation as before — it just also clears the wrapper-local cache now.

## Verification

Added regression tests in `packages/agents-core/test/mcpCache.test.ts` covering both `MCPServerStdio` and `MCPServerSSE`:

- returns fresh tools after `invalidateToolsCache()`
- does not serve stale tools after `close()`
- clears the wrapper cache on `connect()` (reconnect scenario)
- (Stdio) forwards `invalidateToolsCache()` to the underlying shim

Confirmed the tests actually catch the bug: with only the source change reverted, **6 of the new regression tests fail** (`expected [ 'a' ] to deeply equal [ 'b' ]`); with the fix applied, all 27 tests in the file pass.

```
✓ |@openai/agents-core| test/mcpCache.test.ts (27 tests)
```

Also ran the broader MCP test set (71 tests) plus `pnpm -F @openai/agents-core build-check`, ESLint, and Prettier — all clean.

## Changeset

Added `.changeset/mcp-stdio-sse-cache-invalidation.md` (`@openai/agents-core`: patch).

### DIFF
--- a/.changeset/mcp-stdio-sse-cache-invalidation.md
+++ b/.changeset/mcp-stdio-sse-cache-invalidation.md
@@ -1,0 +1,7 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: clear the local tool cache on MCPServerStdio/MCPServerSSE connect(), close(), and invalidateToolsCache()
+
+The public `MCPServerStdio` and `MCPServerSSE` wrappers keep their own `_cachedTools` (separate from the underlying shim's cache). Previously only `MCPServerStreamableHttp` cleared that wrapper cache, so the Stdio/SSE wrappers served stale tools indefinitely after `invalidateToolsCache()`, `close()`, or a reconnect via `connect()`. This mirrors the existing `clearLocalToolsCache()` behavior from `MCPServerStreamableHttp`.
--- a/packages/agents-core/src/mcp.ts
+++ b/packages/agents-core/src/mcp.ts
@@ -223,10 +223,15 @@ export class MCPServerStdio
   get name(): string {
     return this.underlying.name;
   }
-  connect(): Promise<void> {
+  private clearLocalToolsCache(): void {
+    this._cachedTools = undefined;
+  }
+  async connect(): Promise<void> {
+    this.clearLocalToolsCache();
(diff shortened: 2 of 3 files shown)


seratch (changes requested):
The reconnect and close paths remain incomplete: getAllMcpTools() can return its module-level cached tool list before reaching the wrapper, so an Agent may still receive tools from the previous connection.

Please route connect(), close(), and invalidateToolsCache() through one full-cache invalidation path covering the wrapper, shim, and shared server cache. Please also update the tests to populate through getAllMcpTools(), verify that close cannot return cached tools, and verify that reconnect returns the new connection's tools. The current post-close test should not expect listTools() to succeed on a closed server. Once those regressions pass against current main, this should be ready for another review.

seratch:
https://github.com/openai/openai-agents-js/pull/1605 includes your contribution in its commit. Thanks again for working on this!

PR #1510, opened 2026-07-22
feat(core): subscribe to MCP server logging events

### LINKED ISSUE #793: How to enable real-time streaming logs for MCP Tools during agent runs?
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-js/) YES
- **Have you searched for related issues?** Others may have had similar requests YES

### Describe the feature

Description
My MCP tool calls often take a long time (around 2–4 minutes). This is why I raised openai/openai-agents-js#772 about wanting configurable timeouts.

What I’d like is to stream real-time logs while the agent is analyzing, so that I can see progress updates instead of waiting silently. For example, in mcpserver I can use sendLoggingMessage to print logs (modelcontextprotocol/typescript-sdk#311).

Currently, my solution is to manually implement a CustomMCPServerSSE class and set up an extra channel during agent.run to receive logs. This works, but it feels hacky.

Question:  
Is there a more elegant way to implement interactive log streaming for MCP Tools, so that during an agent run I can receive real-time responses/logs without building a separate custom SSE pipeline?

hdimer:
## What

Adds an opt-in `serverLogging` option to the local MCP servers (`MCPServerStdio`, `MCPServerSSE`, `MCPServerStreamableHttp`) so an Agent can subscribe to `notifications/message` logging events emitted by an MCP server.

```ts
const mcpServer = new MCPServerStdio({
  name: 'Logging Demo Server',
  fullCommand: 'pnpm tsx logging-server.ts',
  serverLogging: {
    level: 'info', // optional minimum severity requested from the server
    handler: ({ level, logger, data, meta }) => {
      console.log(`[mcp:${level}] ${logger ?? 'server'}`, data);
    },
  },
});
```

## Why

Fixes #793. Long-running MCP tools (the issue reports 2–4 minute calls) can report progress via `sendLoggingMessage`, but today the SDK offers no way to observe those logs without building a custom SSE/transport pipeline by hand. This surfaces them through a documented `serverLogging.handler`, following the API shape @seratch proposed in the issue.

## How

- New **structural** public types in `mcp.ts` (`MCPServerLoggingLevel` / `MCPServerLoggingMessage` / `MCPServerLoggingHandler` / `MCPServerLoggingOptions`), declared locally so consuming `@openai/agents-core` never forces resolution of the optional `@modelcontextprotocol/sdk` package.
- The `notifications/message` handler is registered **before** `connect()` on every transport, so logs emitted during the initialize handshake are not dropped.
- An optional `level` is sent via `logging/setLevel` **after** connect, but only when the server advertises the `logging` capability. This keeps a reused streamable-HTTP session (an explicit `sessionId`, which skips initialize/capability discovery) from being torn down by a `setLevel` that would otherwise reject. The request is best-effort: a rejecting server logs a warning and stays connected.

## Tests

`packages/agents-core/test/shims/mcp-server/node.test.ts`, across stdio / SSE / streamable-HTTP:

- handler registered before connect, `setLevel` sent after (ordering asserted)
- notification `_meta` mapped to `meta`
- handler-only (no `level`) does **not** send `setLevel`
- explicit `sessionId` reuse skips `setLevel` while notifications still flow
- a rejecting `setLevel` still yields a connected, usable server

`pnpm build && pnpm -r build-check && pnpm test && pnpm lint` all pass. Includes a changeset and a runnable demo (`pnpm -F mcp start:logging`).

## Notes

- Scoped to local MCP servers in `agents-core`; no changes to hosted MCP.
- Happy to adjust the `serverLogging` shape (naming, sync vs async handler, per-transport nuances) to match your preference.

### DIFF
--- a/.changeset/mcp-server-logging.md
+++ b/.changeset/mcp-server-logging.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+feat(core): subscribe to MCP server logging notifications via `serverLogging`
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -316,7 +316,7 @@ const mcpServer = new MCPServerStdio({
 
 The `handler` receives each notification's `level`, optional `logger`, `data` payload, and any protocol `meta`. When `level` is omitted the server decides which messages to emit. Requesting a `level` is best-effort: if the server rejects it the connection stays up and a warning is logged.
 
-Only enable `serverLogging` for servers you trust. A server controls how often it logs and what `data`/`meta` it sends, so treat those payloads as untrusted input. Keep your `handler` from throwing: exceptions raised inside it are swallowed by the MCP client and your logs will silently stop flowing.
+Only enable `serverLogging` for servers you trust. A server controls how often it logs and what `data`/`meta` it sends, so treat those payloads as untrusted input. If your `handler` throws, the error is caught and logged as a warning rather than interrupting the run, but the offending notification is dropped, so prefer a handler that does not throw.
 
 ### Tool filtering
 
--- a/docs/src/content/docs/guides/mcp.mdx
+++ b/docs/src/content/docs/guides/mcp.mdx
@@ -295,6 +295,29 @@ With this setting, a `search` tool from a `docs` server is exposed to the model 
 
 Generated names are ASCII-safe, stay within the function-tool name limit, and avoid local function tool names and enabled handoff names on the same agent. The setting only affects local Streamable HTTP and stdio MCP tools; hosted MCP tools keep their hosted server labels and tool metadata.
 
+### Server logging
+
+Stdio, SSE, and Streamable HTTP MCP servers can emit `notifications/message` logging events while a tool runs. This is useful for surfacing progress from long-running tools instead of waiting silently. Subscribe by passing `serverLogging` to the server:
+
+```ts
+const mcpServer = new MCPServerStdio({
+  name: 'Logging Demo Server',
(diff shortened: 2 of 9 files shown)


hdimer:
Thanks for the review! Addressed both suggestions in b54bd16:

- **Handler error boundary** — the `serverLogging.handler` call in the `notifications/message` callback is now wrapped in a try/catch that logs a warning, so a throwing or rejecting handler can't surface as an unhandled rejection. Updated the docs to describe this caught-and-warned behavior.
- **Public types** — re-exported `MCPServerLoggingLevel`, `MCPServerLoggingMessage`, `MCPServerLoggingHandler`, and `MCPServerLoggingOptions` from the package root alongside `MCPToolErrorFunction`, so consumers can type reusable logging configurations.

Added a regression test for the throwing-handler case. `pnpm build && pnpm -r build-check && pnpm test && pnpm lint` all pass.

seratch:
Thanks for sharing this implementation idea. We're currently focused on other priorities for the upcoming release, plus this feature request isn't a priority at the moment. If we see increasing demand from developers, we may revisit it in the future.

hdimer:
Thanks for the thoughtful response, and no worries on prioritization, that makes sense. I'll leave the PR open in case demand picks up. I've also already addressed the two automated review notes (the handler now catches and warns on a throwing/rejecting `serverLogging.handler` so it can't surface as an unhandled rejection, and the public `MCPServerLogging*` types are re-exported from the package root), with a regression test, so it's in a clean, mergeable state if you ever revisit.

seratch:
Closing this PR for the reason I mentioned at https://github.com/openai/openai-agents-js/issues/793#issuecomment-5204148843

PR #1516, opened 2026-07-24
fix: honor sensitive logging flags across runtime paths

seratch:
This pull request fixes runtime logging paths that could expose model data, tool data, session history, tracing payloads, MCP content, Realtime messages, or arbitrary thrown values when sensitive-data logging was disabled.

It centralizes safe redaction for error, warning, and debug paths; guards payload formatting across core, extensions, and OpenAI integrations; and adds adversarial regression coverage. Diagnostic details remain available when logging flags are disabled, while redacted mode emits only fixed messages and safe value types.


seratch:
@codex review

PR #1520, opened 2026-07-26
docs: add errorFunction example to Tools guide

PengYYYYY:
## Summary
The Tools guide lists \`errorFunction\` in the function-tool options table and tells readers not to throw in error handlers, but it never shows a runnable example of \`errorFunction\` itself.

This PR adds a short "Handling tool errors" subsection with a complete example: a tool whose \`execute\` throws, and an \`errorFunction\` that converts the error into a model-visible message so the run does not crash and the agent can recover or explain the failure.

## Notes
- \`errorFunction\` receives \`(context, error)\` and should return a string rather than throwing.
- Mirrors the error-handling pattern of returning tool errors as normal results.

### DIFF
--- a/.changeset/tool-error-example.md
+++ b/.changeset/tool-error-example.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents": patch
+---
+
+Docs: add `errorFunction` usage example to the Tools guide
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing the idea. Since [this section](https://openai.github.io/openai-agents-js/guides/tools/#options-reference) already mentions the option, we don't plan to add code snippets suggested here.

PR #1525, opened 2026-07-27
fix(core): stop cancelled streams before another model turn

seratch:
This pull request is a follow-up for https://github.com/openai/openai-agents-js/pull/1523 and  fixes a streamed-run cancellation race where an in-flight function tool could observe the abort signal, settle, and then allow the runner to begin another model turn. It adds a single cancellation boundary to the streaming loop so the current action batch and RunState updates settle before orchestration stops. Regression coverage verifies both direct and approval-resumed function-tool runs make exactly one model invocation.

### DIFF
--- a/.changeset/lucky-badgers-itch.md
+++ b/.changeset/lucky-badgers-itch.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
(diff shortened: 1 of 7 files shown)


seratch:
Closing this as its foundation #1523 was reverted

PR #1538, opened 2026-07-29
fix(sandbox): support native Windows paths in path grants

### LINKED ISSUE #1537: Windows host paths can't be expressed in sandbox path grants
**Environment**: Windows 11, Node 24.16.0, main branch (7ff4a9a3), repro against the built dist of `@openai/agents-core`.

While trying the sandbox feature on Windows, I hit a wall with `extraPathGrants`. The docs point Windows users to `DockerSandboxClient` (guides/sandbox-agents/concepts, "On Windows, use DockerSandboxClient or a hosted provider instead") and describe `extraPathGrants` as the way to grant access to "another absolute host directory". But there seems to be no spelling of a Windows host path that `normalizePathGrant` accepts:

```ts
import { Manifest } from '@openai/agents/sandbox';

new Manifest({ extraPathGrants: [{ path: 'C:\\Users\\me\\skills' }] });
// Error: Sandbox path grant path must use "/" separators.

new Manifest({ extraPathGrants: [{ path: 'C:/Users/me/skills' }] });
// Error: Sandbox path grant path must be absolute.

new Manifest({ extraPathGrants: [{ path: '\\\\server\\share\\skills' }] });
// Error: Sandbox path grant path must use "/" separators.

new Manifest({ extraPathGrants: [{ path: '//server/share/skills' }] });
// accepted, but silently normalized to "/server/share/skills"
```

The last case worries me most: a UNC path in forward-slash form passes validation but comes out as a different path.

From reading the code, the grant path is treated as a host path downstream:

- `dockerExtraPathGrantMountArgs` in `docker.ts` uses `grant.path` as both the bind mount source and the target
- `resolveLocalSourcePath` in `sandboxes/shared/localWorkspace.ts` compares `resolve(grant.path)` against resolved host paths
- the docs example (`examples/docs/sandbox-agents/path-grants.ts`) uses the same string for the grant and for a real host fs read

So on a Windows host the POSIX-only validation in `pathGrants.ts` makes the feature unreachable. Even if validation accepted drive letters, the source=target bind mount design would need a decision about what the container side should look like, since `C:\Users\...` is not a valid path inside a Linux container.

Is Windows host path support intended here? If the near-term answer is "not supported", a clearer error plus a note in the docs would already help. If you do want to support it, I can work on a PR in whichever direction you prefer, and I can test on real Windows hardware.

seratch:
This pull request resolves #1537.

It adds an optional `hostPath` to sandbox path grants, keeping `path` as the POSIX path visible inside the sandbox while using the native path for host authorization and Docker bind mounts. Windows paths must be drive-qualified, unsupported UnixLocal combinations fail early, and the documentation and Windows CI coverage are updated.

Native host paths are excluded from serialized session state and restored only from the current trusted manifest. Direct resumes reject unresolved redacted paths, while interrupted runs preserve runtime-agent identities so sessions can be safely rebound.

### DIFF
--- a/.changeset/native-sandbox-path-grants.md
+++ b/.changeset/native-sandbox-path-grants.md
@@ -1,0 +1,6 @@
+---
+'@openai/agents-core': patch
+'@openai/agents-extensions': patch
+---
+
+fix: support native Windows host paths in sandbox path grants (#1537)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -173,7 +173,7 @@ jobs:
       - name: Run Windows sandbox path tests
         if: ${{ needs.changes.outputs.run_ci == 'true' }}
         shell: bash
-        run: pnpm exec vitest --run --project "@openai/agents-core" packages/agents-core/test/sandboxWorkspacePaths.test.ts packages/agents-core/test/sandboxManifest.test.ts
+        run: pnpm exec vitest --run --project "@openai/agents-core" packages/agents-core/test/sandboxWorkspacePaths.test.ts packages/agents-core/test/sandboxManifest.test.ts packages/agents-core/test/sandboxNativePathGrants.test.ts
       - name: Run Windows remote path tests
         if: ${{ needs.changes.outputs.run_ci == 'true' }}
         shell: bash
--- a/packages/agents-core/src/sandbox/internal.ts
+++ b/packages/agents-core/src/sandbox/internal.ts
@@ -1,9 +1,12 @@
 export {
+  assertHostPathGrantsRebound,
+  deserializeHostPathGrantRedactionMetadata,
   deserializeManifest,
   mergeManifestEntryDelta,
   mergeManifestDelta,
   sanitizeEnvironmentForPersistence,
   serializeEnvironmentForPersistence,
+  serializeHostPathGrantRedactionMetadata,
   serializeManifest,
   serializeManifestRecord,
 } from './sandboxes/shared/manifestPersistence';
@@ -17,6 +20,7 @@ export {
   isHostPathWithinRoot,
   relativeHostPathEscapesRoot,
   relativeHostPathEscapesRootOrSelf,
+  sandboxPathGrantHostPath,
 } from './shared/hostPath';
 export {
   hasBackslashPathSeparator,
--- a/packages/agents-core/src/sandbox/localSkills.ts
+++ b/packages/agents-core/src/sandbox/localSkills.ts
@@ -8,7 +8,10 @@ import {
 import { localDir } from './entries';
 import type { Manifest } from './manifest';
 import type { SandboxPathGrant } from './pathGrants';
-import { isHostPathWithinRoot } from './shared/hostPath';
+import {
+  isHostPathWithinRoot,
+  sandboxPathGrantHostPath,
+} from './shared/hostPath';
 
 export type LocalDirLazySkillSourceOptions = {
   /**
@@ -86,7 +89,10 @@ function resolveLocalSkillSourcePath(
   if (
     isHostPathWithinRoot(base, resolvedSourcePath) ||
     sourceGrants.some((grant) =>
-      isHostPathWithinRoot(resolve(grant.path), resolvedSourcePath),
+      isHostPathWithinRoot(
+        resolve(sandboxPathGrantHostPath(grant)),
+        resolvedSourcePath,
+      ),
     )
   ) {
     return resolvedSourcePath;
--- a/packages/agents-core/src/sandbox/pathGrants.ts
+++ b/packages/agents-core/src/sandbox/pathGrants.ts
@@ -5,13 +5,19 @@ import {
 } from './shared/posixPath';
 
 export type SandboxPathGrantInit = {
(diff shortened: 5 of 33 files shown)


Sora-bluesky:
Tested this branch (861b1228) on real Windows 11 with Docker Desktop (engine 29.6.2, WSL2 backend), Node 24.16.0. Everything works as designed.

**Public API matrix** (`new Manifest({ extraPathGrants: [...] })`, against the built dist):

```
path only (backward compat):
  /opt/skills                      -> OK (unchanged)
  C:\Users\me\skills               -> rejected (correct: path is the container-visible POSIX side)
hostPath variants (path=/mnt/shared-data):
  hostPath C:\Users\me\skills      -> OK, kept as C:\Users\me\skills
  hostPath C:/Users/me/skills      -> OK, normalized to C:\Users\me\skills
  hostPath \\server\share\skills   -> rejected: does not support UNC or device paths
  hostPath //server/share/skills   -> rejected: same error (the silent // collapse from #1537 is closed on this field)
  hostPath relative/skills         -> rejected: must be an absolute host path
  hostPath C:\Users\..\me          -> rejected: must not contain parent segments
consumption (sandboxPathGrantHostPath on win32):
  explicit C:\ hostPath            -> returned as-is
  omitted hostPath                 -> falls back to path
  explicit /opt/skills on Windows  -> rejected: must be drive-qualified on Windows
```

**End-to-end Docker bind mount** with a real Windows host directory:

```js
const client = new DockerSandboxClient({ image: 'node:22-bookworm-slim' });
const manifest = new Manifest({
  extraPathGrants: [{ path: '/mnt/shared-data', hostPath: 'C:\\Users\\<user>\\AppData\\Local\\Temp\\claude', readOnly: true }],
});
const session = await client.create(manifest);
await session.execCommand({ cmd: 'cat /mnt/shared-data/mount-probe.txt' });
// -> Process exited with code 0, Output: probe
await session.execCommand({ cmd: 'touch /mnt/shared-data/write-test || echo WRITE_BLOCKED' });
// -> touch: cannot touch '...': Read-only file system, WRITE_BLOCKED
```

The container read the host file through the grant, and readOnly was enforced by the mount.

**Tests on real Windows**: the three files from the Windows CI job (`sandboxNativePathGrants.test.ts`, `sandboxWorkspacePaths.test.ts`, `sandboxManifest.test.ts`) all pass here (53 passed, 0 failed), including the win32-only drive-less rejection case that CI can now actually exercise. The pre-existing Windows failures in the unixLocal/docker unit suites are unchanged on this branch (49 before, 49 after, same distribution), so no regressions from this change.

One small note: passing `//server/share` as `path` still collapses to `/server/share` via normalizePosixPath, but since `path` is now clearly the container-visible POSIX side and hostPath rejects UNC explicitly, I don't think that matters for the host-path concern from #1537.

Thanks for the quick turnaround on this.

Sora-bluesky:
Verified once more on merged main (73abbdc7): the validation matrix and the end-to-end Docker bind mount (read plus readOnly enforcement) both pass on Windows 11. Thanks for the quick design and turnaround here, this unblocks the Windows plus Docker sandbox path.

For the record, one observation related to the last Codex P1: with a split grant on this Windows host, the no-runAs filesystem helpers (readFile, listDir, pathExists) reject the grant path at the workspace policy layer ("escapes the workspace root") rather than opening anything on the host, and execCommand works through the mount as expected. So the helpers can't reach the granted area, but nothing leaks. Happy to file a separate issue if that gap is worth tracking.

seratch:
@Sora-bluesky Thanks, this is useful. Please file a separate issue with the exact `Manifest`, the `readFile`/`listDir`/`pathExists` calls, whether this was a freshly created or resumed session, and the full error stack.

The merged implementation is intended to route these no-`runAs` operations through the container filesystem for split grants, and we have unit coverage for that behavior. So this looks like a real Windows-specific path or lifecycle gap rather than an intentional limitation. The observed behavior is fail-closed on Windows, but we should still understand why the Docker routing was missed and add regression coverage for the exact scenario.

Sora-bluesky:
Correction, and my apologies: the helper observations in my last comment were captured at 468c944, and I failed to re-run that part against merged main before commenting. The matrix and execCommand checks were re-run on main; the readFile/listDir/pathExists probes were not.

On merged main (73abbdc7), with the exact same fresh-session setup (path: /mnt/shared-data, hostPath: C:\..., readOnly, fresh create, no resume), all three helpers work:

```
execCommand cat /mnt/shared-data/mount-probe.txt -> exit 0, "probe"
session.readFile(...)   -> returns the file contents
session.listDir(...)    -> lists the granted directory
session.pathExists(...) -> true
```

Reading the diff, this is the dockerSplitPathGrantContainingPath routing from the final push (plus the WorkspacePathPolicy-based resolution in listDir). So the specific readFile/listDir/pathExists gap I reported does not exist on merged main and there is nothing to understand or cover there; the 468c944 behavior I measured was simply the pre-fix state.

While re-verifying I did find one adjacent case that still reproduces on merged main: execCommand with workdir pointing into a split grant. Filed as #1539 with the exact manifest and full stack, as requested. Sorry for the earlier confusion.

PR #1540, opened 2026-07-29
fix(sandbox): allow Docker workdirs within path grants

### LINKED ISSUE #1539: Docker split path grants: execCommand workdir inside a granted path is rejected
**Environment**: Windows 11, Node 24.16.0, Docker Desktop (engine 29.6.2, WSL2 backend), main at 73abbdc7.

Follow-up from #1538, as asked there. The readFile/listDir/pathExists behavior I described turned out to be measured on the stale pre-fix head 468c944; on merged main all three work through the container routing, so that part needs no action. This issue is the one adjacent case that still reproduces on merged main.

**Exact repro** (fresh session, no resume):

```js
const client = new DockerSandboxClient({ image: 'node:22-bookworm-slim' });
const manifest = new Manifest({
  extraPathGrants: [
    { path: '/mnt/shared-data', hostPath: 'C:\\Users\\<user>\\AppData\\Local\\Temp\\claude', readOnly: true },
  ],
});
const session = await client.create(manifest);

// works (container routing):
await session.execCommand({ cmd: 'cat /mnt/shared-data/mount-probe.txt' }); // exit 0
await session.readFile({ path: '/mnt/shared-data/mount-probe.txt' });       // file contents
await session.listDir({ path: '/mnt/shared-data' });                        // entries
await session.pathExists('/mnt/shared-data/mount-probe.txt');               // true

// fails:
await session.execCommand({ cmd: 'pwd', workdir: '/mnt/shared-data' });
```

**Full error stack**:

```
UserError: Sandbox path "/mnt/shared-data" must stay within /workspace.
    at DockerSandboxSession.resolveLogicalPath (dist/sandbox/sandboxes/unixLocal.js:368:19)
    at DockerSandboxSession.resolveCommandWorkdir (dist/sandbox/sandboxes/docker.js:223:34)
    at DockerSandboxSession.exec (dist/sandbox/sandboxes/unixLocal.js:74:26)
    at DockerSandboxSession.execCommand (dist/sandbox/sandboxes/unixLocal.js:70:60)
```

So workdir resolution goes through resolveLogicalPath, which only accepts workspace-relative paths, while cmd paths and the fs helpers accept granted paths. The same command works with the default workdir and an absolute path in cmd.

If workdir is intentionally workspace-only, feel free to close this; a doc note on the grants docs would then cover it.

seratch:
This pull request resolves #1539.

`DockerSandboxSession` previously resolved `execCommand` workdirs using workspace-only logical path handling. As a result, container paths covered by `extraPathGrants` were rejected before `docker exec`.

This change reuses the existing container filesystem path policy for command workdirs. Granted paths are forwarded through Docker's `-w` option, while ungranted absolute paths continue to fail before process creation. Regression coverage verifies both behaviors.

### DIFF
--- a/.changeset/docker-path-grant-workdirs.md
+++ b/.changeset/docker-path-grant-workdirs.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: allow Docker command workdirs within sandbox path grants (#1539)
--- a/packages/agents-core/src/sandbox/sandboxes/docker.ts
+++ b/packages/agents-core/src/sandbox/sandboxes/docker.ts
@@ -137,6 +137,18 @@ export interface DockerSandboxSessionState extends UnixLocalSandboxSessionState 
 
 export class DockerSandboxSession extends UnixLocalSandboxSession<DockerSandboxSessionState> {
   private containerClosed = false;
+  private readonly mountedPathGrants: Manifest['extraPathGrants'];
+
+  constructor(args: {
+    state: DockerSandboxSessionState;
+    defaultShell?: string;
+    archiveLimits?: SandboxArchiveLimits | null;
+  }) {
+    super(args);
+    this.mountedPathGrants = args.state.manifest.extraPathGrants.map(
+      (grant) => ({ ...grant }),
+    );
+  }
 
   override async resolveFilesystemRunAs(runAs?: string): Promise<undefined> {
     if (runAs && runAs.trim().length > 0) {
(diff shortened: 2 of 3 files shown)


Sora-bluesky:
Verified on this branch (0c67dce4) on Windows 11, Docker Desktop (engine 29.6.2, WSL2 backend), Node 24.16.0, with the same split-grant setup as #1539 (path: /mnt/shared-data, hostPath: C:\..., readOnly, fresh session):

```
workdir /mnt/shared-data                -> exit 0, pwd prints /mnt/shared-data   (threw on 73abbdc7)
workdir /mnt/shared-data/bundled-skills -> exit 0, pwd prints the subdir
workdir /etc                            -> rejected: escapes the workspace root
workdir /mnt/shared-data-evil           -> rejected (no prefix confusion with the granted path)
workdir /workspace                      -> exit 0 (unchanged)
```

No regressions in the adjacent surfaces on this branch: execCommand with absolute cmd paths, readFile, listDir, and pathExists all still work through the grant, and the new workdir regression test passes here on real Windows.

Sora-bluesky:
Re-ran the same checks on merged main (e4158f1f) after the follow-up commits. Same results: granted workdir and subdir exit 0, /etc and the prefix-confusion path are still rejected, and the fs helpers are unaffected.

PR #1558, opened 2026-08-02
feat: support MCP SDK v2 negotiation

seratch:
This pull request adds MCP TypeScript SDK v2 support to the local MCP clients. Stdio and Streamable HTTP clients now negotiate MCP 2026-07-28 with legacy fallback, while deprecated SSE remains legacy-only.

It preserves the released resource pagination, cache invalidation, cancellation, required-task rejection, and reconnect behavior. It also adds real v2 client coverage for modern and legacy negotiation, explicit sessions, session recovery, and stale tool-list isolation.

The exact MCP v2 client and core releases are excluded from the dependency freshness window because installation was otherwise blocked.

see also: https://github.com/openai/openai-agents-python/pull/4106

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -22,6 +22,8 @@ overrides:
 publishBranch: main
 minimumReleaseAge: 10080
 minimumReleaseAgeExclude:
+  - '@modelcontextprotocol/client@2.0.0'
+  - '@modelcontextprotocol/core@2.0.0'
   - openai@6.46.0
 blockExoticSubdeps: true
 strictDepBuilds: true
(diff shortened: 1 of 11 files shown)


pnpm-workspace.yaml
```diff
@@ -22,6 +22,8 @@ overrides:
 publishBranch: main
 minimumReleaseAge: 10080
 minimumReleaseAgeExclude:
+  - '@modelcontextprotocol/client@2.0.0'
+  - '@modelcontextprotocol/core@2.0.0'
```
seratch: TODO: remove these once 7 days pass after 2.0.0 release

PR #1568, opened 2026-08-04
fix: preserve final tool results on guardrail trip

seratch:
This pull request fixes session persistence when an output guardrail trips after a final tool has already executed.

It persists each committed tool call/result pair exactly once, retains only reasoning associated with retained tool calls, and excludes the rejected assistant message and unrelated reasoning. The same persistence path now covers streaming and non-streaming runs, while preserving full ordered output on successful guardrail checks and maintaining existing approved-tool checkpoint and compaction ownership.


seratch:
As mentioned at https://github.com/openai/openai-agents-js/pull/1572, we'll go with a different approach.

PR #1573, opened 2026-08-04
fix(realtime): accept conversation items without a status

### LINKED ISSUE #1571: Realtime: item `status` is required in realtimeMessageItemSchema but optional in the Realtime API — status-less items are dropped from history
## Summary

The Realtime item Zod schemas make the conversation item `status` field **required**, but the OpenAI Realtime API marks it **optional**. When a spec-compliant server sends a `conversation.item.added` / `conversation.item.done` / `conversation.item.retrieved` event whose `item` omits `status` (or sends a value the schema doesn't include), `realtimeMessageItemSchema.parse(...)` throws inside the message handler. The item never enters the SDK's history, and — for user message items — user transcripts silently disappear from `history` / `history_updated`, leaving only assistant text.

## Expected behavior

Per the Realtime API, a conversation item's `status` is optional:

```ts
// openai (openai-node) — RealtimeConversationItem*
status?: 'completed' | 'incomplete' | 'in_progress';
```

The SDK should accept items that omit `status` (and shouldn't drop them from history).

## Actual behavior

In `packages/agents-realtime/src/items.ts`, the message item schemas require `status` and, for user messages, omit `'incomplete'`:

```ts
// user message
status: z.enum(['in_progress', 'completed']),          // required; missing 'incomplete'
// assistant message
status: z.enum(['in_progress', 'completed', 'incomplete']), // required
```

In `packages/agents-realtime/src/openaiRealtimeBase.ts`, the handler for `conversation.item.added` / `.done` / `.retrieved` calls `realtimeMessageItemSchema.parse(...)` when `item.type === 'message'`. Because `status` is required, an item that omits it throws and is dropped.

## Reproduction

1. Connect the Realtime SDK to any server that emits a user message item without a `status` field (valid per the API), e.g.:
   ```json
   {
     "type": "conversation.item.added",
     "item": {
       "id": "msg_1",
       "type": "message",
       "role": "user",
       "content": [{ "type": "input_audio", "transcript": null }]
     }
   }
   ```
2. Observe that the item never appears in `session.history` / the `history_updated` event; the parse throws internally.

## Root cause

`status` on the message item schemas is not marked `.optional()`, so a valid (status-omitted) item fails validation. User messages additionally reject `'incomplete'`, which the API type permits.

## Suggested fix

- Make `status` optional on the item schemas (e.g. `z.enum([...]).optional()`), matching `status?:` in the API types.
- Align the user message enum with the API (`'in_progress' | 'completed' | 'incomplete'`).
- Consider not dropping an item from history when only `status` fails to validate.

## Environment

- `@openai/agents-realtime`: 0.14.2 (schema is unchanged on `main` at time of filing)
- Node: any — this is a schema-validation issue, not runtime-specific
- Server: an OpenAI Realtime-compatible endpoint (not OpenAI's own)

## Workaround

Send an explicit valid `status` (e.g. `"completed"`) on every message item so it satisfies the required schema. (Clients can also rebuild the transcript from raw transport events instead of SDK history.)

---

<sub>Written by claude-opus-4-8[1m], using Claude Code, on behalf of @kraenhansen.</sub>

aryanputta:
Closes #1571.

### Problem

The Realtime API marks a conversation item's `status` optional, and permits `incomplete` on user items:

```ts
status?: 'completed' | 'incomplete' | 'in_progress';
```

`realtimeMessageItemSchema` required `status`, and the user variant allowed only `in_progress | completed`. A spec-compliant `conversation.item.added`, `conversation.item.done`, or `conversation.item.retrieved` event whose item omits `status` therefore threw inside the message handler, so the item never reached history. For user items that dropped the transcript silently, leaving `history` and `history_updated` with assistant text only.

The `response.output_item.*` path was already defensive (`item.status ?? 'completed'`), so the gap was confined to the `conversation.item.*` path, which passes `parsed.item.status` straight through.

### Change

Make `status` optional on the user and assistant variants, and align the user enum with the API by adding `incomplete`. This fixes it at the schema rather than at one call site, so any future consumer of the schema inherits the correct contract.

### Verification

- Two tests added to `openaiRealtimeBase.test.ts`, exercising the reported path end to end via `_onMessage`: an item with no `status`, and a user item marked `incomplete`. Both fail on `main` and pass with this change.
- `pnpm build && pnpm -r build-check && pnpm test && pnpm lint` all pass. Full suite: 4201 passed. The only failures are the two pre-existing `sandboxes/docker.test.ts` cases, which fail identically on a clean checkout without this diff.
- Changeset included, per CONTRIBUTING.

### DIFF
--- a/.changeset/olive-items-keep-status.md
+++ b/.changeset/olive-items-keep-status.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: keep realtime conversation items whose status the server omitted or marked incomplete
--- a/.changeset/olive-items-keep-status.md
+++ b/.changeset/olive-items-keep-status.md
@@ -2,4 +2,4 @@
 '@openai/agents-realtime': patch
 ---
 
-fix: keep realtime conversation items whose status the server omitted or marked incomplete
+fix: normalize a missing realtime item status from the event phase so status-less items still reach history
--- a/packages/agents-realtime/src/items.ts
+++ b/packages/agents-realtime/src/items.ts
@@ -19,7 +19,7 @@ export const realtimeMessageItemSchema = z.discriminatedUnion('role', [
     previousItemId: z.string().nullable().optional(),
     type: z.literal('message'),
     role: z.literal('user'),
-    status: z.enum(['in_progress', 'completed']),
+    status: z.enum(['in_progress', 'completed', 'incomplete']).optional(),
     content: z.array(
       z.object({ type: z.literal('input_text'), text: z.string() }).or(
         z.object({
@@ -35,7 +35,7 @@ export const realtimeMessageItemSchema = z.discriminatedUnion('role', [
     previousItemId: z.string().nullable().optional(),
     type: z.literal('message'),
     role: z.literal('assistant'),
-    status: z.enum(['in_progress', 'completed', 'incomplete']),
+    status: z.enum(['in_progress', 'completed', 'incomplete']).optional(),
     content: z.array(
       z.object({ type: z.literal('output_text'), text: z.string() }).or(
         z.object({
--- a/packages/agents-realtime/src/items.ts
+++ b/packages/agents-realtime/src/items.ts
@@ -19,7 +19,7 @@ export const realtimeMessageItemSchema = z.discriminatedUnion('role', [
     previousItemId: z.string().nullable().optional(),
     type: z.literal('message'),
     role: z.literal('user'),
-    status: z.enum(['in_progress', 'completed', 'incomplete']).optional(),
+    status: z.enum(['in_progress', 'completed']),
     content: z.array(
       z.object({ type: z.literal('input_text'), text: z.string() }).or(
         z.object({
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Before merging, please keep the released `RealtimeMessageItem` contract non-optional. This type is publicly exported, and making `status` optional is a breaking TypeScript change while the changeset is patch-level. Normalize a missing status at the OpenAI event adapter using the existing event-phase behavior (`added` -> `in_progress`, `done`/`retrieved` -> `completed`), while preserving an explicit status including `incomplete`. Please update the regression coverage to drive a `RealtimeSession` and assert that the item reaches `history`/`history_updated` with the normalized status. This fixes the wire-schema mismatch without widening the public history contract.

aryanputta:
Thanks, that's right — making `status` optional widened the exported `RealtimeMessageItem` under a patch changeset, which I should not have done.

Reworked in d5e447b. `realtimeMessageItemSchema` is back to its original shape, and the normalization now happens at the OpenAI event adapter using the phase the event already carries: `conversation.item.added` → `in_progress`, `conversation.item.done` and `conversation.item.retrieved` → `completed`. An explicit status passes through untouched, including `incomplete`.

Coverage now drives a `RealtimeSession` and asserts the item reaches `history` and `history_updated` with the normalized status, plus a per-phase check on the adapter. Both fail without the change. `pnpm build && pnpm -r build-check && pnpm test && pnpm lint` pass; 4201 tests green with the two pre-existing `sandboxes/docker.test.ts` failures unchanged from a clean checkout.

seratch:
https://github.com/openai/openai-agents-js/pull/1582 resolves the issue. Its commit includes you as a co-author.

PR #1586, opened 2026-08-05
fix(realtime): keep a caller-supplied WebRTC media stream alive on close

adityasingh2400:
`OpenAIRealtimeWebRTC.close()` stops every sender track on the peer connection, which includes the track taken from a caller-supplied `options.mediaStream`. That option exists so an application can share or preprocess a stream it owns, so ending its track leaves any level meter, recorder, or second consumer dead, and a later reconnect adds an already ended track so the model just hears silence. It is easy to hit without calling `close()` directly, because the transport also closes itself when the peer connection reports `failed` or stays `disconnected` past the grace period.

The connect path already encodes the right rule: when it abandons an attempt it stops microphone tracks only `if (!this.options.mediaStream)`. This applies the same condition in `close()`, so the transport stops only the microphone it opened itself and always closes the peer connection either way. Two tests cover both sides, and the caller-owned case fails on main while the SDK-owned case keeps passing. `vitest run test/openaiRealtimeWebRtc.test.ts` is green at 38 passed, along with eslint and prettier on the changed files.

### DIFF
--- a/.changeset/quiet-webrtc-keeps-mic.md
+++ b/.changeset/quiet-webrtc-keeps-mic.md
@@ -5,3 +5,5 @@
 fix(realtime): keep a caller-supplied WebRTC media stream alive on close
 
 `OpenAIRealtimeWebRTC.close()` no longer stops the tracks of a `mediaStream` passed in through the transport options. That stream now stays owned by the application, so it remains usable for a level meter, a recorder, or a later reconnect, and the application is responsible for calling `stop()` on its tracks when it is done with it. A microphone the transport opens itself, when `mediaStream` is omitted, is still stopped by `close()` as before.
+
+If the session was muted, `close()` now re-enables the caller-supplied tracks it disabled, so the stream is not handed back permanently silent. Tracks the application disabled itself are left alone.
--- a/.changeset/quiet-webrtc-keeps-mic.md
+++ b/.changeset/quiet-webrtc-keeps-mic.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix(realtime): keep a caller-supplied WebRTC media stream alive on close
--- a/.changeset/quiet-webrtc-keeps-mic.md
+++ b/.changeset/quiet-webrtc-keeps-mic.md
@@ -3,3 +3,5 @@
 ---
 
 fix(realtime): keep a caller-supplied WebRTC media stream alive on close
+
+`OpenAIRealtimeWebRTC.close()` no longer stops the tracks of a `mediaStream` passed in through the transport options. That stream now stays owned by the application, so it remains usable for a level meter, a recorder, or a later reconnect, and the application is responsible for calling `stop()` on its tracks when it is done with it. A microphone the transport opens itself, when `mediaStream` is omitted, is still stopped by `close()` as before.
--- a/.changeset/quiet-webrtc-keeps-mic.md
+++ b/.changeset/quiet-webrtc-keeps-mic.md
@@ -6,4 +6,4 @@ fix(realtime): keep a caller-supplied WebRTC media stream alive on close
 
 `OpenAIRealtimeWebRTC.close()` no longer stops the tracks of a `mediaStream` passed in through the transport options. That stream now stays owned by the application, so it remains usable for a level meter, a recorder, or a later reconnect, and the application is responsible for calling `stop()` on its tracks when it is done with it. A microphone the transport opens itself, when `mediaStream` is omitted, is still stopped by `close()` as before.
 
-If the session was muted, `close()` now re-enables the caller-supplied tracks it disabled, so the stream is not handed back permanently silent. Tracks the application disabled itself are left alone.
+`close()` leaves the mute state alone. A caller-supplied track keeps whatever `enabled` value it had, and `muted` keeps reporting what it reported before the close, matching the default microphone path.
--- a/examples/docs/voice-agents/customWebRTCTransport.ts
+++ b/examples/docs/voice-agents/customWebRTCTransport.ts
@@ -10,10 +10,20 @@ const agent = new RealtimeAgent({
 });
 
 async function main() {
+  // Keep a handle on the stream you pass in. The transport does not stop a caller-supplied
+  // stream on close(), so your application owns it and is responsible for ending it.
+  const mediaStream = await navigator.mediaDevices.getUserMedia({
+    audio: true,
+  });
+
   const transport = new OpenAIRealtimeWebRTC({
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Before merge, please make caller ownership explicit in the mediaStream JSDoc and user-facing example: retain the supplied stream and stop its tracks when the application is actually done. Please also update the changeset to note that close() no longer stops caller-supplied tracks.

This is needed because the behavior shipped in v0.14.2 changes cleanup responsibility, and the current inline getUserMedia() example would otherwise leave no convenient cleanup handle. Also narrow the source comment from the general "only stop tracks this transport created" claim to the exact mediaStream case; changePeerConnection can add other senders. No broader per-track ownership redesign is needed.

adityasingh2400:
Thanks, all three are done in 45f8754.

The `mediaStream` JSDoc now states that the stream stays owned by the application, that `close()` will not stop its tracks, and that you need to keep a reference and stop them yourself once you are done, with the contrast that a microphone the transport opens for you is still stopped on close.

The `customWebRTCTransport` example no longer inlines `getUserMedia()` into the options object. It binds the stream to a `mediaStream` local first so there is a handle to clean up with, and carries the release call with a note that leaving it out keeps the microphone indicator on after the session closes.

The changeset now spells out the cleanup responsibility change rather than just naming the fix.

I also narrowed the source comment as you asked. It no longer claims a general "only stop tracks this transport created" rule and instead says the skip is scoped to the caller-supplied `mediaStream` case specifically, with senders added through `changePeerConnection` left to whoever added them.

Verified locally: the 38 tests in `openaiRealtimeWebRtc.test.ts` pass, eslint is clean on both touched source files, and the `examples/docs` typecheck error count is unchanged from the base at 7, with none of them in the file I edited.

adityasingh2400:
The Codex P2 above is correct, and it is a consequence of this PR rather than a pre-existing issue, so I fixed it in 8e31174.

`mute(true)` disables sender tracks in place instead of replacing them. Before this PR that did not matter, because `close()` stopped those tracks anyway. Now that the caller's stream survives, closing while muted handed the stream back with `enabled` still `false`, and since `close()` clears the peer connection a later `mute(false)` had nothing left to re-enable. Reconnecting with the same stream then added a disabled track and stayed silent, exactly as described.

I kept the fix inside the ownership rule you set. It matches the caller's audio tracks against the senders, so anything added through `changePeerConnection` is untouched, and it only runs when the transport did the muting, so a stream the application disabled itself is left alone.

Two tests cover it. The restore test fails on the previous commit with `expected false to be true` on `track.enabled`, and the second test pins the non-overreach case. The realtime WebRTC and session suites are green at 122 passed, with prettier and tsc clean. The changeset now notes the mute behavior alongside the ownership change.

One thing I deliberately did not change: `#muted` stays `true` across `close()`. Resetting it would also change the transport-owned microphone path, where `.muted` already reports stale state after a close, and that is pre-existing rather than something this PR introduces. Happy to fix that separately if you want it, but it looked like the broader change you asked me to avoid.

@seratch the three items from your review landed earlier in 45f8754, so I think this is ready for another look.

adityasingh2400:
This is the second Codex finding on the same piece of `close()`, so per the complexity-reset rule in AGENTS.md I treated it as a signal to narrow the design rather than add another condition. `ce55723` removes more than it adds, 10 lines for 25.

`close()` now undoes its own mute through the existing call:

```ts
if (this.#muted && this.options.mediaStream) {
  this.mute(false);
}
```

That runs before the state is cleared, so the peer connection is still there. It is the exact inverse of the `mute(true)` that disabled the tracks, it needs no new state field, and `close()` no longer reasons about individual track ownership at all. The sender and caller-track intersection from the previous commit is gone.

On the specific finding, I checked it before changing anything and it is not introduced by this PR. `mute(false)` on `main` sets `sender.track.enabled = true` unconditionally, so a caller-supplied track the application had disabled is re-enabled by a plain `mute(true)` then `mute(false)` pair with no `close()` involved. I confirmed that by running exactly that sequence against `main` behavior in the existing WebRTC test harness, and the track came back `enabled === true`. This PR does not touch `mute()`.

So the behavior Codex describes lives in `mute()`, not in this change, and routing `close()` through `mute(false)` keeps it in the one place where it already is instead of creating a second, slightly different notion of track ownership. Adding a `#tracksDisabledByMute` set would have given this PR a per-track ownership model that `mute()` itself does not have, which is the redesign @seratch asked me to avoid. If you would like `mute()` tightened to only restore what it disabled, I am happy to do that as its own PR, since it changes behavior for callers who never touch `close()`.

Verification: the regression test still fails on the previous commit with `expected false to be true`, the realtime WebRTC and session suites pass at 122, and prettier and tsc are clean.

seratch (changes requested):
One blocker remains in the current head: close() calls mute(false) for caller-supplied streams. That is not an exact inverse of mute(true): it enables every sender and sets #muted to false. As a result, a track the application had already disabled can be enabled by close(), and caller-stream close now changes the public mute state while the default-microphone path preserves it. The new "never muted" test does not exercise that sequence.

Please remove the close-time automatic unmute and its mute-specific changeset claim. Preserve the caller track's current enabled state and the released mute state across close, with a focused regression test. A broader per-track mute-state redesign should be handled separately.

adityasingh2400:
You are right, and my reasoning for that commit was wrong. Removed in bd861d6.

I described `mute(false)` as the exact inverse of `mute(true)`. It is not. It enables every sender rather than only the tracks that were disabled, it flips `#muted`, and it can enable a track the application had disabled before handing the stream over. The disagreement you point out is the clearest symptom: the caller-stream path was rewriting `muted` while the default microphone path preserved it, for no reason a caller could predict.

`close()` no longer touches the mute state at all. The caller's track keeps whatever `enabled` value it had, and `muted` keeps reporting what it reported before the close.

You were also right that the old "never muted" test did not exercise the sequence. It only covered a track the application disabled with no `mute()` call, so the re-enable path was never asserted. Three tests replace it:

- `preserves the mute state of a caller-supplied stream across close()` does the actual `mute(true)` then `close()` sequence and asserts both `track.enabled` and `rtc.muted` survive. On the previous commit it fails with `expected true to be false`.
- `preserves a caller-supplied track the application disabled itself` keeps the earlier non-overreach case.
- `leaves the mute state alone on the transport-owned microphone path too` pins the two paths agreeing, which is the invariant that broke.

The changeset no longer claims the mute behavior and now states that `close()` leaves the mute state alone.

I have not touched per-track mute state in `mute()` itself, since you said that belongs in a separate change.

Verification: the realtime WebRTC and session suites are 123 passed, with prettier and tsc clean.

seratch (approved):

PR #1588, opened 2026-08-05
feat(ai-sdk): support PDF file inputs

### LINKED ISSUE #630: Add support for files (e.g. PDF documents) while using AI SDK
### Describe the feature

We need the ability to pass PDF files using the `input_file` type to the model while using AI SDK. We are hoping to be able to use a `file_id` that references an uploaded file to the provider and/or passing the document data directly to the provider. 

I also saw that some of the code was recently moved around `input_file`, so we're curious what the future of supporting this is: https://github.com/openai/openai-agents-js/commit/0e01da00004aa511a6b45936f97fadf36183e126#diff-7ac46cf68db931f5ac06ee150253a093cfe51c238c02fed7cc5e07fd79f7c21dR105

In order to make this work for us, we've had to copy the `aiSdk` implementation and do something like this:
```typescript
return {
  type: 'file',
  data: c.file,
  mediaType: contentProviderData?.mediaType ?? 'application/octet-stream',
  filename: contentProviderData?.filename,
    providerOptions: {
  ...(contentProviderData ?? {}),
  },
};
```

seratch:
This pull request resolves #630 by adding portable PDF file input support to the AI SDK adapter.

It supports base64 data URLs, raw base64 with PDF metadata, and public HTTP(S) PDF URLs across AI SDK v2, v3, and v4 models. OpenAI-specific file IDs remain unsupported and produce an actionable error directing users to the Responses model integration.

The change preserves existing image behavior, content ordering, filenames, and model-scoped provider options. It includes focused regressions, an example, and a patch changeset.

### DIFF
--- a/.changeset/ai-sdk-file-inputs.md
+++ b/.changeset/ai-sdk-file-inputs.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+feat: #630 support PDF file inputs in the AI SDK adapter
--- a/packages/agents-extensions/src/ai-sdk/index.ts
+++ b/packages/agents-extensions/src/ai-sdk/index.ts
@@ -2,6 +2,7 @@ import type {
   JSONSchema7,
   LanguageModelV2 as LanguageModelV2Base,
   LanguageModelV2CallOptions,
+  LanguageModelV2FilePart,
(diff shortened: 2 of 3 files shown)


timothymorse:
Thank you doing this!

PR #1642, opened 2026-08-11
fix(realtime): seed user transcript into history when a tool call pre-empts item seeding

### LINKED ISSUE #141: The Realtime SDK is not emitting transcript / history event when a message triggers a tool call
### Describe the bug

The Realtime SDK is not emitting transcript / history events when a message from the user triggers a tool call 

### Expected behavior

I would expect to see the message in the history, I can see the tool call but not the message that triggered it

SnowingFox:
## Problem

The Realtime SDK drops the user's transcript / history entry when an audio utterance triggers a tool call. `packages/agents-realtime/src/utils.ts`'s `updateRealtimeHistory()` only merges the `conversation.item.input_audio_transcription.completed` transcript into a user message item that is *already present* in the history. When the utterance triggers a tool call, the user item has not been committed to the local history yet by the time the transcript event arrives, so `history.map(...)` finds nothing to update and the transcript is silently discarded (no `history_updated` event is emitted with the user message).

## Root cause

The `conversation.item.input_audio_transcription.completed` branch of `updateRealtimeHistory()` maps over the existing history and only updates an item whose `itemId` matches `event.item_id`. If no matching user message item is present (the tool-call turn pre-empts the item from being seeded into history), the branch returns the history unchanged and the transcript is lost. The transport's fallback (`conversation.item.retrieve` -> `conversation.item.retrieved` -> `item_update`) does not fire in the tool-call case either, as noted in the maintainer's investigation on the issue, so the transcript never reaches history through any path.

## Fix

In `updateRealtimeHistory()`, before attempting to merge into an existing item, check whether a matching user message item is present in the history. If it is not, seed a placeholder user message item (`type: 'message'`, `role: 'user'`, `status: 'completed'`, `input_audio` content carrying the transcript) and append it to the history. The transcript is now retained and surfaced via `history_updated` even when the tool call pre-empted the item from being committed. The existing in-place merge behavior (item already present) is unchanged.

## Test

Two new unit tests in `packages/agents-realtime/test/utils.test.ts`:

- `seeds a placeholder user message when a tool call pre-empts item seeding`
- `keeps the user transcript when the item is not yet in history`

Both fail on `main` (transcript dropped) and pass with the fix. The existing `merges input audio transcripts into history items` test continues to pass, confirming the in-place merge path is unchanged.

Fixes #141

### DIFF
--- a/.changeset/agents-realtime-history-transcript.md
+++ b/.changeset/agents-realtime-history-transcript.md
@@ -1,0 +1,5 @@
+---
+"@openai/agents-realtime": patch
+---
+
+Seed the user audio transcript into realtime history when a tool call pre-empts item seeding, so `conversation.item.input_audio_transcription.completed` transcripts are retained and surfaced via `history_updated`.
--- a/packages/agents-realtime/src/utils.ts
+++ b/packages/agents-realtime/src/utils.ts
@@ -239,6 +239,28 @@ export function updateRealtimeHistory(
 ): RealtimeItem[] {
   // Merge transcript into placeholder input_audio message
   if (event.type === 'conversation.item.input_audio_transcription.completed') {
+    const hasUserItem = history.some(
+      (item) =>
+        item.itemId === event.item_id &&
+        item.type === 'message' &&
+        'role' in item &&
+        item.role === 'user',
+    );
+
+    if (!hasUserItem) {
+      // The user item has not been committed to history yet (e.g. the utterance
+      // triggered a tool call before the transcript was emitted). Seed a
+      // placeholder user message so the transcript is not dropped.
+      const seededItem: RealtimeMessageItem = {
+        itemId: event.item_id,
+        type: 'message',
+        role: 'user',
+        status: 'completed',
+        content: [{ type: 'input_audio', transcript: event.transcript }],
+      };
+      return [...history, seededItem];
+    }
+
     return history.map((item) => {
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for the contribution. I traced #141 against the current GA Realtime contract and v0.15.0. The report predates GA support, while current Realtime sends `conversation.item.added` when input audio is committed; the supported session path already retains the transcript when a tool call follows. This PR's tests only construct the missing-item branch, so they do not show that current WebRTC behavior still omits the user item.

The fallback also creates a history item without `history_added`, loses the recovered transcript if a later `item_update` has a null transcript, and ignores `content_index`. Because the current need is not demonstrated, I do not think we should expand and repair this fallback yet. I suggest closing this PR for now. We can revisit #141 with a v0.15.0 reproduction that includes the sanitized raw event order around `conversation.item.added`, `conversation.item.done` or `conversation.item.retrieved`, the transcription completion, and the function call.

SnowingFox:
Thanks for the detailed explanation, @seratch — that makes sense given GA Realtime's current behavior (conversation.item.added on input commit + the session path already retaining transcripts after tool calls), and you're right that the fallback branch loses the recovered transcript on a later item_update and ignores content_index. Closing per your direction; if the issue is re-demonstrable against v0.15.0 I'm happy to revisit with a proper reproduction.

PR #1652, opened 2026-08-13
feat(core): support read-only model input filters

kmalkan:
### Summary

`callModelInputFilter` currently receives a deep copy of the full input on every model call. This is safe for filters that modify input, but expensive for filters that only read it.

We run long, tool-heavy agent turns inside Cloudflare Durable Objects, which have a fixed memory limit. As the history grows, repeatedly copying it increases peak memory use and can force us to compact earlier. It also creates new item objects on every call, so identity-based caches cannot be reused.

This PR lets a filter set `readonlyInput = true`. The filter still receives a new array, but the items inside it keep the same identity across model calls. This avoids one full-history copy and allows read-only caches to work across a Runner loop.

Safety stays the same for everyone else:

- Existing filters still receive a deep copy by default.
- Inputs sent to the model and saved to a session are still copied.
- A filter should opt in only if it and its helpers never modify input items.

This does not change the model's context window. It only reduces runtime memory overhead so constrained environments can use more of the available context before compacting.

### Test plan

- Added a real two-model-call Runner/tool-loop test covering stable prepared identity and raw model-request isolation.
- Added focused tests for default mutation isolation, repeated source references, injected occurrences, and independent model/session clones.
- Ran `pnpm exec vitest run packages/agents-core/test/run.test.ts packages/agents-core/test/runner/conversation.test.ts` (214 tests passed).
- Ran `.agents/skills/code-change-verification/scripts/run.sh`, including install, build, declaration/type checks, lint, all 5,634 tests, and changed-file formatting.
- Ran the staged-tree TruffleHog scan with zero findings.

### Issue number

N/A

### Checks

- [x] I've added/updated the relevant documentation
- [x] If using Codex, I've run `/review` before submitting this PR

### Performance

I ran a local benchmark through the real `Runner` boundary using 1,000 history items (~4.1 MB) and a 20-call tool loop. These are median results from 7 isolated runs per mode on an Apple M5 Pro with Node.js 25.9.

| Metric | Default | `readonlyInput` | Change |
| --- | ---: | ---: | ---: |
| `structuredClone` calls | 102,238 | 81,858 | -19.9% |
| No-op filter time | 930.5 ms | 923.8 ms | -0.7% |
| No-op maximum RSS | 441.7 MB | 410.2 MB | -31.5 MB (-7.1%) |
| Memoizing filter time | 1,121.6 ms | 1,027.2 ms | -8.4% |
| Identity-cache hit rate | 0% | 94.9% | +94.9 points |

For this input, the removed pre-filter clone avoids roughly 83 MB of repeated history copying across 20 calls. The no-op time changes very little because model and session copies still remain. The larger gain comes from preserving item identity, which lets read-only caches reuse work across calls.

This is a synthetic local benchmark, not a Cloudflare production measurement. Network time is excluded, and Node RSS should not be treated as Durable Object memory usage.

### DIFF
--- a/.agents/references/conversation-state-ownership.md
+++ b/.agents/references/conversation-state-ownership.md
@@ -23,7 +23,7 @@ Do not create two authoritative histories accidentally. When server-managed cont
 
 ## Filters, Sessions, and Resume
 
-- `callModelInputFilter` receives a deep-cloned prepared payload and must return an input array. Persist filtered clones so session history reflects redaction or truncation rather than the unfiltered source.
+- `callModelInputFilter` receives a deep-cloned prepared payload by default and must return an input array. A filter that sets `readonlyInput` receives a shallow-copied array whose item identities are preserved; that filter and its helpers must treat the items as immutable. Model and persisted session inputs remain isolated clones after filtering so session history reflects redaction or truncation rather than the unfiltered source.
 - Track filtered items back to their originals so delta bookkeeping marks the items the model actually received. Rewind or preserve that mapping according to whether a failed attempt may have advanced server state.
 - `RunState` persists conversation identifiers and primes tracker state from prior model responses. Resume must not resend acknowledged input, lose unsent outputs, or increment the turn count without a model call.
 - Conversation continuation carries context into a new turn. RunState resume continues a paused turn. Do not substitute one for the other.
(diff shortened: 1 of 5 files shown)


seratch:
Thanks again for sharing this idea. https://github.com/openai/openai-agents-js/pull/1654 completed this by applying additional changes to make the new option available for broader surfaces of the SDK. Your contribution credit is included in its commit. The option will be included in the upcoming release.

PR #1653, opened 2026-08-13
fix: preserve complete session transcripts across inline compaction

will-giraldo:
### Summary

This pull request fixes incomplete session persistence when a run produces inline compaction. Ordinary sessions now receive the complete validated batch through `addItems()`, while model replay continues from the latest valid compaction marker.

The built-in `OpenAIResponsesCompactionSession` also trims input-mode compaction requests to the latest logical window, without trimming stored history or the history exposed to its decision hook. Legacy `RunState` recovery and the public `replaceHistoryWithCompaction()` capability remain supported.

This lets session implementations retain complete transcripts, expose a trimmed `getItems()` view, or explicitly replace physical history. Complete retention can increase storage costs, and `addItems()` retains its existing at-least-once retry semantics.

### Test plan

- Added streaming and non-streaming persistence coverage.
- Added coverage for filtered input, malformed and multiple compaction markers, legacy `RunState` recovery, and built-in Responses compaction input.
- Ran the complete repository verification stack: 5,614 tests passed.

### Issue number

N/A

### DIFF
--- a/.changeset/preserve-inline-compaction-session-transcript.md
+++ b/.changeset/preserve-inline-compaction-session-transcript.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve complete session transcripts across inline compaction
--- a/.changeset/preserve-inline-compaction-session-transcript.md
+++ b/.changeset/preserve-inline-compaction-session-transcript.md
@@ -1,5 +1,6 @@
 ---
 '@openai/agents-core': patch
+'@openai/agents-openai': patch
 ---
 
 fix: preserve complete session transcripts across inline compaction
--- a/packages/agents-core/src/runner/sessionPersistence.ts
+++ b/packages/agents-core/src/runner/sessionPersistence.ts
@@ -444,6 +444,36 @@ function buildRunItemPersistencePlan(
   });
 }
 
+function validateSessionItemsBeforeSessionPreparation(options: {
+  session: Session | undefined;
+  state: RunState<any, any>;
+  runItems: RunItem[];
+  sessionInputItems: AgentInputItem[] | undefined;
+  outputBlocked: boolean;
+}): void {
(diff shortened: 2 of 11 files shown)


seratch:
Thanks for working on this. I do not think we should change the default session persistence semantics here.

The released `Session` contract already supports complete physical transcript retention through `replaceHistoryWithCompaction()`: implementations may append the replacement when a leading compaction item is their logical replay boundary. When that capability is absent, the runner intentionally clears and rebuilds the session with the compacted window.

This PR changes that default for all ordinary sessions, but it does not include a linked issue or a concrete supported scenario showing why the existing capability is insufficient. It also leaves `OpenAIResponsesCompactionSession` with different logical windows for its decision context and input-mode request: the request trims the pre-marker prefix, while the default trigger can continue counting it and repeatedly call compaction. So, I am going to close this PR for now.

will-giraldo:
> Thanks for working on this. I do not think we should change the default session persistence semantics here.
The released Session contract already supports complete physical transcript retention through replaceHistoryWithCompaction(): implementations may append the replacement when a leading compaction item is their logical replay boundary. When that capability is absent, the runner intentionally clears and rebuilds the session with the compacted window.
This PR changes that default for all ordinary sessions, but it does not include a linked issue or a concrete supported scenario showing why the existing capability is insufficient. It also leaves OpenAIResponsesCompactionSession with different logical windows for its decision context and input-mode request: the request trims the pre-marker prefix, while the default trigger can continue counting it and repeatedly call compaction. So, I am going to close this PR for now.

@seratch fair enough, have opened an issue [here](https://github.com/openai/openai-agents-js/issues/1668). 
from what I can tell `replaceHistoryWithCompaction()` does not support complete physical transcript retention due to truncating to latest compaction marker before invocation. This makes it difficult to retain the complete history, unless the intended path is to use `RunResult.newItems` or similar.

PR #1669, opened 2026-08-14
fix(openai): keep Chat Completions reasoning on its assistant message

thegoodengineer:
### Summary

Chat Completions responses that carry reasoning do not round trip. Providers that return `reasoning` on the assistant message get it recorded as a standalone `reasoning` item placed before the assistant `message` item, and on replay the converter turns that one provider message back into two assistant messages, with the reasoning on the one that has no content.

Captured live against OpenRouter and `deepseek/deepseek-r1` on `@openai/agents-openai@0.15.0`. Provider response, `choices[0].message`:

```json
{
  "role": "assistant",
  "content": "The capital of France is **Paris**.",
  "refusal": null,
  "reasoning": "Okay, the user asked a straightforward question: [truncated]",
  "reasoning_details": [ { "type": "reasoning.text", "text": "[truncated]", "format": "unknown", "index": 0 } ]
}
```

Second turn request from the same run:

```json
[
  { "role": "user", "content": "What is the capital of France?" },
  { "role": "assistant", "content": null, "reasoning": "[truncated]" },
  { "role": "assistant", "content": [ { "type": "text", "text": "The capital of France is **Paris**." } ] },
  { "role": "user", "content": "And Germany?" }
]
```

It compounds, because a reasoning item is added per turn. Assistant messages in each outgoing request over five turns:

```
turn | assistant msgs | contentless (content: null, no tool_calls)
  1  |       0        |      0
  2  |       2        |      1
  3  |       4        |      2
  4  |       6        |      3
  5  |       8        |      4
```

So a reasoning model on the Chat Completions path currently replays twice the assistant messages it needs, growing with conversation length, and the caller pays input tokens for the empty ones. The same applies to the streaming path, which produces the same history items.

OpenRouter documents passing reasoning back on the assistant message, via `message.reasoning` or `message.reasoning_details`, to preserve reasoning continuity across calls: https://openrouter.ai/docs/guides/best-practices/reasoning-tokens

This is not an OpenAI Chat Completions field, and the SDK already targets it deliberately on both sides. `hasReasoningContent()` in `openaiChatCompletionsModel.ts` reads `message.reasoning`, the streaming converter accumulates `delta.reasoning`, and `ChatCompletionAssistantMessageWithReasoning` exists in the converter so it can be written back out. The read path and the write path just do not line up.

### Fix

When the pending assistant message was created only by a reasoning item, hand its reasoning to the assistant message that immediately follows instead of flushing it as a separate turn:

```json
[
  { "role": "user", "content": "What is the capital of France?" },
  {
    "role": "assistant",
    "content": [ { "type": "text", "text": "The capital of France is **Paris**." } ],
    "reasoning": "[truncated]"
  },
  { "role": "user", "content": "And Germany?" }
]
```

19 lines in one function. No public API change.

Behavior that is deliberately unchanged, and now covered by tests:

- Reasoning followed by a tool call still shares a single assistant message, which is how tool call turns already replayed.
- A trailing reasoning item with no following assistant message is still emitted on its own.
- Reasoning is never merged into a following user or system message.
- `providerData.reasoning` on the assistant message still wins, since provider data is spread last.
- Only the reasoning item directly before an assistant message is merged into it, so a later assistant message is unaffected.

A reasoning item with empty `rawContent` no longer produces an assistant message with neither content nor reasoning.

Out of scope here, happy to follow up if wanted: `reasoning_details` is not carried onto the assistant message at all, and the nested copy of `reasoning` inside the text content part comes from the existing `providerData` passthrough in `extractAllAssistantContent`.

### Test plan

Six cases in `packages/agents-openai/test/openaiChatCompletionsConverter.test.ts`. Four fail on `main` and pass with the change; two are guards that pass in both states so the unchanged paths stay pinned:

- `keeps reasoning on the assistant message it belongs to`
- `drops an empty reasoning item instead of emitting an empty assistant turn`
- `does not steal reasoning from a previous assistant message`
- `does not accumulate contentless assistant messages across turns`
- `keeps reasoning with the tool call message when a tool call follows` (guard)
- `does not attach reasoning to a following user message` (guard)

Commands run locally on Windows:

- `pnpm build` (0 errors)
- `pnpm test` for `packages/agents-openai`: 21 files, 665 tests passing
- full `pnpm test`: the only failures are pre-existing on this platform and unrelated to this change (the sandbox suites that require POSIX path separators, and the RunState fixture hash check that trips on CRLF checkouts)
- `pnpm lint`
- `pnpm -r build-check`

### Issue number

N/A

### Checks

### DIFF
--- a/.changeset/chat-completions-reasoning-message.md
+++ b/.changeset/chat-completions-reasoning-message.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: keep Chat Completions reasoning on the assistant message it belongs to
--- a/packages/agents-openai/src/openaiChatCompletionsConverter.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsConverter.ts
@@ -533,6 +533,10 @@ export function itemsToMessages(
         ]),
       });
     } else if (item.type === 'unknown') {
+      // An unknown provider item ends the current assistant turn. Flushing here
+      // keeps it in history order and stops reasoning from carrying across it
+      // to an assistant message it does not belong to.
+      flushAssistantMessage();
       result.push({
         ...item.providerData,
       } as any);
--- a/packages/agents-openai/src/openaiChatCompletionsConverter.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsConverter.ts
@@ -305,9 +305,28 @@ export function itemsToMessages(
     }
     return currentAssistantMsg;
   };
+  // A reasoning item is emitted right before the assistant message it belongs
+  // to, so hand its reasoning back to that message instead of flushing an
+  // assistant turn that carries no content and no tool calls.
+  const takeReasoningOnlyAssistantMessage = () => {
+    const pending: ChatCompletionAssistantMessageWithReasoning | null =
+      currentAssistantMsg;
+    if (
+      !pending ||
+      pending.content !== null ||
+      (pending.tool_calls && pending.tool_calls.length > 0) ||
+      !('reasoning' in pending)
+    ) {
+      return undefined;
+    }
+    currentAssistantMsg = null;
+    return pending;
+  };
   for (const item of items) {
     if (isMessageItem(item)) {
       const { content, role, providerData } = item;
+      const pendingReasoningMsg =
+        role === 'assistant' ? takeReasoningOnlyAssistantMessage() : undefined;
       flushAssistantMessage();
       if (role === 'assistant') {
         const phase = item.phase ?? providerData?.phase;
@@ -325,9 +344,12 @@ export function itemsToMessages(
           }
           logger.warn(`${message} The phase will be dropped.`);
         }
-        const assistant: ChatCompletionAssistantMessageParam = {
+        const assistant: ChatCompletionAssistantMessageWithReasoning = {
           role: 'assistant',
           content: extractAllAssistantContent(content),
+          ...(typeof pendingReasoningMsg?.reasoning === 'string'
+            ? { reasoning: pendingReasoningMsg.reasoning }
+            : {}),
           ...getProviderDataWithoutReservedKeys(providerData, [
             'role',
             'content',
--- a/packages/agents-openai/test/openaiChatCompletionsConverter.test.ts
+++ b/packages/agents-openai/test/openaiChatCompletionsConverter.test.ts
@@ -957,6 +957,141 @@ describe('itemsToMessages', () => {
     ]);
   });
 
+  test('keeps reasoning on the assistant message it belongs to', () => {
+    const items: protocol.ModelItem[] = [
+      {
+        type: 'reasoning',
+        content: [],
+        rawContent: [{ type: 'reasoning_text', text: 'why' }],
+      } as protocol.ReasoningItem,
+      {
+        id: 'msg_1',
+        type: 'message',
+        role: 'assistant',
+        status: 'completed',
+        content: [{ type: 'output_text', text: 'hello' }],
+      } as protocol.AssistantMessageItem,
+    ];
+    const msgs = itemsToMessages(items);
+    expect(msgs).toEqual([
+      {
+        role: 'assistant',
+        content: [{ type: 'text', text: 'hello' }],
+        reasoning: 'why',
+      },
+    ]);
+  });
+
+  test('drops an empty reasoning item instead of emitting an empty assistant turn', () => {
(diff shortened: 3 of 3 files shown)


seratch:
The converter behavior is internally plausible, but we need evidence that this exact Chat Completions-compatible path occurs with a real provider before adding permanent replay behavior.

Could you share the provider and model, the raw first-turn assistant message containing `reasoning`, the exact second-turn request produced by the current SDK, and the provider error or incorrect observable result? Please also confirm that the same reproduction fails on the latest released SDK version. OpenAI’s Chat Completions API does not define an assistant `reasoning` message field, and the existing Anthropic reports used a different AI SDK adapter path, so those do not establish the need for this converter change.

With a real-provider reproduction, we can reconsider this as a narrow round-trip fix.

thegoodengineer:
Fair ask. One correction first: I wrote that providers reject the payload. I had no basis for that and I withdraw it. No provider error is involved. Here is the live capture instead, and I have rewritten the description around it.

**Provider:** OpenRouter, `deepseek/deepseek-r1`, on `@openai/agents-openai@0.15.0`.

**Raw first-turn assistant message, off the wire:**

```json
{
  "role": "assistant",
  "content": "The capital of France is **Paris**.",
  "refusal": null,
  "reasoning": "Okay, the user asked a straightforward question: [truncated]",
  "reasoning_details": [ { "type": "reasoning.text", "text": "[truncated]", "format": "unknown", "index": 0 } ]
}
```

**Second turn request from the same run:**

```json
[
  { "role": "user", "content": "What is the capital of France?" },
  { "role": "assistant", "content": null, "reasoning": "[truncated]" },
  { "role": "assistant", "content": [ { "type": "text", "text": "The capital of France is **Paris**." } ] },
  { "role": "user", "content": "And Germany?" }
]
```

**The observable result.** No error, OpenRouter answered normally both turns. What is wrong is that one provider message replays as two, the reasoning sits on the message with `content: null`, and it compounds, since a reasoning item is added per turn. Assistant messages per outgoing request over five turns:

```
turn | assistant msgs | contentless (content: null, no tool_calls)
  1  |       0        |      0
  2  |       2        |      1
  3  |       4        |      2
  4  |       6        |      3
  5  |       8        |      4
```

With the patch: 0, 1, 2, 3, 4 and never a contentless one. Same on the streaming path, which produces the same history items. So today a reasoning model on Chat Completions replays twice the assistant messages it needs, growing with conversation length, and the caller pays input tokens on the empty half.

OpenRouter documents passing reasoning back on the assistant message, `message.reasoning` or `message.reasoning_details`, for reasoning continuity across calls: https://openrouter.ai/docs/guides/best-practices/reasoning-tokens

You are right that this is not an OpenAI field. The SDK already targets it deliberately on both sides though. `hasReasoningContent()` reads `message.reasoning`, the streaming converter accumulates `delta.reasoning`, and `ChatCompletionAssistantMessageWithReasoning` exists so it can be written back. The read and write paths just do not line up, which is all this PR changes: 19 lines in one function, no public API change, six tests, four of which fail on `main`.

Repro if you want to rerun it, needs an OpenRouter key and nothing else:

```js
import OpenAI from 'openai';
import { Agent, run, MemorySession, setTracingDisabled, OpenAIChatCompletionsModel } from '@openai/agents';

setTracingDisabled(true);

const client = new OpenAI({
  apiKey: process.env.OPENROUTER_API_KEY,
  baseURL: 'https://openrouter.ai/api/v1',
});

const requests = [];
const responses = [];
const create = client.chat.completions.create.bind(client.chat.completions);
client.chat.completions.create = async (body, options) => {
  requests.push(structuredClone(body));
  const response = await create(body, options);
  responses.push(structuredClone(response));
  return response;
};

const agent = new Agent({
  name: 'Assistant',
  instructions: 'Answer in one short sentence.',
  model: new OpenAIChatCompletionsModel(client, 'deepseek/deepseek-r1'),
  modelSettings: { providerData: { include_reasoning: true }, maxTokens: 1200 },
});
const session = new MemorySession();

await run(agent, 'What is the capital of France?', { session });
await run(agent, 'And Germany?', { session });

console.log(JSON.stringify(responses[0].choices[0].message, null, 2));
console.log(JSON.stringify(requests[1].messages, null, 2));
```

I kept the change narrow on purpose since that is what you asked for. `reasoning_details` is visible in the trace above and is not carried onto the assistant message at all today. Say the word if you want that in this PR and I will add it, otherwise I will leave it as a separate follow up.

seratch (approved):
Confirmed this works better with various model providers on openrouter

PR #1677, opened 2026-08-15
feat(extensions): expose Twilio start listener

### LINKED ISSUE #18: Twilio transport: Allow passing start data to context
### Describe the feature

The twilio start event has important information like the `callSid` and `customParameters`. It would be great to be able to have dynamic instructions for the agent based on `customParameters`. 
https://www.twilio.com/docs/voice/twiml/stream#custom-parameters

### Current behavior
The agent/session is instantiated prior to the twilio start event so you can't use the `twilio_message` event to set the context. 

### Example of desired behavior
```ts
import { TwilioRealtimeTransportLayer } from '@openai/agents-extensions';
import { RealtimeAgent, RealtimeSession } from '@openai/agents/realtime';

const agent = new RealtimeAgent({
  name: 'My Agent',
  instructions: ({ context: ctx }) => `Thank you for calling the ${ctx.customParameters.name} store!`
});

const twilioTransport = new TwilioRealtimeTransportLayer({
  twilioWebSocket: websoketConnection,
});

// Not sure what the interface should look like, but it'd be nice if the twilio/openai connect weren't so tightly coupled
const { start } = await twilioTransport.twilioConnect();

const session = new RealtimeSession(agent, {
  // set your own transport
  transport: twilioTransport,
  context: {
    callSid: start.callSid,
    customParameters: start.customParameters
  }
});

await session.connect({ apiKey: 'your-openai-api-key' });
```

another option would be to store the whole start data on the transport and include the transport in the RunContext
```ts
const agent = new RealtimeAgent({
  name: 'My Agent',
  instructions: ({ transport }) => `Thank you for calling the ${transport.start.customParameters.name} store!`
});
```

sylvesterkaczmarek:
Fixes #18.

Adds `TwilioRealtimeTransportLayer.listen()` so applications can await Twilio's Media Streams `start` event before creating their `RealtimeSession`. This exposes `callSid`, `streamSid`, and `customParameters` early enough to construct application context without adding Twilio-specific APIs to `RealtimeSession`.

The transport caches the start payload, reuses a pending listener promise, and rejects if the Twilio websocket closes or errors before `start` arrives.

Focused formatter, ESLint, and `@openai/agents-extensions` tests pass.


seratch:
We won't apply a fix for the issue for now.

PR #1686, opened 2026-08-15
fix(openai): keep Chat Completions refusals that arrive with content

thegoodengineer:
### Summary

A Chat Completions message can carry `content` and `refusal` at the same time. The non-streaming path built the assistant item with `if` / `else if`, so the text branch won and the refusal was thrown away:

```ts
if (hasContent) {
  output.push({ ...content: [{ type: 'output_text', ... }] });
} else if (message.refusal) {
  output.push({ ...content: [{ type: 'refusal', ... }] });
}
```

The refusal never reaches `result.output`, the run history, or the caller. It is not degraded, it is gone.

The streaming converter does not do this. It collects both parts into one message, and the suite already pins that shape. `emits protocol events for streamed chat completions` in `openaiChatCompletionsStreaming.test.ts` streams `content: 'hello'` plus `refusal: 'nope'` and asserts:

```ts
content: [
  { type: 'output_text', text: 'hello', providerData: { annotations: [] } },
  { type: 'refusal', refusal: 'nope' },
],
```

So the same provider message produced different protocol items depending on whether the caller streamed. The non-streaming behavior was untested, which is why the divergence went unnoticed.

Empty content makes it worse, and this is the case that actually breaks a run. `hasContent` is true for `''` when there are no tool calls, and the comment directly above it notes that Azure returns an empty string where OpenAI returns `null`. Streaming ignores empty content deltas, so it emits a refusal-only message and `turnResolution` raises `ModelRefusalError`. Non-streaming emitted an empty text part instead, and since the runner treats any text part as a final output, the refusal was swallowed and the caller got a blank successful result with no signal that the model refused.

### Fix

Collect both parts, mirroring the streaming converter, and treat empty content as absent when a refusal is present:

```ts
const hasTextContent =
  hasContent && !(message.refusal && message.content === '');

const messageContent: protocol.AssistantContent[] = [];
if (hasTextContent) { messageContent.push({ type: 'output_text', ... }); }
if (message.refusal) { messageContent.push({ type: 'refusal', ... }); }
if (messageContent.length > 0) { output.push({ ...content: messageContent }); }
else if (message.audio) { ... }
```

Resulting behavior, now matching streaming on every case:

| provider message | before | after |
| --- | --- | --- |
| `content: 'text'`, no refusal | text | unchanged |
| `content: null`, refusal | refusal | unchanged |
| `content: ''`, no refusal | empty text | unchanged |
| `content: 'text'`, refusal | text, refusal dropped | text and refusal |
| `content: ''`, refusal | empty text, refusal dropped, blank success | refusal only, `ModelRefusalError` raised |

The audio branch still applies only when there is neither text nor refusal, and the synthesized content-filter path sets `content: null`, so both are untouched. No existing test changed.

### Test plan

Four cases added to `packages/agents-openai/test/openaiChatCompletionsModel.test.ts`. Three fail on `main`, and `keeps empty content when there is no refusal` is a guard that passes in both states:

- `keeps a refusal that arrives alongside content`
- `drops empty content so a refusal stays the only output`
- `surfaces a refusal to the runner when content is empty` (asserts the `ModelRefusalError`, not just the item shape)
- `keeps empty content when there is no refusal` (guard)

Commands run locally on Windows:

- `pnpm test` for `packages/agents-openai`: 21 files, 671 tests passing, no existing test modified
- `tsc --noEmit -p packages/agents-openai/tsconfig.json` (0 errors)
- `pnpm lint`
- prettier clean on both changed files

### Issue number

N/A

### Checks

### DIFF
--- a/.changeset/chat-completions-refusal-with-content.md
+++ b/.changeset/chat-completions-refusal-with-content.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: keep Chat Completions refusals that arrive alongside content
--- a/packages/agents-openai/src/openaiChatCompletionsModel.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsModel.ts
@@ -227,34 +227,33 @@ export class OpenAIChatCompletionsModel implements Model {
         // Azure OpenAI returns empty string instead of null for tool calls, causing parser rejection
         !(message.tool_calls && message.content === '');
 
+      // A message can carry text and a refusal at once, so collect both instead
+      // of letting the text branch drop the refusal. This matches how the
+      // streaming converter assembles assistant content.
+      const messageContent: protocol.AssistantContent[] = [];
       if (hasContent) {
         const { content, ...rest } = message;
-        output.push({
-          id: response.id,
-          type: 'message',
-          role: 'assistant',
-          content: [
-            {
-              type: 'output_text',
-              text: content || '',
-              providerData: rest,
-            },
-          ],
-          status: 'completed',
+        messageContent.push({
+          type: 'output_text',
+          text: content || '',
+          providerData: rest,
         });
-      } else if (message.refusal) {
+      }
+      if (message.refusal) {
         const { refusal, ...rest } = message;
+        messageContent.push({
+          type: 'refusal',
+          refusal: refusal || '',
+          providerData: rest,
+        });
+      }
+
+      if (messageContent.length > 0) {
         output.push({
           id: response.id,
           type: 'message',
           role: 'assistant',
-          content: [
-            {
-              type: 'refusal',
-              refusal: refusal || '',
-              providerData: rest,
-            },
-          ],
+          content: messageContent,
           status: 'completed',
         });
       } else if (message.audio) {
--- a/packages/agents-openai/src/openaiChatCompletionsModel.ts
+++ b/packages/agents-openai/src/openaiChatCompletionsModel.ts
@@ -230,8 +230,15 @@ export class OpenAIChatCompletionsModel implements Model {
       // A message can carry text and a refusal at once, so collect both instead
       // of letting the text branch drop the refusal. This matches how the
       // streaming converter assembles assistant content.
+      //
+      // An empty string is not an answer. Keeping it as a text part next to a
+      // refusal would hide the refusal from the runner, which treats any text
+      // part as a final output. The streaming converter skips empty content
+      // deltas for the same reason.
+      const hasTextContent =
+        hasContent && !(message.refusal && message.content === '');
       const messageContent: protocol.AssistantContent[] = [];
-      if (hasContent) {
+      if (hasTextContent) {
         const { content, ...rest } = message;
         messageContent.push({
           type: 'output_text',
--- a/packages/agents-openai/test/openaiChatCompletionsModel.test.ts
+++ b/packages/agents-openai/test/openaiChatCompletionsModel.test.ts
@@ -1020,6 +1020,68 @@ describe('OpenAIChatCompletionsModel', () => {
     ]);
   });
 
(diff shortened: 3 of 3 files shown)


seratch (changes requested):
Can you share the repro steps with real providers?

thegoodengineer:
Honest answer up front: I do not have a real-provider transcript for this one, and I do not think I can produce one. Here is what I checked, and the evidence I can offer instead.

### Which providers populate `refusal`

- **OpenAI**: populates it on structured-output refusals, and sends `content: null` alongside. That combination already worked before this PR, so it does not exercise the bug.
- **Third-party OpenAI-compatible providers**: I checked OpenRouter with DeepSeek R1 and it does not populate `refusal` at all. Most third-party models do not.
- **Azure OpenAI**: the case this PR is about. The comment directly above `hasContent` in this file already states that Azure returns an empty string where OpenAI returns `null`. Combined with a structured-output refusal that yields `content: '' , refusal: '...'`. I do not have an Azure deployment, so I could not capture it, and I searched this repo's issues for the original Azure report without finding one, so that comment is the only source I have.

If you have an Azure deployment, the repro would be: point `OpenAIChatCompletionsModel` at it, set an `outputType` so structured outputs are in play, send a prompt the model refuses, and inspect `choices[0].message`. On `main`, `result.output` comes back as a single empty `output_text` and the refusal is gone. I would rather say I have not run that than imply I have.

### What does not need a provider

**1. The Python SDK already does exactly this.** `openai-agents-python`, `src/agents/models/chatcmpl_converter.py`, lines 220 to 237:

```python
if message.content:                        # truthy, so '' is skipped
    message_item.content.append(ResponseOutputText(...))
if message.refusal:                        # separate if, not elif
    message_item.content.append(ResponseOutputRefusal(...))
if message_item.content:
    items.append(message_item)
```

All three behaviors this PR adds are already there: both parts collected instead of `if` / `else`, empty content treated as absent, and the message emitted only when it holds something. This PR makes the JS non-streaming path match that. So the question of whether a provider sends this shape looks like it was already settled on the Python side.

**2. This repo's own streaming path disagrees with its non-streaming path.** `emits protocol events for streamed chat completions` in `openaiChatCompletionsStreaming.test.ts` streams `content: 'hello'` with `refusal: 'nope'` and asserts a single message holding both parts. The same message through `getResponse` dropped the refusal. For `content: ''` the divergence is behavioral, not cosmetic: streaming skips empty content deltas so `ModelRefusalError` is raised, while non-streaming emitted an empty text part, which the runner reads as a final answer, so the refusal is swallowed and the caller gets a blank success.

### Deterministic repro

Three of the four tests in this PR fail on `main`, including `surfaces a refusal to the runner when content is empty`, which asserts the `ModelRefusalError` rather than the item shape. `pnpm test packages/agents-openai/test/openaiChatCompletionsModel.test.ts` against `main` shows it.

If a live provider trace is a hard requirement here, I understand, and I am happy to close this and reopen if an Azure report comes in.

seratch:
Thanks for the detailed and candid investigation. The deterministic tests establish that the converter can receive a manually constructed `content: ''` plus `refusal` shape and that the current implementation would swallow the refusal. They do not establish that Azure or another supported provider actually emits that combination.

The existing Azure comment came from #203 and specifically covers `content: ''` alongside `tool_calls`; it is not evidence that structured-output refusals use the same shape. The streaming test and Python implementation are useful consistency and design signals, but they also only show how a structurally possible response is handled, not realistic provider behavior or user impact.

Given that we do not have a provider transcript, a concrete user report, or a documented provider contract for the exact response shape, I do not think we have enough evidence to add and maintain this behavior. Let's close this for now. If we receive an Azure response showing non-empty `refusal` alongside `content: ''`, or another concrete reproduction from a supported provider, we can reopen this and evaluate a focused fix.

PR #1687, opened 2026-08-15
fix(realtime): restore function calls through history updates

### LINKED ISSUE #645: Add support for adding function call items to conversation history via `updateHistory`

seratch:
## Description

This pull request resolves #645.

Previously, `RealtimeSession.updateHistory()` ignored manually added `function_call` items. This prevented applications from restoring conversation history that included tool calls and their recorded outputs.

This change:

- Allows `updateHistory()` to restore a function call without an output.
- Restores completed function calls and their outputs as an ordered `function_call` / `function_call_output` pair.
- Adds optional `callId` and `outputItemId` fields to `RealtimeToolCallItem` so the SDK can preserve API identities.
- Assigns client-generated output item IDs that satisfy the Realtime API's 32-character item ID limit.
- Deletes a restored output item before deleting its associated function call.
- Preserves the existing combined SDK representation of a function call and its output.
- Rejects unsupported in-place function-call edits, unsafe removals, contradictory status/output combinations, and duplicate identities before sending any conversation mutations.
- Keeps the existing optimistic local-history update behavior without introducing connection-scoped replay state.

Directly editing an existing function-call history item remains unsupported. Applications can remove the item in one `updateHistory()` call and add its replacement in a subsequent call.

The restored call/output flow was also verified against the live Realtime API: the model consumed the restored tool output successfully, and the output and call items were subsequently deleted without API errors.

### DIFF
--- a/.changeset/realtime-function-call-history.md
+++ b/.changeset/realtime-function-call-history.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: restore function calls through Realtime history updates (#645)
--- a/packages/agents-realtime/src/items.ts
+++ b/packages/agents-realtime/src/items.ts
@@ -51,6 +51,8 @@ export const realtimeMessageItemSchema = z.discriminatedUnion('role', [
 export const realtimeToolCallItem = z.object({
   itemId: z.string(),
   previousItemId: z.string().nullable().optional(),
+  callId: z.string().optional(),
+  outputItemId: z.string().optional(),
   type: z.literal('function_call'),
   status: z.enum(['in_progress', 'completed', 'incomplete']),
   arguments: z.string(),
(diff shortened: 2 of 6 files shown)


seratch:
The current head still has two material consistency issues: an uncorrelated invalid_tool_call_id error can roll back a different tracked output, and a batch can delete a predecessor before sending a create that still references it. Continuing to patch this state machine is not the best maintenance path.

PR #1689, opened 2026-08-16
fix(sandbox): trace effective sandbox exec workdirs

sylvesterkaczmarek:
### Summary

Record the effective sandbox working directory in `sandbox.exec` trace and event data.

With a run-scoped sandbox cwd, `exec_command` already anchors an omitted or relative `workdir` before calling the sandbox session. The span data was still populated from the raw model argument, so an execution under `tasks/a` could be traced with `workdir` missing, and `reports` could be traced even though the session executed in `tasks/a/reports`.

This change computes the effective workdir once and uses that same value for both the trace payload and `session.execCommand`.

### Test plan

Added event-level regression coverage for an omitted workdir and an explicit relative workdir under `SandboxWorkspaceScope.fromCwd('tasks/a')`. The test verifies that emitted audit events and sandbox execution receive identical effective paths.

### Issue number

None. Found while auditing run-scoped sandbox trace provenance.

### DIFF
--- a/.changeset/calm-traces-workdir.md
+++ b/.changeset/calm-traces-workdir.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix sandbox exec traces to record the effective run-scoped working directory
--- a/.changeset/calm-traces-workdir.md
+++ b/.changeset/calm-traces-workdir.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
-fix sandbox exec traces to record the effective run-scoped working directory
+fix: record the effective workdir in sandbox exec traces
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:
@codex review

seratch (approved):

seratch (changes requested):
Can you resolve the CI errors?

seratch:
Thanks for working through this. The effective `sandbox.exec` workdir fix is valid, but #1697 now provides a newer-base destination for the same run-scoped trace-provenance contract and its corresponding `view_image` case.

Please carry this PR's implementation and omitted/relative workdir regression coverage into #1697.

PR #1691, opened 2026-08-16
fix(sandbox): validate dynamic compaction ratios

sylvesterkaczmarek:
### Summary

Validate `DynamicCompactionPolicy.thresholdRatio` before it is used to derive sandbox compaction thresholds.

The JavaScript SDK currently accepts negative values, values above 1, `NaN`, and infinities. A ratio above 1 can place the compaction threshold beyond the model context window, while negative or non-finite values can produce invalid request parameters.

The equivalent Python sandbox policy constrains its dynamic threshold ratio to the inclusive 0–1 range. This change applies the same invariant in JavaScript and raises `UserError` at configuration time for non-finite or out-of-range values.

### Test plan

Added focused coverage for:

- negative and greater-than-one ratios
- `NaN` and positive/negative infinity
- valid boundary ratios of 0 and 1

### Issue number

None. Found while comparing sandbox compaction assurance invariants across the Python and JavaScript SDKs.

### DIFF
--- a/.changeset/safe-compaction-ratios.md
+++ b/.changeset/safe-compaction-ratios.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+validate dynamic sandbox compaction threshold ratios before use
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:
@codex review

PR #1692, opened 2026-08-16
fix: preserve sandbox failures during error serialization

sylvesterkaczmarek:
### Summary

Preserve the original sandbox operation failure when telemetry error serialization encounters an unserializable thrown value, hostile optional metadata accessor, or hostile error-classification proxy.

`withSandboxSpan` serializes failures before recording the error event and then rethrows the original value. JavaScript permits arbitrary thrown values, proxies, and property getters that can themselves throw, so telemetry serialization could replace the original sandbox failure and prevent the intended error event from being emitted.

This change makes error serialization fail-safe. Non-`Error` values that cannot be stringified or classified receive a generic event message. For `Error` objects, safe fields such as `name` and `message` are preserved independently, while optional `code` or `retryable` metadata that cannot be read is omitted. `withSandboxSpan` continues to rethrow the exact original operation failure.

### Test plan

Added regression coverage for:

- a non-`Error` thrown value whose `toString()` throws
- an `Error` whose optional `code` getter throws
- a proxy whose prototype lookup makes `instanceof Error` throw
- preservation of the exact original thrown value in all cases
- preservation of safe `Error` name/message fields when optional metadata access fails

### Issue number

None. Found while auditing sandbox error and trace integrity.

### DIFF
--- a/.changeset/safe-sandbox-error-serialization.md
+++ b/.changeset/safe-sandbox-error-serialization.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve sandbox failures when error serialization fails
--- a/packages/agents-core/src/sandbox/events.ts
+++ b/packages/agents-core/src/sandbox/events.ts
@@ -133,20 +133,26 @@ export function createChainedSandboxEventSink(
 }
 
 export function serializeSandboxEventError(error: unknown): SandboxEventError {
-  if (error instanceof Error) {
-    const code = readErrorCode(error);
-    const retryable = readErrorRetryability(error);
+  try {
+    if (error instanceof Error) {
+      const code = readErrorCode(error);
+      const retryable = readErrorRetryability(error);
+      return {
+        name: error.name,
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
@codex review

sylvesterkaczmarek:
@codex review

seratch:
Thanks for the careful patch. The failure mode is technically valid, but the current evidence only demonstrates deliberately hostile JavaScript values: a throwing `toString()`, throwing metadata getters, and a proxy that throws during error classification.

There is no linked user report, provider example, or ordinary supported sandbox path showing that these shapes occur in practice. Normal SDK and provider errors already preserve the original failure and structured diagnostics. Given the additional permanent serialization behavior and tests required for these pathological inputs, we do not think this is worth maintaining without evidence of realistic user impact.

I am going to close this PR. If a real provider or custom sandbox implementation is observed producing such an error, please open an issue with the concrete error shape and execution path; we can then consider a narrower fail-safe at the sandbox span boundary.

PR #1695, opened 2026-08-16
feat: allow Docker sandboxes to disable networking

seratch:
This pull request adds opt-in network isolation for Docker sandboxes through `networkMode: "none"`.

It rejects incompatible exposed-port configuration before side effects, preserves the setting across session serialization and replacement, and rebuilds RunState sessions from trusted current configuration. Preserved containers are reused only after Docker inspection confirms that they remain isolated.

The RunState schema and sandbox-state envelope are advanced to prevent older SDK versions from silently dropping the isolation requirement during replacement.

### DIFF
--- a/.changeset/calm-dockers-rest.md
+++ b/.changeset/calm-dockers-rest.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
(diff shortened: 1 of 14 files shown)


seratch:
@codex review

PR #1697, opened 2026-08-16
fix(sandbox): trace effective run-scoped sandbox paths

sylvesterkaczmarek:
### Summary

Record the effective run-scoped sandbox paths in tracing and audit event data for both `view_image` and `exec_command`.

With a run-scoped sandbox cwd, the model-provided path/workdir can differ from the value actually passed to the sandbox session. Previously, tracing could record the raw relative value even though execution used the anchored path.

This change computes each effective path once and uses that same value for both the trace payload and sandbox call:

- `view_image` records and reads the same anchored path
- `exec_command` records and executes in the same effective workdir, including omitted and relative workdirs

User-facing `view_image` error messages continue to use the model-provided path.

This incorporates the related `exec_command` fix and regression coverage from #1689 into this PR as requested in review.

### Test plan

Added event-level regression coverage under `SandboxWorkspaceScope.fromCwd('tasks/a')` verifying:

- `view_image` reads `tasks/a/plot.png` and both start/end events record that path
- omitted `exec_command` workdir resolves to `tasks/a`
- relative `exec_command` workdir `reports` resolves to `tasks/a/reports`
- sandbox execution and emitted audit events receive identical effective paths

A single patch changeset for `@openai/agents-core` is included.

### Issue number

None. Found while auditing run-scoped sandbox trace provenance.

### DIFF
--- a/.changeset/effective-view-image-traces.md
+++ b/.changeset/effective-view-image-traces.md
@@ -2,4 +2,4 @@
 '@openai/agents-core': patch
 ---
 
-fix(sandbox): trace the effective view_image path
+fix(sandbox): trace effective run-scoped sandbox paths
--- a/.changeset/effective-view-image-traces.md
+++ b/.changeset/effective-view-image-traces.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix(sandbox): trace the effective view_image path
--- a/packages/agents-core/src/sandbox/capabilities/filesystem.ts
+++ b/packages/agents-core/src/sandbox/capabilities/filesystem.ts
@@ -548,16 +548,17 @@ class FilesystemCapability extends Capability {
           'Filesystem sandbox sessions must provide viewImage().',
         );
       }
+      const effectivePath = this._workspaceScope?.anchor(path) ?? path;
       try {
         return await withSandboxSpan(
           'sandbox.view_image',
           {
-            path,
+            path: effectivePath,
             run_as: this._runAs,
           },
           async () =>
             await session.viewImage!({
-              path: this._workspaceScope?.anchor(path) ?? path,
+              path: effectivePath,
               runAs: this._runAs,
             } satisfies ViewImageArgs),
           this.tracingParent(details),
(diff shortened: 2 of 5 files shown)


sylvesterkaczmarek:
@codex review

seratch (changes requested):
Thanks for the contribution. The `view_image` trace mismatch is real, and computing the effective path once for both the event/span payload and `session.viewImage()` is the right implementation.

However, #1689 fixes the same run-scoped trace-provenance invariant for `exec_command`: the raw `workdir` can differ from the directory passed to `session.execCommand()`. We should keep this contract in one PR rather than merge separate implementations, tests, and changesets.

Please use #1697 as the destination: incorporate the `effectiveWorkdir` change and omitted/relative workdir regression coverage from #1689, and keep one `@openai/agents-core` changeset.

sylvesterkaczmarek:
Addressed the review request in ba9f810: #1697 now includes the `effectiveWorkdir` change and omitted/relative workdir regression coverage from #1689, while keeping a single `@openai/agents-core` changeset. I also updated the PR title and description to reflect the combined scope. Ready for re-review.

seratch:
The test you've added fails to compile:

```
Run pnpm -r build-check
Scope: 29 of 30 workspace projects
examples/ai-sdk-v1 build-check$ tsc --noEmit
packages/agents-core build-check$ tsc --noEmit -p ./tsconfig.test.json
examples/ai-sdk-v1 build-check: Done
Error: packages/agents-core build-check: test/sandboxTracePaths.test.ts(45,34): error TS18048: 'event.data' is possibly 'undefined'.
packages/agents-core build-check: Failed
/home/runner/work/openai-agents-js/openai-agents-js/packages/agents-core:
[ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL] @openai/agents-core@0.16.0 build-check: `tsc --noEmit -p ./tsconfig.test.json`
Exit status 2
```

sylvesterkaczmarek:
Fixed the fresh CI typecheck failure in c577f70. The only failing diagnostic was `event.data` possibly being undefined in the new regression test; the assertion now uses optional access. Build, lint, coverage, and the Windows sandbox-path job had already passed on the previous run.

seratch (approved):

sylvesterkaczmarek:
@codex review

seratch:
@sylvesterkaczmarek Please avoid mentioning Codex unless the automatic review trigger fails for some reason. The Agents SDK repositories normally trigger a review automatically on every push.

PR #1698, opened 2026-08-16
fix(realtime): preserve omitted noise reduction

sylvesterkaczmarek:
### Summary

Preserve omission of `audio.input.noiseReduction` when converting the GA Realtime session config.

`toNewSessionConfig` currently maps an omitted `noiseReduction` field to `null` whenever an input audio object exists. This means a caller that only sets `audio.input.format` silently sends an explicit noise-reduction disable instead of leaving that setting unspecified.

The new-config path is documented as a shallow normalization of the supplied GA config. This change keeps omitted noise reduction omitted while continuing to preserve explicit `null` and explicit noise-reduction configurations. The deprecated config conversion remains unchanged.

### Test plan

Added focused coverage showing that:

- setting only `audio.input.format` leaves `noiseReduction` undefined and absent after JSON serialization
- explicit `noiseReduction: null` remains `null`
- an explicit `{ type: 'near_field' }` configuration is preserved

A patch changeset for `@openai/agents-realtime` is included.

### Issue number

None. Found while auditing GA Realtime session config normalization.

### DIFF
--- a/.changeset/preserve-realtime-noise-reduction.md
+++ b/.changeset/preserve-realtime-noise-reduction.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix(realtime): preserve omitted input noise reduction
--- a/packages/agents-realtime/src/clientMessages.ts
+++ b/packages/agents-realtime/src/clientMessages.ts
@@ -217,7 +217,7 @@ export function toNewSessionConfig(
     const inputConfig = audioInput
       ? {
           format: normalizeAudioFormat(audioInput?.format),
-          noiseReduction: audioInput?.noiseReduction ?? null,
+          noiseReduction: audioInput.noiseReduction,
           transcription: audioInput?.transcription,
           turnDetection: audioInput?.turnDetection,
         }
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
@codex review

seratch:
Thanks for investigating this. The distinction between an omitted Realtime session field and an explicit `null` is meaningful, but this patch does not change the payload sent by the SDK. After `toNewSessionConfig()` returns `undefined`, `_getMergedSessionConfig()` substitutes `DEFAULT_OPENAI_REALTIME_SESSION_CONFIG.audio.input.noiseReduction`, which is `null`, so `noise_reduction: null` is still emitted by the WebSocket, WebRTC, SIP, and session-update paths.

I am going to close this PR for now. If there is a concrete supported scenario where preserving omission changes the observable service behavior, please provide that reproduction. We can then reconsider a fix at the payload-merging boundary with a regression test against `buildSessionPayload()` or the emitted `session.update` event.

sylvesterkaczmarek:
Addressed the payload-boundary finding. Format-only input config now keeps `noise_reduction` omitted in `buildSessionPayload()` and the emitted JSON shape, while an empty config still gets the SDK default `null` and explicit `null` remains preserved. Added final-payload regression coverage. @codex review

PR #1699, opened 2026-08-16
fix(sandbox): make provider error inspection fail-safe

sylvesterkaczmarek:
### Summary

Keep remote sandbox provider failures inside the SDK's provider-error boundary even when the thrown value cannot be safely inspected.

The shared provider helpers inspect arbitrary SDK/provider failures for message, status, retryability, response payload, and nested cause data. JavaScript permits proxies and property accessors that can themselves throw. In those cases, `instanceof`, property reads, or string conversion could replace the original provider failure with a JavaScript inspection error before `withProviderError` produced its intended `SandboxProviderError`.

This change makes the exported provider diagnostic helpers fail-safe. Normal provider errors retain their existing rich metadata. Uninspectable failures use a generic message, empty structured details, and unknown retryability. UserError passthrough is preserved when classification is safe, and resume error handling uses the same guarded boundary.

### Test plan

Added regression coverage showing that a hostile provider proxy whose property and prototype access throw:

- produces a safe generic provider message
- produces empty structured details and unknown retryability
- is wrapped by `withProviderError` as `SandboxProviderError`
- remains inside the provider error boundary during resume handling

A patch changeset for `@openai/agents-extensions` is included.

### Issue number

None. Found while auditing remote sandbox provider failure integrity.

### DIFF
--- a/.changeset/failsafe-provider-errors.md
+++ b/.changeset/failsafe-provider-errors.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+fix(sandbox): make provider error inspection fail-safe
--- a/packages/agents-extensions/src/sandbox/shared/session.ts
+++ b/packages/agents-extensions/src/sandbox/shared/session.ts
@@ -10,6 +10,9 @@ import { isRecord } from './typeGuards';
 
 export { withSandboxSpan };
 
+const UNINSPECTABLE_PROVIDER_ERROR =
+  'Provider error could not be inspected safely.';
+
 export async function closeRemoteSessionOnManifestError(
   providerName: string,
   session: { close(): Promise<void> },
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
@codex review

seratch:
Thanks for investigating this. I confirmed that a deliberately hostile proxy can make the current diagnostic inspection throw a secondary JavaScript error. However, the PR does not show that any supported sandbox provider returns this kind of value, and the regression tests only establish synthetic reachability rather than a realistic provider or user impact.

The broader arbitrary-value guarantee is also incomplete because a successful `message` read is still assumed to produce a string. I am going to close this PR for now. If a supported provider can reproduce an inspection-unsafe error in practice, please share the exact provider, operation, thrown value shape, and observable consequence so we can revisit the narrowest appropriate fix.

sylvesterkaczmarek:
Addressed the malformed-message finding. Provider messages are now safely coerced before leaving the diagnostic boundary; Symbols remain diagnosable and unstringifiable objects retain the generic fallback. Added regression coverage for both cases. @codex review

PR #1700, opened 2026-08-16
fix(realtime): normalize unsupported PCM rates

sylvesterkaczmarek:
### Summary

Normalize Realtime PCM audio format objects to the only sample rate supported by the OpenAI Realtime API.

The current JavaScript converter only replaces non-finite PCM rates. Finite values such as 16000, 44100, or 48000 pass through unchanged even though the Realtime API's PCM format declares the rate as always 24000. This can allow an invalid session config to reach the API.

This change normalizes every PCM rate other than 24000 to 24000. The supported rate is preserved unchanged, and PCMU/PCMA formats are unaffected. This also matches the existing Python Agents SDK behavior for unsupported PCM rates.

### Test plan

Added focused coverage for:

- finite unsupported PCM rates including 16000, 44100, and 48000
- non-finite PCM rates
- the supported 24000 Hz rate
- unchanged PCMU and PCMA formats

A patch changeset for `@openai/agents-realtime` is included.

### Issue number

None. Found while comparing the Agents SDK Realtime format normalization against the OpenAI Realtime API contract and Python SDK behavior.

### DIFF
--- a/.changeset/normalize-realtime-pcm-rate.md
+++ b/.changeset/normalize-realtime-pcm-rate.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix(realtime): normalize PCM audio to the supported 24 kHz rate
--- a/packages/agents-realtime/src/clientMessages.ts
+++ b/packages/agents-realtime/src/clientMessages.ts
@@ -286,11 +286,7 @@ export function normalizeAudioFormat(
 ): RealtimeAudioFormatDefinition | undefined {
   if (!format) return undefined;
   if (typeof format === 'object') {
-    if (
-      format.type === 'audio/pcm' &&
-      typeof format.rate === 'number' &&
-      !Number.isFinite(format.rate)
-    ) {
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
@codex review

seratch:
@sylvesterkaczmarek Agents SDK repos automatically trigger codex review for each push. So, you don't usually need to mention codex.

seratch:
Thanks for investigating this. The Realtime API currently supports 24 kHz PCM, but changing the configured rate is not equivalent to converting the audio. In the WebSocket path, `sendAudio()` forwards the caller's original bytes without resampling, so rewriting a 16 kHz input format to 24 kHz can turn an actionable configuration error into incorrectly interpreted audio.

The PR also does not include a supported user scenario or reproduction showing that automatic normalization is needed beyond API/Python parity. The supported approach is to provide 24 kHz PCM data with `rate: 24000`, or use PCMU/PCMA.

I am going to close this PR. If a real supported scenario demonstrates that the current behavior is insufficient, we can revisit a narrower design that rejects unsupported rates before connection rather than silently relabeling the audio.

PR #1701, opened 2026-08-16
fix: validate model retry counts

sylvesterkaczmarek:
### Summary

Validate `modelSettings.retry.maxRetries` before runner-managed retries begin.

The retry loop currently uses the configured numeric value directly. `NaN` makes the retry exhaustion comparison permanently false, so a policy returning `true` can retry indefinitely; `Infinity` is likewise unbounded. Fractional and negative values also do not represent a valid bounded retry count.

This change requires `maxRetries` to be a non-negative finite integer when provided and fails before issuing a model request. The default remains zero and valid integer retry counts are unchanged. Validation is shared by streaming and non-streaming retry entrypoints.

### Test plan

Added focused coverage for:

- negative, fractional, `NaN`, and infinite retry counts
- valid zero and positive integer retry counts
- unchanged zero default
- user-facing run failure before any model request for invalid configuration

### Issue number

None. Found while auditing the new runner-managed retry controls.

### DIFF
--- a/.changeset/validate-model-retry-counts.md
+++ b/.changeset/validate-model-retry-counts.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: validate model retry counts before execution
--- a/packages/agents-core/src/runner/modelRetry.ts
+++ b/packages/agents-core/src/runner/modelRetry.ts
@@ -163,6 +163,20 @@ export function validateModelTimeoutMs(
   return timeoutMs;
 }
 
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
@codex review

seratch:
Once we start having this type of validation, we need to have lots of similar logic across the SDK. So, we'd like to hold off adding this one for now.

sylvesterkaczmarek:
Understood. Closing this rather than introducing one-off validation ahead of a broader retry-config policy. Thanks for clarifying.

PR #1702, opened 2026-08-16
fix(core): validate the remaining sandbox compaction thresholds

thegoodengineer:
### Summary

#1691 added a finite check for `DynamicCompactionPolicy.thresholdRatio`. The same constructor's second argument and the sibling `StaticCompactionPolicy` were left unchecked, so the same invalid values still get through:

```ts
new StaticCompactionPolicy(Number.NaN).compactThreshold();            // NaN
new StaticCompactionPolicy(-240000).compactThreshold();               // -240000
new DynamicCompactionPolicy(0.9, Number.NaN).compactThreshold('x');   // NaN
new DynamicCompactionPolicy(0.9, -1).compactThreshold();              // -1
```

`fallbackThreshold` is not a rare path: `compactThreshold()` returns it for every model the SDK does not recognise, and for any call with no model at all.

Both classes are public API through `@openai/agents-core/sandbox`, and the value flows straight into the request that `CompactionCapability.samplingParams` builds. `NaN` does not survive JSON, so the capability emits a null threshold:

```
samplingParams: {"context_management":[{"type":"compaction","compact_threshold":null}]}
```

A negative or `Infinity` threshold is worse in a way, since it serialises fine and silently changes when compaction triggers.

### Fix

The same finite check #1691 introduced, applied to both absolute thresholds through one shared helper:

```ts
function assertCompactionThreshold(value: number, label: string): void {
  if (!Number.isFinite(value) || value < 0) {
    throw new UserError(
      `${label} must be a finite number greater than or equal to 0.`,
    );
  }
}
```

`0` stays valid, matching the existing boundary test that accepts a `thresholdRatio` of `0`.

### Test plan

Seven cases added to `packages/agents-core/test/sandboxCompactionPolicy.test.ts`, all failing on `main`:

- `StaticCompactionPolicy` rejects `NaN`, `Infinity`, `-Infinity` and `-1`
- `DynamicCompactionPolicy` rejects a `NaN`, `Infinity` or negative `fallbackThreshold`

Plus three that pass in both states, pinning what must not change: valid thresholds of `0` and `240000` still round-trip, a valid `fallbackThreshold` is still returned for unrecognised models and for no model, and `samplingParams` still serialises to a finite `compact_threshold`.

Verified locally on Windows:

- `pnpm test packages/agents-core/test/sandboxCompactionPolicy.test.ts`: 18 passing, 7 of them failing on `main`
- `tsc --noEmit -p packages/agents-core/tsconfig.json`: 0 errors
- `pnpm lint`
- Wider `agents-core` run shows no new failures. The pre-existing ones on this platform are the sandbox suites that need POSIX path separators and symlink support, and the RunState fixture hash that trips on a CRLF checkout.

### Issue number

Follow-up to #1691

### Checks

### DIFF
--- a/.changeset/validate-compaction-thresholds.md
+++ b/.changeset/validate-compaction-thresholds.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: validate the remaining sandbox compaction thresholds
--- a/packages/agents-core/src/sandbox/capabilities/compaction.ts
+++ b/packages/agents-core/src/sandbox/capabilities/compaction.ts
@@ -7,11 +7,20 @@ export abstract class CompactionPolicy {
   abstract compactThreshold(model?: string): number;
 }
 
+function assertCompactionThreshold(value: number, label: string): void {
+  if (!Number.isFinite(value) || value < 0) {
+    throw new UserError(
+      `${label} must be a finite number greater than or equal to 0.`,
+    );
+  }
+}
+
 export class StaticCompactionPolicy extends CompactionPolicy {
   readonly threshold: number;
 
   constructor(threshold: number = 240000) {
     super();
+    assertCompactionThreshold(threshold, 'StaticCompactionPolicy.threshold');
     this.threshold = threshold;
   }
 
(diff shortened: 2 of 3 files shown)


seratch:
We'd like to avoid client-side validation like this one as much as possible.

Issue #1708, opened 2026-08-18
pnpm test fails on Windows: eleven scripts use POSIX inline env assignment and cmd.exe rejects it


ErenAta16:
### Describe the bug

`pnpm test` fails immediately on Windows. pnpm runs scripts through `cmd.exe` there, and the script begins with a POSIX inline environment assignment, which `cmd.exe` does not understand.

`CONTRIBUTING.md:45` gives `pnpm test` as the command to run, and `CONTRIBUTING.md:70` and `AGENTS.md:219` give `CI=1 pnpm test`. Both fail on a default Windows setup, and the second fails in PowerShell too.

### Reproduction

```
> pnpm test
$ CI=1 NODE_ENV=test OPENAI_AGENTS_TEST_PROFILE=full vitest
'CI' is not recognized as an internal or external command,
operable program or batch file.
 ELIFECYCLE  Command failed with exit code 1.
```

Isolated from this repo, so it is the shell rather than anything in the suite. Same pnpm 11.18.0, same machine, two scripts in a throwaway package:

```
script without an env prefix        runs
script with "CI=1 NODE_ENV=test"    'CI' is not recognized as an internal or external command
```

Pointing `script-shell` at a POSIX shell makes it work:

```
> pnpm config set script-shell "C:/Program Files/Git/bin/bash.exe"
> pnpm run posix-style
CI=1 NODE_ENV=test
```

so a contributor who has already configured that will not see this, which may be why it has gone unnoticed.

### Scope

Eleven scripts in the root `package.json` start with an inline assignment: `test`, `test:review`, `test:watch`, `test:coverage`, `test:integration`, `test:integration:sdk-core`, `test:integration:sdk-capabilities`, `test:integration:api-contract`, `test:rclone-pin`, `test:workflow-scripts`, `api-contract:test`.

`cross-env` is not currently a dependency.

### Why CI does not catch it

`windows-sandbox-paths` in `.github/workflows/test.yml` avoids the problem in three independent ways, so the Windows leg stays green:

```yaml
env:
  CI: "1"
  NODE_ENV: test          # job-level, not inline
...
  - shell: bash           # not cmd
    run: pnpm exec vitest --run --project "@openai/agents-core" ...   # not `pnpm test`
```

Each of those is reasonable on its own. Together they mean CI never executes the script a Windows contributor is told to run.

### Expected behavior

`pnpm test` should work on Windows, or the prerequisite should be written down.

Three shapes, and I have no preference:

[...]

PR #1710, opened 2026-08-18
fix: make test scripts work on Windows

mikemikimike:
Make all test scripts shell-neutral on Windows by using explicit Vitest modes. Full and review modes preserve CI and profile behavior through Vitest configuration, watch remains watch-based, and coverage documentation uses the portable command.


mikemikimike:
Thanks for offering to test this on the Windows setup that reproduces #1708. The branch is now pushed. Could you please run the affected scripts, especially `pnpm test` and `pnpm test:watch` with a Vitest passthrough argument, and report whether they start successfully under your setup? The full suite on my Windows environment reaches Vitest but Unix-specific tests fail because `/bin/sh` is unavailable.

ErenAta16:
Ran this on the Windows setup that produced #1708, as offered. Branch `a4a893e`, Windows 11, node v24.16.0, pnpm 11.18.0. Short version: the fix works, the passthrough question I raised is answered, and the three test failures I hit are not yours.

**The mechanism.** There is no `.npmrc` in the repo, so pnpm runs scripts through `cmd.exe`. Same shell, both forms:

```
> CI=1 NODE_ENV=test OPENAI_AGENTS_TEST_PROFILE=full node -e ...
'CI' is not recognized as an internal or external command,
operable program or batch file.                                    exit 1

> node_modules\.bin\cross-env CI=1 NODE_ENV=test OPENAI_AGENTS_TEST_PROFILE=full node -e ...
{"CI":"1","NODE_ENV":"test","OPENAI_AGENTS_TEST_PROFILE":"full"}   exit 0
```

All three variables arrive, which is the part worth checking since two of them gate test behaviour.

**Completeness.** I parsed every `package.json` tracked in the repo rather than only the root one: 44 files, 11 scripts now wrapped, 0 left carrying a bare POSIX inline assignment. That matches the eleven in the report exactly, with nothing hiding in the workspace packages.

**It runs.** `pnpm test:workflow-scripts` now reaches vitest and executes 27 tests. Three fail:

```
run-integration-tests-managed.test.mjs  removes the API key from local children and gives only the validated environment to tests
workflow-contracts.test.mjs             starts no example child when service-account provenance is invalid
workflow-contracts.test.mjs             refuses to signal a stale unowned background pid
```

Those are not from this change. On `main` at `2d68a10`, bypassing the broken script by exporting `CI` and `NODE_ENV` in the shell and calling vitest directly, the same three fail with identical counts (3 failed, 24 passed, 27 total). They are process-spawn and signal behaviours that differ on Windows, and they are a separate problem from this one.

**Passthrough, which is the thing I asked about.** cross-env is transparent. Same arguments through an unwrapped and a wrapped script:

```
pnpm run bare    -- --run zzz foo   ->  argv ["--","--run","zzz","foo"]  FOO=unset
pnpm run wrapped -- --run zzz foo   ->  argv ["--","--run","zzz","foo"]  FOO=1
pnpm run bare       --run zzz foo   ->  argv ["--run","zzz","foo"]       FOO=unset
pnpm run wrapped    --run zzz foo   ->  argv ["--run","zzz","foo"]       FOO=1
```

Byte-identical apart from the variable, so nothing to fix. The literal `--` in the first two rows is pnpm forwarding its own separator and it predates this change. Confirmed end to end in the repo: `pnpm test --run zzz-no-such-file` gives `No test files found` with `filter: zzz-no-such-file`, so filters reach vitest through the wrapper.

One consequence worth knowing rather than acting on: `pnpm test -- --run <filter>` sends vitest a literal `--` and the filter is ignored, while `pnpm test --run <filter>` works. Identical before and after this PR, since cross-env does not touch argv.

**CI.** Only `changeset-validation` has reported on the PR so far, so the ubuntu test job has not yet confirmed the other half. The wrapping is shell-neutral and the four rows above show argv is unchanged, so I would not expect a surprise there, but it is still unverified rather than verified.

Happy to re-run any of this if you push again.

seratch (changes requested):
Thanks for addressing #1708 and for validating the behavior on Windows. After reviewing the dependency and its maintenance status, I would prefer not to introduce `cross-env`. The upstream repository is archived, and the latest version adds a transitive dependency with incomplete license-file packaging.

Please revise this PR to make the test scripts shell-neutral without adding a dependency. Use explicit Vitest commands such as `vitest run`, move `NODE_ENV=test` and the full/review profile selection into Vitest configuration, and preserve the current `CI` behavior, including rejection of `.only` in non-watch test commands. `test:watch` should remain explicitly watch-based.

Please also replace `CI=1 pnpm test` with `pnpm test` in `CONTRIBUTING.md` and `AGENTS.md`, since the external POSIX assignment would still fail before pnpm starts on Windows. This is a somewhat broader change, but it removes the shell dependency at its source and gives the repository a more maintainable cross-platform test configuration.

ErenAta16:
Re-ran on the same Windows setup at `2aa373e4`. The `--mode` approach is better than what I suggested, and it holds up here. Details below, including one file that still carries the old form.

**The thing I most wanted to check.** Dropping `cross-env` means the variables are now set inside `vitest.config.ts`, which runs in the main process, while tests run in worker threads (`pool: 'threads'`). Whether a `process.env` mutation in the config reaches a worker is not obvious, and on Windows worker startup differs, so I measured it instead of reasoning about it. Temporary probe inside `workspace-test-config`, removed afterwards:

```
mode=full     NODE_ENV=test  CI=1     PROFILE=full
mode=review   NODE_ENV=test  CI=1     PROFILE=review
mode=watch    NODE_ENV=test  CI=null  PROFILE=full
```

Those are exactly the three old script environments, including `CI` being absent under watch, which the old `test:watch` also did not set. So the mode mapping is faithful and not just approximately right.

**What you asked me to run.**

```
pnpm test:watch reviewTestProfile    1 file, 4 tests passed, then sat in watch mode until killed
pnpm test reviewTestProfile          1 file, 4 tests passed
```

Both start and both take the passthrough argument, so filters still reach vitest with the wrapper gone.

**No regression against the baseline.** `pnpm test:workflow-scripts` gives 3 failed, 24 passed, 27 total, identical to what `main` produces on this machine and to what the earlier cross-env version of this branch produced. Those three are the pre-existing Windows process and signal differences I described before, not yours.

**Completeness.** Rescanned all 44 tracked `package.json` files: 0 scripts left with a bare POSIX inline assignment, and 0 `cross-env`, so nothing is half-migrated.

**One file still has the old form.** `AGENTS.md` and `CONTRIBUTING.md` both moved from `CI=1 pnpm test` to `pnpm test`, but this one was missed:

```
.agents/skills/test-coverage-improver/SKILL.md:23
  - **Run coverage**: Execute `CI=1 pnpm test:coverage` at repo root.
```

Two reasons it is worth fixing in this PR. It is the exact form that fails in `cmd.exe` with `'CI' is not recognized as an internal or external command`, so an agent following that instruction on Windows hits the original bug. And it is now redundant everywhere, not just broken on Windows, because `test:coverage` runs `--mode full` and the config sets `CI` itself.

mikemikimike:
Fixed in aefafc44: .agents/skills/test-coverage-improver/SKILL.md now instructs agents to run pnpm test:coverage, removing the remaining POSIX-only CI=1 assignment.\n\nValidation: pnpm test --run reviewTestProfile passes (4/4), and git diff --check passes. pnpm test:workflow-scripts starts successfully and has the same three Windows-specific process/path failures documented in the prior test report (24 passed, 3 failed).

mikemikimike:
The intent is not to make the lowest-level tests unconditionally run with CI=1. The old root scripts used POSIX inline assignments, so on Windows cmd.exe rejected them before Vitest could start. This PR preserves the existing per-entry-point behavior by selecting a Vitest mode: the config applies NODE_ENV=test and CI=1 for the non-watch full and review modes, while watch deliberately leaves CI unset.

test:coverage represents the non-watch/full test path and explicitly runs vitest run --mode full --coverage. Its former CI=1 is now supplied by full mode rather than by the caller shell. Removing CI=1 from the skill instruction removes a redundant POSIX-only prefix, so the documented command works on Windows as well.

ErenAta16:
No disagreement on the design. The measurement I posted was checking exactly the mapping you describe, and it came out matching: `full` and `review` get `CI=1`, `watch` deliberately does not, and all three get `NODE_ENV=test` inside the worker thread. My point about `test:coverage` was the same as yours, that the prefix is now supplied by `--mode full` rather than by the caller.

Confirmed the fix landed at `4b48d9ba`:

```
-- **Run coverage**: Execute `CI=1 pnpm test:coverage` at repo root.
+- **Run coverage**: Execute `pnpm test:coverage` at repo root.
```

One line of the same file is still on the old assumption, and it is unchanged from `main` rather than something this PR introduced:

```
SKILL.md:14
1. From the repo root run `pnpm test:coverage` (set `CI=1` if needed) to regenerate `coverage/`.
```

By your own reasoning that parenthetical is now stale, since `--mode full` supplies it, and the obvious way for a reader to act on "set `CI=1`" at a `cmd.exe` prompt is the prefix that this PR just removed everywhere else. It is advice rather than a command, so nothing breaks if it stays. Mentioning it only because it is the same sentence one line up, and cheaper to drop now than to find later.

mikemikimike:
I re-ran the current PR head on Windows. The shell-neutral test entry points start successfully: pnpm test:coverage --run reviewTestProfile passes (4/4), and pnpm test:watch reviewTestProfile enters watch mode and passes the same filter.

One question on acceptance scope: pnpm test:workflow-scripts still reports 24 passed / 3 failed on Windows. The failures appear independent of this PR: one assertion assumes slash path separators for Verdaccio, and two expect Unix shell/process exit statuses (78 and 1) but receive null on Windows. Should those platform-specific workflow-contract failures be addressed in this PR before approval, or tracked separately from the shell-neutral test-script fix?

mikemikimike:
Good catch. Removed the stale Quick Start parenthetical in b2f6c9a4, so both coverage instructions now consistently say to run pnpm test:coverage without manually setting CI. Re-ran pnpm test:coverage --run reviewTestProfile on Windows: 4/4 passed.

ErenAta16:
Confirmed at `b2f6c9a4`. `CI=1` no longer appears anywhere in that file, and line 14 reads:

```
1. From the repo root run `pnpm test:coverage` to regenerate `coverage/`.
```

That closes everything I raised. To summarise where this ended up from the Windows side, so a reviewer does not have to reconstruct it from the thread: the mode mapping reproduces all three original script environments inside the worker thread, passthrough arguments still reach vitest, `test:workflow-scripts` matches the `main` baseline exactly with no new failures, all 44 tracked `package.json` files are free of bare POSIX inline assignments, and the docs no longer instruct a form that fails in `cmd.exe`.

Nothing outstanding from me. Thanks for taking the report this far rather than stopping at the eleven scripts.

seratch (changes requested):
Thanks for reworking this into a dependency-free Vitest mode design. The Windows evidence now demonstrates that the shell-neutral entry points preserve the required full, review, watch, and argument-forwarding behavior.

Two focused blockers remain before re-review:

1. Please restore `pnpm-lock.yaml` to the base version. No dependency specifications change anymore, but the current lockfile diff upgrades many unrelated resolutions.
2. In `vitest.config.ts`, set `OPENAI_AGENTS_TEST_PROFILE` from the selected mode and then derive the profile through the existing `isReviewTestProfile()` parser. The current `mode === 'review'` check bypasses that parser, so direct profile-based invocations are ignored and invalid values no longer fail fast even though the helper and its tests still define that behavior. Please update the contract test to verify this connection.

Please also refresh the PR description, which still describes the superseded `cross-env` implementation.

mikemikimike:
Addressed both blockers in f200abbe. pnpm-lock.yaml is restored to the base version with no remaining lockfile diff. vitest.config.ts now sets OPENAI_AGENTS_TEST_PROFILE from the selected mode and then calls the existing isReviewTestProfile() parser; the profile contract test asserts that connection. I also refreshed the PR description. Validation: pnpm test --run reviewTestProfile passes (4/4); pnpm test:workflow-scripts retains the documented Windows baseline (24 passed, 3 unrelated platform-specific failures).

seratch (approved):

PR #1722, opened 2026-08-19
fix(realtime): safely serialize guardrail feedback details

sylvesterkaczmarek:
### Summary

Prevent Realtime output-guardrail recovery from failing when a guardrail returns `outputInfo` that cannot be serialized by `JSON.stringify`.

`GuardrailFunctionOutput.outputInfo` is intentionally typed as `any`. Circular objects and values such as `BigInt` could therefore throw while the SDK was constructing the tripwire error or corrective feedback, interrupting the guardrail recovery path before the normal event/interruption/feedback flow completed.

### Changes

- add one safe `formatRealtimeGuardrailOutputInfo()` formatter
- use that formatter for both the `OutputGuardrailTripwireTriggered` message and model feedback
- preserve the existing `JSON.stringify` representation for normally serializable values
- fall back to the existing `toSmartString` utility only when JSON serialization fails
- keep circular and other non-JSON values from escaping as serialization exceptions
- include a patch changeset for `@openai/agents-realtime`

### Test plan

Coverage now verifies both the formatter and the actual `RealtimeSession` recovery path for circular and `BigInt` `outputInfo` values. The session regression asserts that:

- `guardrail_tripped` is emitted
- the active response is interrupted
- corrective feedback is sent to the model

The existing plain-object feedback assertion remains in place to verify that normal JSON formatting is unchanged.

### Issue number

None. Found while auditing the Realtime guardrail recovery path.


seratch (changes requested):
Thanks for the contribution. I reproduced this through the actual RealtimeSession path against both the base and the current PR head. A serializable outputInfo emits guardrail_tripped, interrupts once, and sends feedback. With circular or BigInt outputInfo, both revisions instead emit a JSON serialization TypeError with no guardrail event, no interruption, and no feedback.

The failure occurs because #runOutputGuardrails interpolates JSON.stringify(outputInfo) into OutputGuardrailTripwireTriggered before the new feedback helper is reached.

Please use one safe outputInfo formatter for both the tripwire error and feedback, and add a RealtimeSession regression test asserting the guardrail event, interruption or cancellation, and feedback for these non-JSON values.

sylvesterkaczmarek:
Addressed. The tripwire error and feedback now use the same formatRealtimeGuardrailOutputInfo() formatter, and I added a RealtimeSession regression covering circular and BigInt outputInfo that asserts guardrail_tripped, interruption, and corrective feedback. CI is currently awaiting the external-contributor approval gate; ready for re-review.

sylvesterkaczmarek:
Still active and ready for re-review. The requested RealtimeSession fix is addressed on the current head `4e7da8a`: both the tripwire error and corrective feedback now use the same safe `outputInfo` formatter, with regression coverage for circular and BigInt values through the actual session recovery path.

CI is also green on the current head. Could you please take another look when convenient?

seratch (approved):

PR #1723, opened 2026-08-20
fix(core): validate model retry counts

sylvesterkaczmarek:
### Summary

Validate `modelSettings.retry.maxRetries` before starting model execution.

The retry loops currently read the configured value directly. Non-finite or non-integer values therefore produce unsafe or surprising behavior: `Infinity` can keep retrying without a finite bound, fractional values effectively create an implicit rounded retry count, and negative values silently disable retries rather than being rejected as invalid configuration.

This change requires `maxRetries`, when provided, to be a non-negative finite integer and applies the same validation to both non-streaming and streaming model execution.

### Changes

- add a shared `validateModelMaxRetries()` boundary alongside the existing timeout validation
- use it before both retry loops begin
- preserve the existing default of zero retries when the option is omitted

### Test plan

Added focused coverage verifying that:

- negative, fractional, `NaN`, and infinite retry counts are rejected
- zero and positive integers remain valid
- an omitted retry count still resolves to zero
- invalid configuration fails before the underlying model receives any request

### Risk

Low. Valid existing retry configurations are unchanged. The behavior changes only for values that cannot represent a finite retry count safely.

### Issue number

None. Found while auditing the runner-managed retry boundary.

### DIFF
--- a/.changeset/validate-model-retry-counts.md
+++ b/.changeset/validate-model-retry-counts.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: validate model retry counts before execution
--- a/packages/agents-core/src/runner/modelRetry.ts
+++ b/packages/agents-core/src/runner/modelRetry.ts
@@ -163,6 +163,20 @@ export function validateModelTimeoutMs(
   return timeoutMs;
 }
 
+export function validateModelMaxRetries(modelSettings: ModelSettings): number {
+  const maxRetries = modelSettings.retry?.maxRetries ?? 0;
+  if (
+    !Number.isFinite(maxRetries) ||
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for rebasing this work. This is substantively the same change as #1701, which we closed after deciding not to add one-off retry-count validation ahead of a coherent retry-configuration policy. The new PR does not add a broader policy or new evidence that changes that maintainer decision. It also retains unresolved boundaries: validation happens after local session input can be persisted, and finite-but-unsafe integers are still accepted.

I am going to close this PR as a duplicate of the prior decision. Please revisit this only if maintainers decide to introduce a broader retry-configuration validation contract or there is new production evidence that justifies changing that scope.

PR #1724, opened 2026-08-20
fix(sandbox): make provider error inspection fail-safe

sylvesterkaczmarek:
### Summary

Keep sandbox provider failures inside the SDK's error boundary even when the thrown value cannot be inspected normally.

Provider integrations may throw arbitrary values. Today, diagnostic handling assumes operations such as `instanceof`, property access, and string conversion are safe. A Proxy with throwing traps, or an `Error` whose `message` is malformed, can therefore make the error-inspection path throw a second exception instead of returning a `SandboxProviderError` with safe diagnostics.

### Changes

- make provider message extraction fall back to a stable generic diagnostic when inspection or coercion throws
- make provider detail and retryability extraction fail closed to `{}` and `null`
- guard `UserError` classification against hostile prototype access
- keep resume/not-found inspection inside the same fail-safe boundary
- preserve the existing rich provider diagnostics for ordinary errors
- add a patch changeset for `@openai/agents-extensions`

### Test plan

Added focused regression coverage for:

- Proxy-backed provider errors whose property/prototype access throws
- wrapping hostile provider failures as `SandboxProviderError`
- non-string `Error.message` values
- message values whose string conversion itself throws
- hostile failures during resume/recreate classification

### Risk

Low. Normal provider errors continue through the existing diagnostic extraction path. The new behavior is limited to cases where inspecting the provider's thrown value would itself fail.

### Issue number

None. Found while auditing sandbox provider failure handling.

### DIFF
--- a/.changeset/failsafe-provider-errors.md
+++ b/.changeset/failsafe-provider-errors.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+fix(sandbox): make provider error inspection fail-safe
--- a/packages/agents-extensions/src/sandbox/shared/session.ts
+++ b/packages/agents-extensions/src/sandbox/shared/session.ts
@@ -10,6 +10,9 @@ import { isRecord } from './typeGuards';
 
 export { withSandboxSpan };
 
+const UNINSPECTABLE_PROVIDER_ERROR =
+  'Provider error could not be inspected safely.';
+
 export async function closeRemoteSessionOnManifestError(
   providerName: string,
   session: { close(): Promise<void> },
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for following up on #1699. That PR was closed because the hostile Proxy and malformed-message cases showed synthetic reachability but no supported provider, operation, or user impact. This submission adds more synthetic cases but does not provide the requested real-provider evidence, so the need gate is unchanged. The advertised fail-safe boundary is also still incomplete because provider-controlled detail formatting can throw after the new catches.

I am going to close this PR. If a supported sandbox provider reproduces this in practice, please share the provider, operation, thrown value shape, and observable consequence; we can then choose the narrowest owning boundary rather than growing a general hostile-object inspection contract.

PR #1725, opened 2026-08-20
fix(core): preserve sandbox failures during error serialization

sylvesterkaczmarek:
### Summary

Prevent sandbox telemetry serialization from replacing the original operation failure when the thrown value has hostile or malformed error metadata.

`withSandboxSpan()` records an error event before rethrowing the operation failure. The event serializer currently assumes that `instanceof Error`, `String(error)`, and optional properties such as `code` and `retryable` can be read safely. Those operations can themselves throw for Proxy-backed values or accessor properties, causing telemetry handling to obscure the failure the sandbox operation actually raised.

### Changes

- make Error classification fail-safe
- read `name`, `message`, `code`, and `retryable` defensively
- fall back to a stable diagnostic when an arbitrary thrown value cannot be converted to text
- preserve all safely readable Error fields when only optional metadata access fails

### Test plan

Added regressions verifying that:

- an unserializable thrown value is still rethrown by `withSandboxSpan()` unchanged
- the emitted error event uses a safe fallback message
- a throwing optional `code` getter does not discard a valid Error name/message
- Proxy-backed Error classification failures do not replace the original sandbox exception

### Risk

Low. Ordinary Error serialization is unchanged in shape. The additional guards only apply when classification, property access, or string conversion throws.

### Issue number

None. Found while auditing sandbox telemetry failure handling.

### DIFF
--- a/.changeset/safe-sandbox-error-serialization.md
+++ b/.changeset/safe-sandbox-error-serialization.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve sandbox failures when error serialization fails
--- a/packages/agents-core/src/sandbox/events.ts
+++ b/packages/agents-core/src/sandbox/events.ts
@@ -48,6 +48,8 @@ export type SandboxHttpEventSinkOptions = {
 };
 
 const sandboxEventSinks = new Set<SandboxEventSink>();
+const UNSERIALIZABLE_SANDBOX_ERROR_MESSAGE =
+  'Sandbox operation failed with an unserializable error.';
 
 export function addSandboxEventSink(sink: SandboxEventSink): () => void {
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for revisiting #1692. That PR was closed because the serialization edge case lacked a realistic provider or user path, and this submission does not add that evidence. It addresses the earlier hostile instanceof case, but the claimed preservation guarantee still does not hold on normal traced sandbox paths: after withSandboxSpan rethrows, the outer setSpanError path can read or stringify the same hostile value and replace the original failure.

I am going to close this PR rather than widen the tracing contract around a synthetic case. If a supported sandbox or provider path produces this shape in practice, please provide the provider, operation, thrown value, and observable consequence so we can evaluate a shared tracing-boundary fix.

PR #1751, opened 2026-08-24
fix(agents-extensions): decode sandbox stream output incrementally

pacocartones:
### Summary

`execCommand` decoded each streamed `stdout`/`stderr` SSE event with its own `new TextDecoder().decode(...)` call. The Cloudflare worker splits command output into base64 chunks at arbitrary byte offsets, so any multi-byte character straddling two events was decoded as U+FFFD replacement characters, and the corruption persisted in the string returned to the caller.

This keeps one streaming decoder per stream (`{ stream: true }`) and flushes both once every event has been processed, so split sequences are reassembled.

The command output is produced by a remote sandbox and reaches the model as tool output, so the split points are not under the caller's control.

### Test plan

Added `preserves multi-byte characters split across streamed output chunks` to `packages/agents-extensions/test/sandbox/cloudflare.test.ts`, splitting UTF-8 sequences mid-character across events. It fails on `main` (`caf�� �� 日本語`) and passes with this change.

`cloudflare.test.ts`: 65/65. Full `agents-extensions` suite: 1045/1045. `pnpm lint` and `pnpm build` clean.

### Issue number

N/A

### Checks

### DIFF
--- a/.changeset/cloudflare-sandbox-utf8-chunk-split.md
+++ b/.changeset/cloudflare-sandbox-utf8-chunk-split.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+fix: preserve multi-byte characters split across Cloudflare sandbox output chunks
--- a/packages/agents-extensions/src/sandbox/cloudflare/sandbox.ts
+++ b/packages/agents-extensions/src/sandbox/cloudflare/sandbox.ts
@@ -840,6 +840,11 @@ export class CloudflareSandboxSession implements SandboxSession<CloudflareSandbo
     }
 
     const decoder = new TextDecoder();
+    // The worker splits stdout/stderr into base64 chunks at arbitrary byte
+    // offsets, so a multi-byte character can straddle two events. Keep one
+    // streaming decoder per stream so those sequences survive the split.
+    const stdoutDecoder = new TextDecoder();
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for the contribution. I traced this against the official Cloudflare Sandbox bridge used by `CloudflareSandboxClient`. The bridge receives each `onOutput` payload as a string and UTF-8-encodes that complete string before base64-wrapping it in a stdout/stderr SSE event. Therefore a valid UTF-8 byte sequence cannot straddle two bridge events. If corruption happened earlier while Cloudflare decoded the process stream, the original bytes would already be lost and this client-side decoder could not recover them.

The new regression test manually creates byte-split base64 events that the supported bridge does not emit, so it proves the branch is constructible but not a supported user impact. I am going to close this PR for now. We can revisit if there is a capture or real-path reproduction from the supported bridge showing one UTF-8 sequence split across two stdout/stderr SSE events.

pacocartones:
Thanks for tracing this through the supported bridge. Agreed: because the bridge UTF-8-encodes each complete JavaScript string before base64-wrapping it, a valid multi-byte sequence cannot straddle two emitted SSE events. If corruption occurred before that boundary, an incremental decoder here could not recover the original bytes. The regression test therefore constructs an unsupported event shape, so closing is the right call. I will only revisit this with a capture or reproduction from the official bridge showing the split on the supported path.

PR #1753, opened 2026-08-24
fix(core): reject malformed Base64 consistently

praneethhere:
## Summary

This fixes runtime-dependent Base64 validation in `@openai/agents-core`.

Node's permissive `Buffer` decoder silently accepted malformed inputs such as `AQI!` and `AQI-`, while the browser `atob` and manual decoding paths rejected them. This caused identical input to behave differently depending on the runtime.

The decoder now:

* Validates and normalizes standard Base64 before selecting a runtime backend.
* Preserves support for ASCII whitespace and omitted padding.
* Rejects invalid characters, malformed padding, impossible lengths, and non-standard Base64URL alphabet input with a consistent `TypeError`.
* Adds regression coverage across the `Buffer`, `atob`, and manual decoder paths.
* Adds caller-level coverage for malformed persisted binary manifest content.

## Test plan

* `pnpm test packages/agents-core/test/utils/base64.test.ts packages/agents-core/test/sandboxShared.test.ts`

  * 69 tests passed.
* Full code-change verification under Node.js 22.23.2:

  * Build, build-check, dist-check, lint, and formatting passed.
  * 206 test files passed.
  * 6,277 tests passed and 2 skipped.
* TruffleHog scanned the staged changes with zero verified or unverified secrets.

## Changeset

Included a patch changeset for `@openai/agents-core`.

### DIFF
--- a/.changeset/tidy-base64-decoders.md
+++ b/.changeset/tidy-base64-decoders.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: reject malformed Base64 consistently across runtime decoders
--- a/packages/agents-core/src/utils/base64.ts
+++ b/packages/agents-core/src/utils/base64.ts
@@ -57,17 +57,19 @@ export function decodeBase64ToUint8Array(value: string): Uint8Array {
     return new Uint8Array();
   }
 
+  const padded = normalizeBase64(value);
+
   const globalBuffer =
     typeof globalThis !== 'undefined' && (globalThis as any).Buffer
       ? (globalThis as any).Buffer
       : undefined;
 
   if (globalBuffer) {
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for taking the time to investigate this. I confirmed that the decoder backends do not share the same acceptance rules, especially because Node's base64 decoder accepts the Base64URL alphabet. However, the SDK's manifest serializer always emits canonical standard Base64, and the caller-level regression test reaches the claimed failure only by mutating SDK-generated state with `as any`. The PR does not show a supported producer, real RunState/sessionState, or user-visible consequence for these malformed values.

I am going to close this PR for now. If a supported client or real persistence workflow produces such data and the mismatch causes an incorrect resume or materialized file, please share that evidence and we can revisit the narrowest fix.

PR #1771, opened 2026-08-28
fix: reconcile pending computer calls on stream abort

### LINKED ISSUE #1770: fix(agents-core): `createGuardrailTracker.throwIfError()` skips awaiting pending in-flight guardrails
### The bug

In `packages/agents-core/src/runner/guardrails.ts`, `createGuardrailTracker().throwIfError()` checks `if (error)` before awaiting `promise`:

```ts
const throwIfError = async () => {
  if (error) {
    if (promise) {
      await promise;
    }
    throw error;
  }
};
```

When input guardrails run in parallel (`runInParallel: true`), `error` is initially `undefined` while the promise is still resolving. 

When `await guardrailTracker.throwIfError()` is invoked at admission checkpoints (such as in `run.ts` lines 1875, 2894, and 3115), `if (error)` evaluates to `false` and the function returns immediately without awaiting `promise`.

As a consequence, in-flight guardrail errors and tripwires are bypassed at turn setup, allowing unvalidated input to be admitted into the execution turn before the guardrail settles.

### Debug information

- Agents SDK version: Latest `main` (`@openai/agents-core` 0.3.x)
- Related package versions: `@openai/agents-core`
- Runtime and version: Node.js 22+
- Operating system: All
- Model and model provider: OpenAI
- Does the issue reproduce with the latest Agents SDK release? Yes
- Does the issue occur consistently or intermittently? Consistently whenever guardrails are still resolving when `throwIfError()` is called.

### Repro steps

```typescript
import { Agent, run, defineInputGuardrail } from '@openai/agents';

const slowFailingGuardrail = defineInputGuardrail({
  name: 'slow_check',
  runInParallel: true,
  run: async () => {
    // Simulate an async tripwire or remote validation check
    await new Promise((resolve) => setTimeout(resolve, 50));
    return {
      outputInfo: { error: 'Blocked by guardrail' },
      tripwireTriggered: true,
    };
  },
});

const agent = new Agent({
  name: 'TestAgent',
  inputGuardrails: [slowFailingGuardrail],
});

// When run() executes turn preparation, throwIfError() runs while slowFailingGuardrail
// is still pending, returning immediately rather than waiting for the tripwire to fail.
const result = await run(agent, 'Hello');
```

### Expected behavior

`throwIfError()` should await `promise` before evaluating `error`, ensuring that any in-flight guardrail failure is awaited and rethrown:

```ts
const throwIfError = async () => {
  if (promise) {
    await promise;
  }
  if (error) {
    throw error;
  }
};
```

In comparison, `awaitCompletion()` already follows this correct pattern:
```ts
const awaitCompletion = async (options?: { suppressErrors?: boolean }) => {
  if (promise) {
    await promise;
  }
  if (error && !options?.suppressErrors) {
    throw error;
  }
};
```

code-1705:
### Summary

When a stream aborts, the runner synthesizes results for in-flight tool calls to prevent execution deadlocks. Previously, the `StreamAbortReconciliationState` omitted tracking for `computer_call`, which would cause the runner to hang if a computer call was in-flight when the abort happened. 

This update adds `pendingComputerCalls` to the reconciliation state and provides fallback screenshot generation, bringing computer calls to parity with functions, shell calls, and other streaming tools.

### Test plan

Added unit tests in `streamReconciliation.test.ts` to verify that `computer_call` stream aborts are properly tracked and synthesized into a fallback 1x1 screenshot result. Verified all tests, typechecking, linting, and build pass locally using the full code change verification stack.

### Issue number

Closes #1770 

### Checks

### DIFF
--- a/.changeset/stream-abort-reconcile-computer-calls.md
+++ b/.changeset/stream-abort-reconcile-computer-calls.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: reconcile pending computer calls on stream abort
--- a/packages/agents-core/src/run.ts
+++ b/packages/agents-core/src/run.ts
@@ -3022,7 +3022,10 @@ export class Runner extends RunHooks<any, AgentOutputType<unknown>> {
                     preparedCall,
                   ),
                   conversationId: preparedCall.conversationId,
-                  modelSettings: preparedCall.modelSettings,
+                  modelSettings: {
+                    ...preparedCall.modelSettings,
+                    toolChoice: 'none',
+                  },
                   _internal: preparedCall.modelRequestInternal,
                   tools: preparedCall.serializedTools,
                   toolsExplicitlyProvided: preparedCall.toolsExplicitlyProvided,
--- a/packages/agents-core/src/runner/streamReconciliation.ts
+++ b/packages/agents-core/src/runner/streamReconciliation.ts
@@ -29,6 +29,7 @@ export type StreamAbortReconciliationState = {
   pendingShellCalls: Map<string, PendingStreamedToolCall>;
   pendingApplyPatchCalls: Map<string, PendingStreamedToolCall>;
   pendingProgramCalls: Map<string, string>;
+  pendingComputerCalls: Map<string, Pick<ComputerUseCallItem, 'callId'>>;
 };
 
 // 1x1 transparent PNG data URL used when a computer result requires an image
@@ -102,6 +103,7 @@ export function createStreamAbortReconciliationState(): StreamAbortReconciliatio
     pendingShellCalls: new Map(),
     pendingApplyPatchCalls: new Map(),
     pendingProgramCalls: new Map(),
+    pendingComputerCalls: new Map(),
   };
 }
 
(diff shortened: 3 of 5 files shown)


sylvesterkaczmarek (commented):
The `toolChoice: 'none'` follow-up fixes the earlier forced-computer-call failure mode, but I don't see a regression test that reaches that orchestration boundary. The added `streamReconciliation.test.ts` cases verify pending computer-call tracking and synthetic output construction only; they would still pass if the `run.ts` override were removed. Since the original bug occurs when the reconciliation model request inherits a forced tool choice and emits another `computer_call`, could we add one runner-level regression with an initial required computer tool choice and assert that the reconciliation request uses `toolChoice: 'none'` and terminates without creating another pending call? That would pin the exact failure mode without widening the change.

code-1705:
Added the runner-level regression test in run.test.ts (reconciles streamed pending computer calls using toolChoice none under required tool choice). It configures an agent with toolChoice: 'required', streams an aborting computer call, and asserts that the follow-up reconciliation request overrides toolChoice to 'none' and completes without triggering another forced call.

PR #1774, opened 2026-08-29
fix(agent.ts): correct typo in comments for agent tools description

morozow:
## Description
This pull request fixes a typo in the JSDoc comment for the Agent class method that returns all agent tools. The comment read "MCPl" (with a stray `l`) and "ALl" (with an uppercase first `L`) – corrected to "All agent tools, including the MCP and function tools." for clarity.

No runtime behavior is affected; this is a source-level documentation correction only.


packages/agents-core/src/agent.ts
```diff
@@ -1124,7 +1124,7 @@ export class Agent<
   }
 
   /**
-   * ALl agent tools, including the MCPl and function tools.
+   * ALl agent tools, including the MCP and function tools.
```
seratch:
```suggestion
   * All agent tools, including the MCP and function tools.
```

seratch: can you fix this as well?

morozow: Done – fixed this as well. Thanks for catching it.

PR #1781, opened 2026-09-01
fix(examples): run Prisma setup on Windows

sylvesterkaczmarek:
### Summary

Fix the Prisma memory example on Windows under current Node versions.

`examples/memory/start-prisma.ts` selects `pnpm.cmd` on Windows and passes it directly to `spawnSync`. Node 20.12.2 and newer reject direct `.cmd`/`.bat` spawning, so `pnpm -F memory start:prisma` can fail before Prisma runs.

This is a remaining occurrence of the Windows shim pattern reported in #1775. The open #1776 fix covers the verification runner and `scripts/prettier-changed.mjs`, but does not cover the Prisma memory example.

### Changes

- route Prisma commands through `ComSpec` / `cmd.exe` on Windows
- keep direct `pnpm` spawning on non-Windows platforms
- preserve the existing Prisma arguments, working directory, environment, exit-code, and signal handling
- add focused regression coverage for Windows, non-Windows, and custom `ComSpec` invocation

### Test plan

- added focused invocation tests under the workspace Vitest project
- full Windows runtime execution is left to CI/reviewer validation

### Issue number

No separate issue. Found while checking the remaining Windows `.cmd` spawn sites after #1775.

### Checks

- [x] Added focused regression coverage
- [x] No published package code changed, so no changeset is required

### DIFF
--- a/examples/memory/prisma-command.ts
+++ b/examples/memory/prisma-command.ts
@@ -1,0 +1,30 @@
+export type PrismaCommand = 'db-push' | 'generate';
+
+type PrismaInvocation = {
+  command: string;
+  args: string[];
+  windowsVerbatimArguments?: boolean;
+};
+
+const PRISMA_COMMAND_ARGS = {
+  'db-push': ['prisma', 'db', 'push', '--schema', './prisma/schema.prisma'],
+  generate: ['prisma', 'generate', '--schema', './prisma/schema.prisma'],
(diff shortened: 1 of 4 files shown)


sylvesterkaczmarek:
Still relevant. All current checks are green, including the Windows sandbox-path coverage; this remains a small examples-only fix for Prisma setup on Windows. Would appreciate maintainer review when convenient.

PR #1785, opened 2026-09-01
fix(examples) avoid deprecated Windows shell spawning

### LINKED ISSUE #1782: Windows image-generation example uses deprecated shell:true argument spawning
The `examples/tools/image-generation.ts` helper opens generated files on Windows with:

```ts
spawnSync('cmd', ['/c', 'start', '', filePath], { shell: true });
```

Current Node deprecates passing `args` with `shell: true` (since Node 22.15.0 / 23.11.0), and the extra shell layer also makes Windows path handling more fragile than necessary. Node's documented alternative is to spawn `cmd.exe` directly and pass the command arguments to it.

A focused fix can keep the current `start` behavior while invoking `%ComSpec%`/`cmd.exe` directly, preserving the file path as a distinct argument and leaving macOS/Linux behavior unchanged. I can send a small PR with regression coverage for the platform-specific invocation.

sylvesterkaczmarek:
### Summary

Fix the Windows file-opening path in the image-generation example.

`examples/tools/image-generation.ts` currently launches `cmd` with an argument array and `shell: true`. Current Node versions deprecate passing `args` when `shell: true`, and the extra shell layer makes Windows path handling unnecessarily fragile.

This changes the example to spawn `%ComSpec%` / `cmd.exe` directly on Windows while keeping the existing `start` behavior. macOS continues to use `open`, and Linux continues to use `xdg-open`.

### Changes

- remove the deprecated `shell: true` argument-spawning pattern on Windows
- invoke `%ComSpec%` / `cmd.exe` directly with the file path kept as a separate argument
- preserve existing macOS and Linux behavior
- add focused regression coverage for Windows, macOS, Linux, and custom `ComSpec`

### Test plan

- added focused platform-invocation tests under the workspace Vitest project
- no published package code changed, so no changeset is required

### Issue

Closes #1782

### DIFF
--- a/examples/tools/image-generation.ts
+++ b/examples/tools/image-generation.ts
@@ -3,15 +3,11 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { spawnSync } from 'node:child_process';
+import { getFileOpenInvocation } from './open-file';
 
 function openFile(filePath: string): void {
-  if (process.platform === 'darwin') {
-    spawnSync('open', [filePath], { stdio: 'inherit' });
-  } else if (process.platform === 'win32') {
-    spawnSync('cmd', ['/c', 'start', '', filePath], { shell: true });
-  } else {
-    spawnSync('xdg-open', [filePath], { stdio: 'inherit' });
-  }
+  const invocation = getFileOpenInvocation(filePath);
+  spawnSync(invocation.command, invocation.args, { stdio: 'inherit' });
 }
 
 async function main() {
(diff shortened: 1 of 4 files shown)


seratch:
The deprecation warning is a valid observation, but the report does not establish an image-opening failure, and the new tests only compare invocation arrays. I do not think the additional platform helper and test configuration are worth maintaining for the demonstrated impact.

I recommend closing #1782 and #1785 as currently proposed. We can consider a small inline cleanup separately when there is a concrete compatibility requirement, without introducing this helper abstraction.

PR #1786, opened 2026-09-01
fix(openai): bound websocket keepalive timers

sylvesterkaczmarek:
### Summary

Reject Responses WebSocket keepalive durations that exceed Node's maximum timer delay.

`pingIntervalMs` and `pingTimeoutMs` currently accept any positive finite number, then pass it to `setInterval`/`setTimeout`. Node timers support delays up to 2,147,483,647 ms; larger values overflow and can turn an intended long keepalive interval or pong timeout into almost immediate activity or failure.

### Changes

- cap both keepalive durations at 2,147,483,647 ms
- keep `null`/`undefined` behavior unchanged
- document the bound in the public option JSDoc
- add regression coverage for the maximum accepted value and maximum + 1 rejection
- add a patch changeset for `@openai/agents-openai`

### Test plan

- focused unit coverage in `responsesWebSocketConnection.test.ts`
- full repository execution is left to GitHub Actions

### Issue number

None. Found while auditing Responses WebSocket timer handling.

### Checks

- [x] Added regression tests
- [x] Added a changeset
- [x] No public API signature changes

### DIFF
--- a/.changeset/calm-ws-timers.md
+++ b/.changeset/calm-ws-timers.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+Reject Responses WebSocket keepalive durations that exceed Node's maximum timer delay.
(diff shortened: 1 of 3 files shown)


seratch:
The tests show what happens with invalid or extreme numeric settings, but they do not establish an ordinary configuration path that produces them. Please provide a concrete application configuration and caller-visible failure before expanding runtime validation. For WebSocket keepalive, null already provides the supported disabled setting.

PR #1787, opened 2026-09-01
fix(core): reject malformed persisted sandbox base64

sylvesterkaczmarek:
### Summary

Sandbox manifests persist binary file content as base64. On Node, `Buffer.from(value, 'base64')` accepts malformed inputs that browser `atob` rejects, so corrupted or tampered persisted state can silently resume with empty or altered file bytes instead of failing.

Validate standard base64 before decoding so Buffer, `atob`, and the manual fallback reject malformed data consistently. Valid padded and unpadded base64 and ASCII whitespace remain supported; serialization and the persisted manifest format are unchanged.

### Test plan

- added cross-runtime decoder regression coverage for valid whitespace/unpadded input and malformed characters, padding, and base64url input
- added a sandbox manifest persistence regression proving malformed persisted binary content is rejected
- verified the final branch is exactly one commit ahead of current upstream `main` with only the intended four files changed
- full monorepo verification is left to GitHub Actions

### Issue number

N/A. Found while auditing sandbox persistence/resume handling.

### Checks

- [x] I've added new tests
- [x] No documentation change is required; the public API and persisted format are unchanged
- [ ] I've confirmed all repository verification steps pass
- [x] I've added a patch changeset for `@openai/agents-core`

### DIFF
--- a/.changeset/strict-sandbox-base64.md
+++ b/.changeset/strict-sandbox-base64.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: reject malformed base64 in persisted sandbox files consistently
--- a/packages/agents-core/src/utils/base64.ts
+++ b/packages/agents-core/src/utils/base64.ts
@@ -49,6 +49,18 @@ export function encodeUint8ArrayToBase64(data: Uint8Array): string {
   return result;
 }
 
+function normalizeBase64ForDecoding(value: string): string {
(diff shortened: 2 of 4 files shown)


seratch:
The SDK writer already emits valid standard base64. The PR demonstrates manually corrupted persisted data, but not an ordinary corruption source or a security boundary that this validation would protect. I would close this unless a concrete resume failure establishes the need.

PR #1788, opened 2026-09-01
fix(core): preserve empty handoff descriptions

sylvesterkaczmarek:
### Summary

Preserve an explicit empty `toolDescriptionOverride` when creating a handoff.

`handoff()` currently applies the override only when the string is truthy, so:

```ts
handoff(agent, { toolDescriptionOverride: '' })
```

silently falls back to the generated handoff description.

That disagrees with the public optional-string override contract and with the existing `Handoff.clone({ toolDescription: '' })` and `Agent.asTool({ toolDescription: '' })` behavior, both of which preserve an explicit empty description.

This changes the check from truthiness to `undefined`, so omission still uses the generated description while `''` remains an explicit model-facing value.

### Test plan

- added public-boundary regression coverage showing an explicit empty handoff description reaches both `Handoff.toolDescription` and `getHandoffAsFunctionTool().description`
- added a guard showing an omitted override still keeps the generated handoff description
- verified the final branch is exactly one commit ahead of current upstream `main` with only the intended three files changed

### Issue number

N/A. Found while auditing description override parity.

### Checks

### DIFF
--- a/.changeset/preserve-empty-handoff-description.md
+++ b/.changeset/preserve-empty-handoff-description.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve explicit empty handoff tool descriptions
--- a/packages/agents-core/src/handoff.ts
+++ b/packages/agents-core/src/handoff.ts
@@ -429,7 +429,7 @@ export function handoff<
     handoff.toolName = config.toolNameOverride;
   }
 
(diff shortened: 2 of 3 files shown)


seratch:
The change makes an empty override distinguishable from an omitted one, but the PR does not demonstrate a workflow that needs an empty handoff description or a user-visible problem with the generated description.

PR #1789, opened 2026-09-01
fix(core): isolate MCP cache keys

sylvesterkaczmarek:
### Summary

Fix shared MCP tool caching for cache keys and server names that collide with JavaScript Object prototype properties.

The shared cache currently stores entries in ordinary `{}` objects. `MCPToolCacheKeyGenerator` is a public callback that may return any string, and MCP server names are also strings. Keys such as `__proto__` and `constructor` therefore resolve inherited properties before an actual cache entry exists.

For example, a first cached lookup using `__proto__` can treat `Object.prototype` as an `MCPTool[]` instead of calling `listTools()`. A server named `__proto__` can similarly collide with the per-server cache-key index and fail when the cache tries to call `.add()` on the inherited object.

Use null-prototype records for both cache tables so every string key is treated purely as data. Cache key generation, cache contents, invalidation semantics, and normal server names are unchanged.

### Test plan

- added regression coverage for public custom cache keys `__proto__` and `constructor`
- added regression coverage for a server named `__proto__`
- each regression verifies the second lookup is cached and invalidation forces a fresh `listTools()` call
- verified the branch is exactly one commit ahead of current upstream `main` with only the intended three files changed

### Issue number

N/A. Found while auditing shared MCP cache key handling.

### Checks

### DIFF
--- a/.changeset/isolate-mcp-cache-keys.md
+++ b/.changeset/isolate-mcp-cache-keys.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: isolate MCP cache keys from object prototype properties
--- a/packages/agents-core/src/mcpToolCache.ts
+++ b/packages/agents-core/src/mcpToolCache.ts
@@ -1,7 +1,8 @@
 import type { MCPTool } from './mcpShared';
 
-export const cachedMcpTools: Record<string, MCPTool[]> = {};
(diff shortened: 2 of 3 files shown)


seratch:
The null-prototype records handle the constructed keys correctly. However, the report does not identify an affected MCP server, cache-key producer, or trust boundary that turns this into a practical issue.

PR #1790, opened 2026-09-01
fix(openai): preserve websocket session callback errors

sylvesterkaczmarek:
### Summary

Preserve the callback failure when both a `withResponsesWebSocketSession(...)` callback and provider cleanup fail, even if the callback `Error` cannot accept cleanup metadata.

The helper already treats the callback failure as primary and tries to attach the cleanup failure as `cause` or `cleanupError`. For a frozen or otherwise non-extensible `Error`, assigning that metadata throws a `TypeError`, which currently replaces the original application error.

Make cleanup-error metadata attachment best-effort. Mutable errors still receive the cleanup failure as metadata; frozen errors remain the original rejection. Successful callback and cleanup behavior is unchanged.

### Test plan

- added regression coverage showing mutable callback errors still receive the cleanup error as `cause`
- added regression coverage showing a frozen callback error remains the thrown value when provider cleanup also fails
- verified the branch is exactly one commit ahead of current upstream `main` with only the intended three files changed

### Issue number

N/A. Found while auditing Responses WebSocket session cleanup failure handling.

### Checks

### DIFF
--- a/.changeset/preserve-websocket-session-callback-error.md
+++ b/.changeset/preserve-websocket-session-callback-error.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: preserve callback failures when websocket session cleanup metadata cannot be attached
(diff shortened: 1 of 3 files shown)


seratch:
Preserving the primary callback error is a useful contract, but this PR only demonstrates a manually frozen Error combined with cleanup failure. Please identify an ordinary supported error producer that reaches this case.

PR #1791, opened 2026-09-01
fix(openai): keep Conversations history limits local

sylvesterkaczmarek:
### Summary

`OpenAIConversationsSession.getItems(limit)` uses the SDK history limit as a total item cap, while the Conversations API `limit` is a per-page size. Passing large SDK limits through unchanged can exceed the provider's page-size bound.

Keep the total limit local while retaining an efficient provider page size:

- omitted SDK limits preserve the existing ascending-history request without a provider `limit`;
- finite positive limits request descending pages with `limit: Math.min(limit, 100)`;
- client pagination continues across page boundaries until enough Agent SDK items have been collected;
- the existing reverse/trim logic still returns the newest requested items in chronological order, including when one provider item expands into multiple Agent items.

### Test plan

- restored the existing focused Conversations session tests from current `main`, so small finite limits continue to verify the existing provider request shape;
- added a pagination regression using the OpenAI client's `ConversationCursorPage` implementation rather than a flat async iterator;
- the regression requests 150 history items from two 100-item descending provider pages, verifies the first page is capped at 100, verifies the second request carries the `after` cursor and the same page size, and verifies the SDK returns exactly `user-50` through `user-199` in chronological order;
- pushed-head CI is currently waiting at `action_required` for fork workflow approval.

### Issue number

N/A. Found while auditing Conversations session pagination parity.

### Checks

- [x] I've added new tests
- [x] Documentation change is not required; this preserves the existing `Session.getItems(limit)` contract
- [ ] Pushed-head CI is awaiting workflow approval

### DIFF
--- a/.changeset/local-conversation-history-limit.md
+++ b/.changeset/local-conversation-history-limit.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: keep Conversations session history limits local
--- a/packages/agents-openai/src/memory/openaiConversationsSession.ts
+++ b/packages/agents-openai/src/memory/openaiConversationsSession.ts
@@ -160,7 +160,6 @@ export class OpenAIConversationsSession
     const itemGroups: AgentInputItem[][] = [];
     let total = 0;
     const iterator = this.#client.conversations.items.list(conversationId, {
-      limit,
       order: 'desc' as const,
     });
 
--- a/packages/agents-openai/src/memory/openaiConversationsSession.ts
+++ b/packages/agents-openai/src/memory/openaiConversationsSession.ts
@@ -160,6 +160,7 @@ export class OpenAIConversationsSession
     const itemGroups: AgentInputItem[][] = [];
     let total = 0;
     const iterator = this.#client.conversations.items.list(conversationId, {
+      limit: Math.min(limit, 100),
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
The total history limit must not exceed the Conversations API's per-page limit, but removing the page size entirely unnecessarily changes existing requests. With the documented default of 20, getItems(100) can require five pages for 100 ordinary items instead of one.

Please retain the local total limit and clamp a defined API page size to 100, preserving the current behavior when the limit is omitted. The regression should cover actual page boundaries and returned item order; a flat mock iterator does not verify the pagination behavior.

sylvesterkaczmarek:
Addressed the review. Finite history reads now keep the SDK total limit local while using `Math.min(limit, 100)` as the Conversations API page size. I also replaced the flat 200-item mock with focused coverage that uses `ConversationCursorPage` across two real page boundaries, including the `after` cursor and returned chronological order. The PR is mergeable; pushed-head CI is currently waiting for fork workflow approval.

sylvesterkaczmarek:
Addressed the pagination feedback on the current head: the local total limit remains separate, the API page size is clamped to 100, and the focused Conversations session coverage has been restored around the page/ordering behavior.

seratch (approved):

PR #1792, opened 2026-09-01
fix(realtime): validate guardrail debounce length

sylvesterkaczmarek:
### Summary

Validate `outputGuardrailSettings.debounceTextLength` before Realtime guardrails use it for transcript debounce math.

The public setting documents a positive character interval, with `-1` as the only special value for running guardrails only when the complete transcript is available. Other values currently produce broken or misleading behavior:

- `0` divides by zero, causing the first debounce index to become `Infinity` and suppressing later incremental runs
- negative values other than `-1` silently behave like the final-transcript-only sentinel
- `NaN` can make the debounce comparison fail open on every delta
- infinities can suppress incremental execution

Reject invalid values at configuration time with `UserError`. Positive finite values, the default of `100`, and the documented `-1` sentinel are unchanged.

### Test plan

- extended `guardrail.test.ts` to preserve the default and a positive override
- added coverage for the documented `-1` sentinel
- added rejection coverage for `0`, `-2`, `NaN`, positive infinity, and negative infinity

### Issue number

N/A. Found while auditing Realtime guardrail configuration boundaries.

### Checks

- [x] No documentation update is required; the existing public JSDoc already documents the supported values

### DIFF
--- a/.changeset/validate-realtime-guardrail-debounce.md
+++ b/.changeset/validate-realtime-guardrail-debounce.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-realtime': patch
+---
+
+fix: validate realtime output guardrail debounce lengths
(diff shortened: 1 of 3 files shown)


seratch:
The tests show what happens with invalid or extreme numeric settings, but they do not establish an ordinary configuration path that produces them. Please provide a concrete application configuration and caller-visible failure before expanding runtime validation.

PR #1793, opened 2026-09-01
fix(openai): validate default mode settings

sylvesterkaczmarek:
### Summary

Validate the SDK-wide OpenAI API and Responses transport setters at runtime.

Both setters expose literal TypeScript types, but JavaScript callers and runtime values can still pass arbitrary strings. Today those invalid values are stored directly:

- `setOpenAIAPI('response')` makes `shouldUseResponsesByDefault()` return false, silently selecting Chat Completions.
- `setOpenAIResponsesTransport('ws')` makes `shouldUseResponsesWebSocketByDefault()` return false, silently selecting HTTP.

Reject unsupported values before mutating the defaults. Valid `responses` / `chat_completions` and `http` / `websocket` settings are unchanged.

### Test plan

- added regression coverage for an invalid OpenAI API runtime value and verified the prior `responses` setting remains active
- added regression coverage for an invalid Responses transport runtime value and verified the prior `websocket` setting remains active
- existing valid-mode tests continue to cover both supported values for each setter
- reset the mutable defaults after each test so the file remains order-independent

### Issue number

N/A. Found while auditing runtime validation of SDK-wide default mode setters.

### Checks

- [x] No documentation update is required; supported values are unchanged

### DIFF
--- a/.changeset/validate-openai-default-modes.md
+++ b/.changeset/validate-openai-default-modes.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: validate default OpenAI API and Responses transport modes
--- a/packages/agents-openai/src/defaults.ts
+++ b/packages/agents-openai/src/defaults.ts
@@ -1,3 +1,4 @@
+import { UserError } from '@openai/agents-core';
 import { loadEnv } from '@openai/agents-core/_shims';
 import METADATA from './metadata';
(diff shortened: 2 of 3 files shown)


seratch:
Both setters already expose a literal-string contract. The current reproductions bypass that contract with casts and do not show a supported producer or an SDK-owned side effect that warrants another validation layer. I would close this PR unless a concrete runtime integration demonstrates that need.

PR #1794, opened 2026-09-01
fix(core): validate tracing batch options

sylvesterkaczmarek:
### Summary

Validate the `BatchTraceProcessor` options that control batch progress and the recurring export timer.

`maxBatchSize` is currently used directly as the `Array.splice()` delete count. A value of `0` therefore makes normal scheduled/threshold exports remove and export an empty batch while leaving the trace buffer untouched, so the export loop cannot make progress and the queue can eventually fill and drop traces. Fractional and non-finite values likewise do not represent a valid batch size.

`scheduleDelay` is also passed directly to the runtime timer. Non-finite, non-positive, or values above the JavaScript timer limit can be normalised or overflow to near-zero delays in Node, turning the recurring export loop into a tight timer loop instead of the configured schedule.

Fail fast when `maxBatchSize` is not a positive integer, or when `scheduleDelay` is not a positive finite value at or below `2147483647` milliseconds. Existing valid queue sizing, trigger ratios, and batch sizes larger than the queue remain unchanged.

### Test plan

- added focused validation coverage for `maxBatchSize` values `0`, `-1`, `1.5`, `NaN`, and `Infinity`
- added focused validation coverage for `scheduleDelay` values `0`, `-1`, `NaN`, both infinities, and a value above the JavaScript timer limit
- verified v0.17.0 has the same unvalidated constructor behavior

### Issue number

N/A. Found while auditing tracing processor runtime configuration boundaries.

### Checks

### DIFF
--- a/.changeset/validate-tracing-batch-options.md
+++ b/.changeset/validate-tracing-batch-options.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: validate tracing batch size and schedule delay options
--- a/packages/agents-core/src/tracing/processor.ts
+++ b/packages/agents-core/src/tracing/processor.ts
@@ -1,6 +1,7 @@
 import { Span as TSpan } from './spans';
 import { Trace } from './traces';
 import logger, { logModelAndToolActionError } from '../logger';
+import { UserError } from '../errors';
 import {
   timer as _timer,
   isTracingLoopRunningByDefault,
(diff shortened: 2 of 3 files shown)


seratch:
The tests show what happens with invalid or extreme numeric settings, but they do not establish an ordinary configuration path that produces them. Please provide a concrete application configuration and caller-visible failure before expanding runtime validation.

PR #1795, opened 2026-09-01
fix(core): handle processor replacement rejections

sylvesterkaczmarek:
### Summary

Handle rejected shutdown promises when replacing tracing processors.

`MultiTracingProcessor.setProcessors()` is synchronous but invokes each old processor's async `shutdown()` without observing its returned promise. If a custom processor rejects during shutdown, the rejection can surface as unhandled while the replacement processor list has already been installed.

Keep the synchronous replacement contract, attach a rejection handler to each old-processor shutdown, and report failures through the existing redaction-aware tracing error logger. Replacement processors continue to be installed immediately.

### Test plan

- replaces a processor whose `shutdown()` returns a rejected promise
- verifies replacement remains synchronous/non-throwing
- verifies the rejected shutdown is observed and logged
- verifies the replacement processor is installed and receives subsequent trace events

### Issue number

N/A. Found while auditing tracing processor replacement lifecycle.

### Checks

- [x] Documentation change is not required; public replacement semantics are unchanged

### DIFF
--- a/.changeset/handle-tracing-processor-replacement-rejections.md
+++ b/.changeset/handle-tracing-processor-replacement-rejections.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: handle rejected tracing processor shutdowns during processor replacement
--- a/packages/agents-core/src/tracing/processor.ts
+++ b/packages/agents-core/src/tracing/processor.ts
@@ -358,7 +358,13 @@ export class MultiTracingProcessor implements TracingProcessor {
   setProcessors(processors: TracingProcessor[]): void {
     logger.debug('Shutting down old processors');
     for (const processor of this.#processors) {
-      processor.shutdown();
+      void processor.shutdown().catch((error) => {
+        logModelAndToolActionError(
+          logger,
+          'Error shutting down replaced tracing processor',
+          error,
+        );
+      });
     }
     this.#processors = processors;
   }
--- a/packages/agents-core/test/tracing.test.ts
+++ b/packages/agents-core/test/tracing.test.ts
@@ -1624,9 +1624,9 @@ describe('withTrace & span helpers (integration)', () => {
 describe('MultiTracingProcessor', () => {
   it('should call all processors shutdown when setting new processors', () => {
     const processor1 = new TestProcessor();
-    processor1.shutdown = vi.fn();
+    processor1.shutdown = vi.fn(async () => {});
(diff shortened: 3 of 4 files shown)


seratch (changes requested):
The shutdown-rejection path is worth fixing, and the catch belongs at this fire-and-forget boundary. However, handledRejection() already attaches a rejection handler before the SDK receives the Promise, so the current test cannot demonstrate that processor replacement prevents an unhandled rejection.

Please replace that fixture with an actually rejecting processor and verify through the replacement path that the SDK handles the rejection and the replacement remains usable. Keep the implementation scoped to the existing asynchronous processor contract.

sylvesterkaczmarek:
Addressed the review on the latest branch head. The replacement test now uses an actually rejecting `shutdown()` promise, exercises it through `MultiTracingProcessor.setProcessors()`, and verifies the replacement remains installed and usable. The pre-handled rejection fixture is gone.

sylvesterkaczmarek:
Updated the regression per your review: the replaced processor now returns a genuine rejected shutdown promise with no pre-attached handler. The replacement path contains it, logs the failure, and the replacement remains usable. Focused test passes.

seratch:
@codex review again

seratch (changes requested):
Thanks for updating the regression test. The genuine rejected shutdown promise addresses my previous feedback, and the runtime fix looks sound.

One existing fixture still needs updating: in `packages/agents-core/test/tracing.test.ts`, the test "should call all processors shutdown when setting new processors" assigns bare `vi.fn()` shutdown mocks. These return `undefined`, so the new `.catch()` call throws; coverage CI confirms this failure.

Please make both shutdown mocks return resolved promises, for example with `vi.fn(async () => {})`, and rerun the checks. This corrects the fixture to match the existing `Promise<void>` contract. No additional runtime fallback is needed.

sylvesterkaczmarek:
Addressed in `47b1a170`. The two `shutdown` spies in `should call all processors shutdown when setting new processors` now return resolved promises, matching the existing `Promise<void>` contract. Full `packages/agents-core/test/tracing.test.ts` passes 79/79, Prettier and `git diff --check` pass, and the pre-commit TruffleHog scan found no secrets.

sylvesterkaczmarek:
@seratch the last requested fixture update is in 47b1a170: both shutdown mocks now return resolved Promise<void>s, and the focused tracing suite passed 79/79. When convenient, could you re-review the current head?

PR #1796, opened 2026-09-01
fix(core): handle span processor rejections

sylvesterkaczmarek:
### Summary

Handle rejected async tracing-processor hooks invoked from the synchronous `Span.start()` and `Span.end()` APIs.

`TracingProcessor.onSpanStart()` and `onSpanEnd()` return promises, but the span lifecycle methods currently discard those promises. A custom processor rejection can therefore become an unhandled promise rejection even though the caller has no promise to await or catch.

Attach rejection handlers to both fire-and-forget hooks and report failures through the SDK's existing redaction-aware tracing error logger. Span lifecycle timing and the synchronous public API remain unchanged.

### Test plan

- adds a processor whose async span-start and span-end hooks reject
- verifies `Span.start()` and `Span.end()` remain synchronous/non-throwing
- verifies both rejected hook failures are observed and logged by the SDK

### Issue number

N/A. Found while auditing async tracing hooks called from synchronous APIs.

### Checks

- [x] Documentation change is not required; public API behavior is unchanged

### DIFF
--- a/.changeset/handle-span-processor-rejections.md
+++ b/.changeset/handle-span-processor-rejections.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: handle rejected async tracing processor hooks from synchronous span lifecycle methods
--- a/packages/agents-core/src/tracing/spans.ts
+++ b/packages/agents-core/src/tracing/spans.ts
@@ -1,4 +1,4 @@
-import logger from '../logger';
+import logger, { logModelAndToolActionError } from '../logger';
 import { TracingProcessor } from './processor';
 import {
   generateSpanId,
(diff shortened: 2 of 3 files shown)


seratch (changes requested):
The asynchronous span-hook rejection path is worth fixing. The current handledRejection() fixture suppresses the unhandled-rejection signal itself, so the logging assertions do not establish the protection claimed by this PR.

Please use an actually rejecting registered processor and exercise the span lifecycle through the supported tracing path. Verify that the rejection is contained and the surrounding operation completes. Additional fallback behavior for processors that violate the Promise-returning contract is not needed.

sylvesterkaczmarek:
Addressed the review on the latest branch head. The test now registers an actually rejecting processor with `setTraceProcessors()` and exercises it through `withTrace()` / `withCustomSpan()`, verifying the traced operation completes while both span-hook rejections are contained and logged. The pre-handled rejection fixture is gone.

sylvesterkaczmarek:
Updated the regression per your review: the processor now returns genuinely rejected span-hook promises, is registered through the public tracing API, and the test exercises the supported withTrace/withCustomSpan lifecycle while asserting the surrounding operation completes. Focused test passes.

seratch:
@codex review again

seratch (approved):

PR #1797, opened 2026-09-01
fix(core): preserve tracing clone fidelity

sylvesterkaczmarek:
### Summary

Preserve tracing behavior when cloning trace/span objects.

`Trace.clone()` currently reconstructs the clone without the original processor, so a trace created with a custom processor silently switches to the SDK default processor after cloning. The no-op tracing subclasses also lose their no-op semantics when cloned: `NoopTrace` inherits `Trace.clone()`, and `NoopSpan` inherits `Span.clone()`, producing live base-class objects.

This change:
- keeps the original tracing processor on normal `Trace` clones
- keeps cloned `NoopTrace` instances as `NoopTrace`
- keeps spans with no-op trace/span IDs as `NoopSpan`
- preserves existing clone IDs, metadata, timestamps, errors, and tracing API-key behavior

### Test plan

- verifies a cloned trace continues dispatching lifecycle events to its custom processor
- verifies a cloned `NoopTrace` remains a no-op and serializes to `null`
- verifies a cloned `NoopSpan` remains a no-op and cannot emit span lifecycle callbacks

### Issue number

N/A. Found while auditing tracing clone fidelity.

### Checks

- [x] Documentation change is not required; this restores existing tracing semantics

### DIFF
--- a/.changeset/preserve-tracing-clones.md
+++ b/.changeset/preserve-tracing-clones.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve tracing processor and no-op behavior when cloning traces and spans
--- a/packages/agents-core/src/tracing/spans.ts
+++ b/packages/agents-core/src/tracing/spans.ts
@@ -273,6 +273,13 @@ export class Span<TData extends SpanData> {
   }
 
   clone(): Span<TData> {
+    if (
+      this.traceId === NOOP_TRACE_OR_SPAN_ID ||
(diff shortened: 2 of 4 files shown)


seratch:
The clone differences are real, but the current tests do not establish the practical impact of this combined change. Please show a supported tracing workflow, preferably a run/resume case using a custom processor, where the current code loses a required lifecycle notification or export and this patch restores it.

Please assess the no-op changes separately. The normal disabled-run paths already clear no-op trace state, so class identity and toJSON differences alone do not demonstrate unintended export. I am withholding a merge recommendation until the affected consumer behavior is established.

PR #1806, opened 2026-09-04
fix(core): preserve CRLF line endings in applyDiff

### LINKED ISSUE #1804: applyDiff mixes LF into CRLF files
## Summary

`applyDiff()` preserves existing CRLF lines but inserts new lines with LF only, producing a mixed-line-ending file after an ordinary patch.

The helper is publicly exported and is also used by Unix-local and Docker sandbox patch application, so this affects Windows-style files edited through the SDK.

## Reproduction

On current `main` at `8e862b3380a577df1315bef17f351c1b58c2938b`:

```ts
const input = 'alpha\r\nbeta\r\n';
const diff = '@@\n alpha\n+inserted\n beta';
const result = applyDiff(input, diff);

console.log(JSON.stringify(result));
```

Actual:

```text
"alpha\r\ninserted\nbeta\r\n"
```

Expected:

```text
"alpha\r\ninserted\r\nbeta\r\n"
```

The resulting newline sequence is therefore CRLF, LF, CRLF.

## Root cause

Update-mode input is split on `\n`, which leaves the existing `\r` attached to CRLF lines. Diff lines are normalized to LF and inserted content has no `\r`. `applyChunks()` then joins everything with `\n`, so retained lines reconstruct as CRLF while inserted lines reconstruct as LF.

## Expected behavior

When an existing file consistently uses CRLF, inserted/replacement lines should use CRLF as well. Existing LF behavior and create-file behavior should remain unchanged.

sylvesterkaczmarek:
### Summary

Preserve CRLF line endings when `applyDiff()` inserts or replaces lines in an existing CRLF file.

Update-mode CRLF input is normalized only for logical patch matching and then reconstructed with the original CRLF separator. Files using LF or mixed line endings retain their existing behavior, create-file mode is unchanged, and files without a final newline preserve that state.

### Concrete downstream workflow

`applyDiff()` is used by the sandbox/editor paths to rewrite files on disk. A CRLF-enforced project can then run a formatter check on the patched file. Before this fix, inserting a line into an otherwise CRLF TypeScript file produces mixed line endings, so a Prettier check configured with `endOfLine: 'crlf'` rejects the result even though the source file passed that policy before the patch.

Added `applyDiff.crlf-workflow.test.ts` to exercise that workflow directly: patch a CRLF TypeScript file, then require Prettier's CRLF formatting check to accept the resulting content.

### Test plan

- existing `packages/agents-core/test/utils/applyDiff.test.ts` coverage includes CRLF insertion/replacement, EOF changes without a final newline, mixed endings, and existing patch behavior
- new workflow regression checks the patched CRLF TypeScript result with Prettier using `endOfLine: 'crlf'`
- the previous head passed `.agents/skills/code-change-verification/scripts/run.sh`, the full suite, formatting, lint, and pre-commit checks; pushed-head CI is required for the added workflow regression

### Issue number

Closes #1804

### Checks

- [x] I've run `pnpm test` and `pnpm test:examples` on the previous head
- [x] I've run `.agents/skills/code-change-verification/scripts/run.sh` on the previous head
- [ ] I've confirmed all verification steps pass on the current head; awaiting pushed-head CI

### DIFF
--- a/.changeset/preserve-crlf-patches.md
+++ b/.changeset/preserve-crlf-patches.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix: preserve CRLF line endings when applyDiff inserts or replaces lines in CRLF files. (#1804)
--- a/packages/agents-core/src/utils/applyDiff.ts
+++ b/packages/agents-core/src/utils/applyDiff.ts
@@ -17,8 +17,23 @@ export function applyDiff(
     return parseCreateDiff(diffLines);
   }
 
-  const { chunks } = parseUpdateDiff(diffLines, input);
-  return applyChunks(input, chunks);
+  const lineEnding = updateLineEnding(input);
+  const normalizedInput =
+    lineEnding === '\r\n' ? input.replace(/\r\n/g, '\n') : input;
+  const { chunks } = parseUpdateDiff(diffLines, normalizedInput);
+  return applyChunks(normalizedInput, chunks, lineEnding);
+}
+
+function updateLineEnding(input: string): '\n' | '\r\n' {
+  let hasNewline = false;
+  for (let index = 0; index < input.length; index += 1) {
+    if (input[index] !== '\n') continue;
+    hasNewline = true;
+    if (index === 0 || input[index - 1] !== '\r') {
+      return '\n';
+    }
+  }
+  return hasNewline ? '\r\n' : '\n';
 }
 
 type Chunk = { origIndex: number; delLines: string[]; insLines: string[] };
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Thanks for the contribution. The reproduction confirms that an ordinary patch introduces LF line endings into a CRLF file, and the implementation appears narrowly scoped.

Before we take this on, could you add a concrete workflow where the mixed endings cause an incorrect result, such as a file consumer misreading the output or a required line-ending check rejecting it? The current report establishes the byte-level difference, but not its downstream consequence. That evidence would help justify making CRLF preservation part of the shared helper’s behavior.

sylvesterkaczmarek:
Added the concrete downstream case you asked for. The new regression patches a CRLF TypeScript file through `applyDiff()` and then runs Prettier's `endOfLine: 'crlf'` check against the result, which is the kind of required formatting workflow the pre-fix mixed output breaks. I also updated the PR description to make that consequence explicit.

seratch (approved):

PR #1813, opened 2026-09-04
fix(core): wait for in-flight trace exports in forceFlush

### LINKED ISSUE #1812: BatchTraceProcessor.forceFlush can return while an export is still in flight
### Describe the bug

`BatchTraceProcessor.forceFlush()` only checks whether the internal buffer is non-empty. A scheduled or threshold-triggered export removes its batch from that buffer before awaiting the exporter, so `forceFlush()` can observe an empty buffer and resolve immediately even though the export is still running.

This makes an explicit flush an unreliable completion barrier: a caller can `await forceFlush()` and then terminate a worker/request while the trace export it intended to flush is still pending.

### Reproduction

A deterministic test can configure a low export trigger and a large batch size so the second trace drains the whole buffer into a delayed exporter:

1. Buffer the first trace.
2. Add a second trace, triggering an export that removes both items from the buffer and then blocks inside `exporter.export()`.
3. Wait until the exporter has started.
4. Call `forceFlush()` while the export is still blocked.

On current `main`, `forceFlush()` resolves before the exporter is released.

### Expected behavior

`forceFlush()` should not resolve until exports that were already in flight when the flush began have settled. It should preserve existing exporter-error handling and must not stop the scheduled export loop.

`shutdown()` already waits while `#exportInProgress` is true; the gap is specific to the non-terminal `forceFlush()` lifecycle.

sylvesterkaczmarek:
## Summary

Make `BatchTraceProcessor.forceFlush()` wait for trace exports that were already in flight when the flush began.

A scheduled or threshold-triggered export removes its batch from the internal buffer before awaiting the exporter. Previously, `forceFlush()` could therefore observe an empty buffer and return even though that export was still pending.

The processor now tracks active export promises and awaits the ones already running when `forceFlush()` starts. The normal scheduled export loop, exporter error handling, and shutdown behavior are unchanged.

## Tests

- Added a deterministic regression that drains the buffer into a blocked exporter, calls `forceFlush()`, and verifies the flush does not resolve until the exporter is released.
- `packages/agents-core/test/tracing.test.ts` — 80 passed.
- Agents-core test TypeScript, ESLint, Prettier, and `git diff --check` pass.
- `.agents/skills/code-change-verification/scripts/run.sh` passed end to end: install, build, package build/dist checks, lint, full tests, and changed-file formatting.
- Pre-commit formatting, lint, and secret scan passed.

Fixes #1812.

### DIFF
--- a/.changeset/wait-for-inflight-trace-flush.md
+++ b/.changeset/wait-for-inflight-trace-flush.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix(core): wait for in-flight trace exports during force flush (#1812)
--- a/packages/agents-core/src/tracing/processor.ts
+++ b/packages/agents-core/src/tracing/processor.ts
@@ -134,6 +134,7 @@ export class BatchTraceProcessor implements TracingProcessor {
   #exportInProgress = false;
   #timeoutAbortController: AbortController | null = null;
   #activeExportAbortControllers = new Set<AbortController>();
+  #activeExportPromises = new Set<Promise<void>>();
 
   constructor(
     exporter: TracingExporter,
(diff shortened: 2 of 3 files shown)


seratch (changes requested):
Before merging, please add focused concurrency coverage: hold an earlier export open while forceFlush() starts a buffered batch, reject one export while the other remains pending, and verify that flushing waits for the survivor and subsequent exports still work. Please also cover timed shutdown aborting an export while forceFlush() is waiting, using an exporter that settles on cancellation, and assert that both operations finish.

The current regression covers the empty-buffer success case; these additions would cover failure and cancellation through the new promise tracking.

sylvesterkaczmarek:
Added the requested concurrency coverage in `7cfdbe2`. One regression keeps an earlier export pending while `forceFlush()` starts a buffered batch that fails, verifies flush still waits for the surviving export, then verifies a later export succeeds. The second runs `forceFlush()` alongside timed shutdown and uses an exporter that settles on cancellation, verifying both operations complete. The full tracing test file passes 82/82; formatting and diff checks pass.

seratch (changes requested):
Thanks for adding the coverage. The timed-shutdown test addresses that part of the review.

The concurrent-failure test still checks `flushSettled` too early: `secondAttempted` signals exporter entry, and one microtask turn does not let the failed batch unwind through `forceFlush()` and its `.then()` handler. The assertion can therefore pass on the original implementation before `releaseFirst()` is called.

Please keep the first export blocked, yield through an event-loop task boundary, and then assert that flushing remains pending. Verify that this regression fails without the in-flight waiting logic and passes with it.

sylvesterkaczmarek:
Updated the concurrent-failure regression in `4f6eb16` to yield across a real event-loop task boundary before checking `flushSettled`. I also verified the test has teeth: it passes on the current implementation, fails when the in-flight wait is temporarily removed, then passes again after restoring the production code. Full tracing file: 82/82; formatting and diff checks pass.

seratch (approved):

PR #1815, opened 2026-09-04
fix(core): stop stale trace export loops after shutdown

### LINKED ISSUE #1814: Timed tracing shutdown can re-arm the export loop after shutdown returns
### Describe the bug

`BatchTraceProcessor.shutdown(timeout)` can return after its timeout while a scheduled export is still finishing. If the exporter ignores the abort signal and settles later, the already-running scheduled callback continues with `#runExportLoop()` and creates a new timer after shutdown has completed.

That leaves the tracing export loop alive after a terminal shutdown call.

### Deterministic reproduction

Using fake timers and a custom exporter that deliberately blocks and ignores the abort signal:

1. Create `BatchTraceProcessor` with a short `scheduleDelay`.
2. Queue a trace and advance to the scheduled export.
3. Keep `exporter.export()` blocked.
4. Call `shutdown(5)` and advance the timeout so shutdown returns.
5. Release the exporter.
6. Flush microtasks.

On current `main`, a new scheduled timer appears after step 5 even though shutdown has already returned.

The race comes from the scheduled callback:

```ts
this.#timeout = this.#timer.setTimeout(async () => {
  await this.#exportBatches();
  this.#runExportLoop();
}, this.#scheduleDelay);
```

`shutdown()` clears the timer that exists when its `finally` block runs, but an export callback that finishes later can call `#runExportLoop()` afterwards.

### Expected behavior

After `shutdown()` returns, an already-running export may finish, but it must not re-arm the periodic export loop. An explicit later `start()` can still be allowed to start the loop again.

sylvesterkaczmarek:
## Summary

Prevent a scheduled `BatchTraceProcessor` callback from re-arming the export timer after a timed shutdown has already returned.

The export loop now carries a lifecycle generation. `start()` creates a new generation and `shutdown()` invalidates the current one. Scheduled callbacks verify that their generation is still active before exporting or scheduling the next timer, so an exporter that settles after shutdown cannot resurrect the loop.

An explicit later `start()` still starts a fresh export loop.

## Tests

- Added a deterministic fake-timer regression covering timed shutdown while an exporter ignores cancellation and settles later.
- Added coverage proving an explicit restart after shutdown still exports normally.
- Agents-core TypeScript, ESLint, Prettier, and `git diff --check` pass.
- `.agents/skills/code-change-verification/scripts/run.sh` passed end to end: install, build, package build/dist checks, lint, full tests, and changed-file formatting.
- Pre-commit formatting, lint, and secret scan passed.

Fixes #1814.

### DIFF
--- a/.changeset/stop-stale-trace-loop.md
+++ b/.changeset/stop-stale-trace-loop.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+fix(core): stop stale trace export loops from re-arming after shutdown (#1814)
--- a/packages/agents-core/src/tracing/processor.ts
+++ b/packages/agents-core/src/tracing/processor.ts
@@ -134,6 +134,7 @@ export class BatchTraceProcessor implements TracingProcessor {
   #exportInProgress = false;
   #timeoutAbortController: AbortController | null = null;
   #activeExportAbortControllers = new Set<AbortController>();
+  #exportLoopGeneration = 0;
 
   constructor(
     exporter: TracingExporter,
@@ -160,8 +161,9 @@ export class BatchTraceProcessor implements TracingProcessor {
   }
 
   start(): void {
+    const generation = ++this.#exportLoopGeneration;
     this.#timeoutAbortController = new AbortController();
-    this.#runExportLoop();
+    this.#runExportLoop(generation);
   }
 
   async #safeAddItem(item: Trace | Span): Promise<void> {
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for the clear reproduction. It shows that a delayed exporter can re-arm the timer after timed shutdown, but the current evidence stops at the timer count. The timer is unref'ed on Node, so that alone does not establish a process-lifetime problem, and the report does not yet show lost or duplicated traces or meaningful resource accumulation.

I am closing this PR and #1814 for now rather than adding a restart-generation contract for the timer-only observation. A concrete supported workflow with a measurable downstream consequence would justify reopening the work. The forceFlush trace-loss issue in #1812 remains a separate fix.

PR #1824, opened 2026-09-04
fix(openai): reset compaction state after pop

### LINKED ISSUE #1823: Popping compaction session history retains the previous response chain
### Describe the bug

`OpenAIResponsesCompactionSession.popItem()` removes the latest item from the local session history but leaves the wrapper's retained `responseId` and `lastStore` unchanged.

That leaves two representations of the conversation out of sync: the local session no longer contains the popped item, while the retained Responses `previous_response_id` still refers to server-side history that includes it. A later compaction in `previous_response_id` mode can therefore compact the stale server chain and replace the local history with output derived from content the caller explicitly removed.

### Reproduction

```ts
const compact = vi.fn();
const underlyingSession = new MemorySession({
  initialItems: [
    {
      type: 'message',
      role: 'assistant',
      status: 'completed',
      content: [{ type: 'output_text', text: 'old' }],
    },
  ],
});
const session = new OpenAIResponsesCompactionSession({
  client: { responses: { compact } } as any,
  underlyingSession,
  shouldTriggerCompaction: () => false,
});

await session.runCompaction({ responseId: 'resp_old', store: true });
await session.popItem();
await session.runCompaction({
  force: true,
  compactionMode: 'previous_response_id',
});
```

On current `main`, the final call proceeds with `previous_response_id: 'resp_old'` even though the local history covered by that response ID has been mutated by `popItem()`.

### Expected behavior

A successful `popItem()` that actually removes an item should invalidate retained response-chain state. A later `previous_response_id` compaction should require a newly supplied response ID before making any client call. If `popItem()` returns no item, existing response-chain state should remain unchanged.

This is the `popItem()` counterpart to #1821: both operations mutate durable local history, but `popItem()` has its own success/no-op boundary and needs separate regression coverage.

sylvesterkaczmarek:
## Summary

Invalidate retained Responses compaction chain state after `OpenAIResponsesCompactionSession.popItem()` actually removes local history.

Previously, `popItem()` updated the local session and compaction caches but kept `responseId` and `lastStore`. A later `previous_response_id` compaction could therefore reuse server-side history that still contained the item the caller had removed locally.

The response-chain state now resets immediately after a successful pop, before any later cache refresh can fail. A no-op pop preserves the existing chain, and a newly supplied response ID establishes a fresh chain normally.

## Tests

- `openaiResponsesCompactionSession.test.ts`: 36 passed.
- Full repository code-change verification passed end to end on the final diff.
- Full suite: 208 test files, 6,447 tests passed.
- All build, dist, lint, formatting, changeset, and pre-commit checks passed.

Fixes #1823.

### DIFF
--- a/.changeset/reset-compaction-pop-state.md
+++ b/.changeset/reset-compaction-pop-state.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix(openai): reset compaction response state after popping local history (#1823)
--- a/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
+++ b/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
@@ -285,6 +285,8 @@ export class OpenAIResponsesCompactionSession
       if (!popped) {
         return popped;
       }
+      this.responseId = undefined;
+      this.lastStore = undefined;
       if (this.sessionItems) {
         const index = this.sessionItems.lastIndexOf(popped);
         if (index >= 0) {
--- a/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
+++ b/packages/agents-openai/src/memory/openaiResponsesCompactionSession.ts
@@ -296,28 +296,12 @@ export class OpenAIResponsesCompactionSession
       }
       this.responseId = undefined;
       this.lastStore = undefined;
-      if (this.sessionItems) {
-        const index = this.sessionItems.lastIndexOf(popped);
-        if (index >= 0) {
-          this.sessionItems.splice(index, 1);
-        } else {
-          this.sessionItems = await this.underlyingSession.getItems();
-        }
-      }
-      if (this.compactionCandidateItems) {
-        const isCandidate = selectCompactionCandidateItems([popped]).length > 0;
-        if (isCandidate) {
-          const index = this.compactionCandidateItems.indexOf(popped);
-          if (index >= 0) {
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
Addressed the ambiguous pop failure and rebased onto current main. If the backing session can commit a pop and then lose the acknowledgement, the wrapper now invalidates response-chain state plus both cached history views before rethrowing. Added a regression that commits-then-rejects, verifies the old response ID is unusable, and verifies the next compaction decision reloads the actually persisted history. Focused pop/clear tests pass (5/5).

seratch (changes requested):
Thanks for addressing the ambiguous pop failure. One related path still allows removed context to return: when the backing pop succeeds but the subsequent getItems() refresh rejects, both old history caches remain populated. Since the response ID has now been cleared, a later automatic compaction uses those stale items as input.

Please invalidate both history caches immediately after a successful nonempty pop and let the next compaction reload them. Add a regression using a cloning backend whose first post-pop history read fails, then verify that recovery builds the compaction request from the remaining persisted history without the removed item.

sylvesterkaczmarek:
Addressed in `52fab15`. After any successful non-empty pop, both cached history views are now invalidated immediately, so a cloned return value or failed post-pop refresh cannot leave removed context cached. Added the requested cloning-backend regression: the first authoritative history reload fails, a retry reloads the persisted remainder, and the compact request contains only the surviving item. Focused compaction-session tests pass 39/39; formatting and diff checks pass.

seratch (approved):

PR #1825, opened 2026-09-04
fix(openai): preserve WebSocket auth query parameters

rioyu123:
### Summary

This pull request fixes WebSocket authentication query conversion when parameter names collide with properties inherited from `Object.prototype`.

The previous plain-object accumulator could fail to preserve parameters such as `__proto__` and incorrectly combine parameters such as `constructor`, causing the query supplied to custom `authHeaders()` implementations to differ from the actual WebSocket handshake URL.

Query parameters are now accumulated in a `Map` and materialized with `Object.fromEntries()`. This preserves prototype-sensitive keys and repeated values while retaining the existing ordinary-object callback shape and empty-query behavior.

Regression coverage exercises both the shared conversion helper and the WebSocket handshake path, including repeated and prototype-sensitive parameter names.

### Test plan

- Confirmed the WebSocket regression fails with the baseline helper and passes with the fix.
- Focused helper, WebSocket, and hosted multi-agent tests: 3 files, 119 tests passed.
- Ran `.agents/skills/code-change-verification/scripts/run.sh` on Linux against the exact final files: install, build, workspace typechecks, distribution checks, lint, full tests, and formatting passed. Full test suite: 209 files, 6,447 tests passed.

### Checks

- [x] I've added new tests.
- [x] I've added a patch changeset for `@openai/agents-openai`.
- [x] I've run `pnpm test` and the `pnpm test:examples` equivalent, `pnpm -r build-check`.
- [x] I've run `.agents/skills/code-change-verification/scripts/run.sh`.
- [x] I've confirmed all verification steps pass on Linux.
- [ ] I've added/updated the relevant documentation (not applicable: no API or usage change).
- [ ] I've run `pnpm test:integration` (not applicable to this focused fix).
- [ ] If using Codex, I've run `/review` before submitting this PR (independent Claude reviews completed instead).

### DIFF
--- a/.changeset/preserve-websocket-auth-query.md
+++ b/.changeset/preserve-websocket-auth-query.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: preserve all WebSocket authentication query parameters
--- a/packages/agents-openai/src/responsesUtils.ts
+++ b/packages/agents-openai/src/responsesUtils.ts
@@ -13,22 +13,22 @@ export function normalizeInstructions(
 export function searchParamsToAuthHeaderQuery(
   searchParams: URLSearchParams,
 ): Record<string, string | string[]> | undefined {
-  const query: Record<string, string | string[]> = {};
+  const query = new Map<string, string | string[]>();
   let hasEntries = false;
 
   for (const [key, value] of searchParams.entries()) {
     hasEntries = true;
-    const existingValue = query[key];
+    const existingValue = query.get(key);
     if (typeof existingValue === 'undefined') {
-      query[key] = value;
+      query.set(key, value);
     } else if (Array.isArray(existingValue)) {
       existingValue.push(value);
     } else {
-      query[key] = [existingValue, value];
+      query.set(key, [existingValue, value]);
     }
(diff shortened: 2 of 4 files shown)


sylvesterkaczmarek (commented):
Moving accumulation to `Map` fixes the actual semantic bug rather than special-casing `__proto__`: key lookup no longer consults `Object.prototype`, repeated values remain ordered, and `Object.fromEntries` preserves the existing plain-object callback contract. The handshake regression is especially useful because it proves `authHeaders()` sees the same query semantics as the eventual WebSocket URL. Empty-query behaviour also stays unchanged. I don’t see a blocker here.

seratch:
The helper behavior is reproducible, but the PR does not establish an affected authentication integration. Please identify a real provider or signed URL that requires these query names and show the failing connection. Without that evidence, I would close this audit-only change.

rioyu123:
Understood. The current reproduction demonstrates query-conversion behavior, not an affected authentication integration. I'll leave this closed. If I encounter a provider or signed URL that actually issues these query names and a connection that fails because of them, I'll bring that here rather than reopening on the existing tests.

PR #1835, opened 2026-09-05
chore: run repository skill tests in a dedicated CI lane

seratch:
This pull request adds pnpm test:repo-skills and a dedicated CI workflow covering existing Python and Node skill tests. It isolates child credentials and temporary files, verifies failure propagation and cleanup, and includes three offline changeset-validator fixtures.


sylvesterkaczmarek (commented):
This is a useful CI boundary; these repository-skill tests were otherwise easy to miss because they sit outside the normal package test discovery. The environment isolation is the strongest part: credentials and `PYTHONPATH` are removed and Git network access is disabled, so a green run is much less likely to depend on developer-machine state.

One thing I would make explicit is the treatment of `CI` / `GITHUB_ACTIONS`. The child environment whitelist currently drops them. If the skills are intentionally being tested as environment-neutral tools, a short comment near the whitelist would document that contract. If any skill has CI-specific behavior, this lane will not exercise that branch.

PR #1851, opened 2026-09-08
fix(core): avoid duplicate computer screenshots

### LINKED ISSUE #1735: Computer screenshot action captures twice before returning observation
### Describe the bug

When a computer tool call contains an explicit `screenshot` action, the Agents JS runner invokes `computer.screenshot()` for the action and then invokes it again unconditionally to build the returned computer observation.

In `packages/agents-core/src/runner/toolExecution.ts`, `_runComputerActionAndScreenshot()` currently does both:

1. `case 'screenshot': await computer.screenshot(runContext)`
2. after all actions, `await computer.screenshot(runContext)` again

The first screenshot result is discarded.

### Reproduction

Use a `computerTool` whose `screenshot()` mock increments a counter and returns a different value for each call, then execute a `computer_call` with:

```ts
action: { type: 'screenshot' }
```

The existing `executeComputerActions` path calls `screenshot()` twice. The returned `computer_call_result` contains the second capture rather than the capture produced by the requested screenshot action.

### Expected behavior

A standalone screenshot action should capture once and reuse that capture as the returned observation. Non-screenshot actions should continue to take a fresh screenshot after the action.

This also matches the Python Agents SDK behavior, which reuses the explicit screenshot when it is already the final observation.

### Why it matters

Duplicate capture adds unnecessary latency/cost for remote computer implementations and can return a different screen state from the screenshot action the model requested.

WallaceScott240:
## Summary

Avoid duplicate screenshots when a computer tool action is itself a `screenshot`.

Previously, an explicit screenshot action called `computer.screenshot()` during action execution and then captured another screenshot immediately afterward for the tool result.

This change preserves the screenshot returned by an explicit screenshot action and reuses it as the final observation when appropriate.

## Changes

- reuse the screenshot returned by an explicit `screenshot` action
- continue capturing a final screenshot for non-screenshot computer actions
- preserve correct final-observation semantics for batched actions
- add regression coverage for:
  - `[screenshot, action]`
  - `[action, screenshot]`
- update existing expectations that encoded the previous double-capture behavior

## Validation

- full repository code-change verification passed
- 220 test files passed
- 6705 tests passed
- 2 skipped
- lint/build/type checks passed

Fixes #1735

### DIFF
--- a/.changeset/plenty-moles-wait.md
+++ b/.changeset/plenty-moles-wait.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-core': patch
+---
+
+Avoid duplicate screenshots for explicit computer screenshot actions.
--- a/packages/agents-core/src/runner/toolExecution.ts
+++ b/packages/agents-core/src/runner/toolExecution.ts
@@ -1523,10 +1523,15 @@ async function _runComputerActionAndScreenshot(
   runContext: RunContext,
   signal?: AbortSignal,
 ): Promise<{ type: 'completed'; output: string } | { type: 'cancelled' }> {
+  let screenshot: string | undefined;
+
   for (const action of getComputerToolActions(toolCall)) {
     if (signal?.aborted) {
       return { type: 'cancelled' };
     }
+
+    screenshot = undefined;
+
     switch (action.type) {
       case 'click':
         await computer.click(action.x, action.y, action.button, runContext);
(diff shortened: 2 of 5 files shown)


seratch:
Thanks for the contribution and regression coverage. The duplicate capture problem in #1735 is worth fixing, but #1827 already implements the simpler direction requested in the earlier reviews: treat screenshot actions as no-ops during batch dispatch and capture the final screen once afterward.

This PR restores screenshot caching and invalidation, which still captures and discards an intermediate image for [screenshot, click]. The official computer-use handlers use the single-final-capture approach, and we have not established a need for intermediate captures.

I am going to close this PR in favor of #1827 so we can complete one implementation.

PR #1887, opened 2026-09-12
chore: migrate the memory example to Prisma 7

jbeckwith-oai:
### Summary

Updates the private memory example to Prisma 7 with matching CLI, client, and SQLite adapter dependencies. Moves the connection URL into `prisma.config.ts`, preserves schema-relative database paths and existing millisecond timestamps, keeps SQLite query parameters out of the native filename while translating `socket_timeout`, and generates the real client before typechecking instead of using fallback declarations. Generated-client loading is deferred so file-backed examples also work before Prisma setup. This completes the coordinated migration proposed by #1881.

Preserves Prisma 6's transaction lock waiting with a version-pinned patch to both better-sqlite3 adapter distributions. With an external writer holding the database for 300 ms and `socket_timeout=1`, the unpatched Prisma 7 transaction failed immediately; it now waits and succeeds. `BEGIN IMMEDIATE` acquires the write lock before transaction reads, and a failed begin releases the adapter mutex so the same client remains usable after timeout. When an external reader instead blocks COMMIT after INSERT/DELETE, adapter rollback checks the native transaction state and executes SQL ROLLBACK before releasing the mutex. The rejected operation retains committed history, releases write ownership, and leaves both a second connection and the same client usable. The Session API, schema, and record encoding are unchanged.

The 15 Prisma memory regression tests now run after example compilation in both existing Node 22 and Node 24.3 CI jobs. The existing path filter and workflow permissions are unchanged.

The native `better-sqlite3` dependency requires an explicit build allowance. Please include maintainer security review of the dependency/install boundary and the pinned adapter patch. Recheck the contention tests before removing the patch on an adapter upgrade.

### Validation

Rebased onto `7f135afc` after the upstream dependency update. Only the lockfile needed resolution; all example code, regression tests and the adapter patch retain their delivered bytes. Upstream dependency entries are preserved. The focused checks, two fresh review rounds, full stack and actual example startup were rerun on the integrated result. The subsequent failed-COMMIT cleanup also passed two new independent review rounds, all 15 focused tests, the complete verification stack, and the actual example E2E harness.

- All 15 focused example tests pass, including real subprocess contention for append/pop at BEGIN and COMMIT, timeout atomicity, second-connection access and same-client recovery, Prisma 6 storage, query URLs, and FileSession imports before generation.
- Differential Prisma 6 / patched Prisma 7 tests match for short-lock success and one-second timeout, with retained history and recovery in both transaction modes.
- Before the source-preserving rebase, 4,000 modeled operations and 5,454 assertions pass across macOS Node 22 and a pristine Linux Node 24 pnpm installation: transaction modes, reconnect, large batches, Unicode, path/query variants, trigger-induced failures, injected-client ownership, corrupt databases, and external locks. Fresh and frozen installs apply the patch; installed CJS/ESM bytes and the lockfile patch hash match.
- Actual example startup passes with a local mock Responses API: streaming, non-streaming, restart, default/query database URLs, and persisted history forwarding. No live model calls.
- Two consecutive clean independent review rounds cover the complete diff, including transaction lifecycle and dependency security.
- The full repository verification stack passes on the reviewed content: frozen install, build, workspace typechecks, declaration checks, lint, formatting, and all 6,905 tests in 225 files. Docker tests use local Colima and a host-mounted temporary directory.

No changeset: only a private example and its dependency/install and CI configuration change. Live-provider suites are outside this SQLite migration; the actual example is exercised by the local API harness.


examples/memory/prisma-database.ts
```diff
@@ -0,0 +1,9 @@
+import { resolve } from 'node:path';
+
+// Prisma 6 resolves SQLite file URLs relative to the schema directory. Keep the
+// CLI and driver adapter on that same database, regardless of the caller's cwd.
+export function resolvePrismaDatabaseUrl(databaseUrl: string): string {
+  return databaseUrl.startsWith('file:')
+    ? `file:${resolve(__dirname, 'prisma', databaseUrl.slice('file:'.length))}`
```
HAYDEN-OAI: [P2] Keep SQLite URL parameters out of the adapter filename

With `DATABASE_URL=file:./dev.db?socket_timeout=5`, this preserves the query suffix in the resolved URL. The normal `start:prisma` setup still uses Prisma's native schema engine, which separates the query and prepares `prisma/dev.db`; the new better-sqlite3 adapter only removes `file:` and opens the remaining string literally. Session operations therefore open a separate `prisma/dev.db?socket_timeout=5` and fail because its `Session` table is missing, leaving the existing history inaccessible. This is a previously supported SQLite setting: [Prisma 6.19's connection documentation](https://github.com/prisma/prisma-engines/blob/2ba551f319ab1df4bc874a89965d8b3641056773/quaint/src/pooled.rs#L15-L36) explicitly defines query parameters and SQLite `socket_timeout`. Separate the database path from these parameters when configuring the adapter; any intentionally unsupported legacy settings should instead produce an explicit migration error before database setup, not select another file.

HAYDEN-OAI (commented):
Reviewed the complete 13-file diff at `351ebec376a2ae27664a42974b9cc39e59a84cd6`, including configuration/startup paths, persisted Prisma 6 records, generated-client import behavior, the manifest/lockfile update, and the native install allowance. One P2 compatibility issue is left inline: a SQLite connection URL with query parameters makes database setup and session operations target different files.

The existing Session record encoding, millisecond timestamp format, and supplied-client behavior otherwise look preserved. The 30 non-memory workspace dependency closures are unchanged.

Static review only; no installation, build, test, CI, mergeability, database, or native-code execution. Dependency trust limits remain: Prisma CLI 7.10.0 has no advertised npm provenance attestation, and npm lockfile integrity does not authenticate better-sqlite3 binaries fetched separately during installation. Those artifact trust chains were not verified by this review.

examples/memory/prisma-database.ts
jbeckwith-oai: Fixed in 6ddb80f45e68657114437b3fbad7ec0dd03023e1. CLI configuration keeps the query string, while the better-sqlite3 adapter receives only the resolved file path. The factory also maps socket_timeout from seconds to the native millisecond timeout. The Prisma 6 fixture test now uses a relative URL with socket_timeout=7, verifies retained history and append/reconnect, and checks PRAGMA busy_timeout equals 7000. That test reproduced the missing-table failure before the fix and passes afterward. All nine Prisma tests, actual streaming/non-streaming/restart startup, and the full 6,889-test verification stack pass locally after rebasing onto the Vitest 5 main branch.

jbeckwith-oai:
Fixed the confirmed transaction-contention regression in c96f38e5bf0181e72c547896576d868323768b0a. The version-pinned adapter patch restores Prisma 6's `BEGIN IMMEDIATE` behavior and releases the adapter mutex on a failed begin. A 300 ms external write lock now waits and succeeds; a lock beyond `socket_timeout=1` times out without partial changes, and the same client works again after release.

Added four subprocess regression cases covering append/pop wait and timeout recovery. All 13 focused tests pass, alongside 4,000 modeled operations / 5,454 assertions on macOS Node 22 and pristine Linux Node 24, the Prisma 6 comparison, actual example startup, two clean independent review rounds, and the full 6,889-test repository stack. CI is running on the pushed commit.

examples/memory/package.json
```diff
@@ -17,9 +18,10 @@
     "start:oai-compact-stateless": "tsx oai-compact-stateless.ts",
     "start:file": "tsx file.ts",
     "start:file-hitl": "tsx file-hitl.ts",
-    "start:prisma": "tsx start-prisma.ts"
+    "start:prisma": "tsx start-prisma.ts",
+    "test:prisma": "prisma generate && vitest run --config vitest.prisma.config.ts"
```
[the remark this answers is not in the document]
jbeckwith-oai: Fixed in 73d4faab1c83567990e9c6231dae0578a5dd6040. Added `pnpm --filter memory test:prisma` immediately after example compilation in both existing Node 22 and Node 24.3 matrix jobs, under the existing `run_ci` gate. The relevant example, patch, dependency and workflow paths already activate that gate. Verified the actual CI step succeeded in both jobs: each ran all 13 Prisma tests, followed by all 6,905 SDK tests. All current-head checks are green.

Validation: https://github.com/openai/openai-agents-js/actions/runs/34712304748

examples/memory/sessions/prisma.ts
```diff
@@ -148,8 +152,27 @@ export async function createPrismaSession(
...
     }
+    const [databaseUrl, query] = resolvePrismaDatabaseUrl(
+      process.env.DATABASE_URL!,
+    ).split('?');
+    const socketTimeout = new URLSearchParams(query).get('socket_timeout');
+    // File-backed sessions share this module's exports and do not need a
+    // generated Prisma client. Load it only when creating a Prisma session.
+    const { PrismaClient } = await import('@prisma/client');
+    prisma = new PrismaClient({
+      adapter: new PrismaBetterSqlite3(
```
markstuart-oai, line 163: [P1] Roll back the SQLite connection after a failed COMMIT

With the normal rollback-journal database, an external reader can let `BEGIN IMMEDIATE` and the append/delete succeed but keep `COMMIT` blocked until `socket_timeout`. SQLite then leaves the transaction active ([transaction semantics](https://www.sqlite.org/lang_transaction.html#implicit_versus_explicit_transactions)). In Prisma 7.10, the failed-COMMIT branch calls only `transaction.rollback()` and removes the transaction from its active map; this adapter’s `rollback()` only releases its mutex, and the added patch does not change that method. The rejected Session operation therefore leaves an open write transaction on the connection: subsequent transactional appends/pops fail with “cannot start a transaction within a transaction”, and other writers remain blocked until the connection is closed. Extend the existing adapter patch to issue SQL `ROLLBACK` whenever the connection is still in a transaction before releasing its mutex, and cover reader-blocked COMMIT with retained-history and same-client recovery checks. The current external-writer tests exercise failure at BEGIN and do not reach this path.

markstuart-oai (commented):
Reviewed the complete diff at `73d4faab`, including the updated CI step, persisted Prisma 6 data, native adapter selection and version-pinned patch, with independent reviewers. One substantive transaction-cleanup issue remains inline, so I am leaving feedback without approval. The current patch repairs BEGIN contention but does not clean up a failed COMMIT.

Validation: inspected the successful current-head Node 22 CI log (all 13 Prisma regressions and 6,905 SDK tests), and confirmed the Node 24 job also passed. Checked the exact Prisma 7.10 client transaction-manager source and adapter rollback implementation; a separate SQLite-only probe confirmed that reader-blocked COMMIT leaves the transaction active until explicit rollback. The complete Session reproduction was not executed locally because dependency installation did not finish; native artifact provenance was not verified.

examples/memory/sessions/prisma.ts
jbeckwith-oai: Fixed in 4ac0ecffe45a2ce117f3bb91899501b54110a632. Both version-pinned adapter distributions now check `client.inTransaction` in `rollback()` and execute SQL `ROLLBACK` before releasing the mutex in `finally`. The check skips redundant SQL when Prisma has already rolled the transaction back, and Prisma preserves the original COMMIT error.

Added two real SQLite subprocess regressions for reader-blocked append/pop COMMIT. A Vitest spy calls the real session lookup and then starts the reader, so INSERT/DELETE reaches the intended transaction before COMMIT times out. The tests verify retained history, a second connection's ability to append/pop before the failed client is reused, and subsequent append/pop on that same client. SQL traces confirm the prior patch reaches INSERT/DELETE, fails COMMIT with SQLITE_BUSY, and leaves the other connection blocked; the correction passes both cases.

All 15 Prisma tests and actual example startup/restart/persistence passed locally, along with the full 6,905-test verification stack. Both Node 22 and Node 24.3 CI Prisma steps pass on this commit: https://github.com/openai/openai-agents-js/actions/runs/34775073409 . Two fresh independent review rounds found no remaining in-scope blockers. The lockfile change only updates the existing patch hash; no dependency versions or Session APIs changed.

markstuart-oai (approved):
Re-reviewed `4ac0ecffe45a2ce117f3bb91899501b54110a632`, including all three files changed since my prior review and the full 15-file scope against `7f135afcde3f75082a212f34f604e097d0b3bb46`. The failed-COMMIT cleanup finding is resolved; no actionable findings remain.

The adapter now rolls back an active SQLite transaction before releasing its mutex, while avoiding a second SQL rollback after normal cleanup. Both module formats match. The new reader-blocked append/pop tests verify unchanged history, independent-client access, and recovery on the original client. This stays within the existing transaction lifecycle without adding retries or another abstraction.

[Exact-head CI](https://github.com/openai/openai-agents-js/actions/runs/34775073409) passed: both Node 22 and 24.3 ran all 15 Prisma regressions, including the two new COMMIT cases, and 6,905 SDK tests. Validation used source inspection and hosted logs; no local installation, tests, native-code execution, or provider calls. Native artifact provenance remains unverified.

seratch (approved):

PR #1889, opened 2026-09-12
fix: trim Blaxel S3-compatible bucket prefixes in linear time

jbeckwith-oai:
## Summary

Blaxel S3-compatible mounts now trim prefix boundary slashes with a linear scan while preserving the existing mount target. This covers S3, R2, and GCS HMAC mounts, including empty prefixes and interior slashes, without changing public APIs or credential handling.

Adds nine regression cases through public client creation and a patch changeset for `@openai/agents-extensions`.

## Validation

- Full repository verification passed: build, workspace/declaration type checks, lint, formatting, and 6,898 tests across 225 files.
- Docker tests ran against local Colima with a shared temporary directory.
- Independent security and compatibility review completed; two consecutive adversarial-review rounds passed.

Please include maintainer security review of the sandbox mount-command boundary.

### DIFF
--- a/.changeset/tidy-blaxel-bucket-prefix.md
+++ b/.changeset/tidy-blaxel-bucket-prefix.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+fix: trim Blaxel S3-compatible bucket prefixes in linear time while preserving mount targets.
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Reviewed the complete diff at `db2f8daf` with an independent reviewer, including the sandbox mount-command boundary. The two local scans preserve absent, empty, slash-only, and interior-slash prefixes across S3, R2, and GCS HMAC routes. The resulting bucket still uses the shared shell-quoting path, and credential selection and cleanup remain intact. No actionable findings; the direct replacement needs no additional abstraction.

Validation: independent prefix-equivalence and complete-command probes passed, including shell-quoting round trips with synthetic inputs. I inspected the successful Node 22 CI run, including all 97 Blaxel tests and the 6,898-test suite; Node 24 and changeset checks are also green. No live provider calls were made.

PR #1890, opened 2026-09-12
fix(openai): treat blank OPENAI_WEBSOCKET_BASE_URL as unconfigured

sbguangha:
### Summary

`.env` files often contain `OPENAI_WEBSOCKET_BASE_URL=` (empty). Docs treat a missing websocket URL as “use the OpenAI client base URL”. The env helper returned the empty string, so `new URL(websocketBaseURL ?? client.baseURL)` kept `""` and threw instead of falling back.

This matches the blank-`CODEX_API_KEY` contract: only a non-empty value is configured.

### Test plan

- Added regression tests: blank / whitespace-only `OPENAI_WEBSOCKET_BASE_URL` → `undefined`; a real URL is still returned after trim.
- `pnpm exec vitest run packages/agents-openai/test/defaults.test.ts` — 9 passed.

### Issue number

None. Existing-behavior fix; CONTRIBUTING allows a direct PR.

### DIFF
--- a/.changeset/blank-websocket-base-url.md
+++ b/.changeset/blank-websocket-base-url.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-openai': patch
+---
+
+fix: treat a blank OPENAI_WEBSOCKET_BASE_URL as unconfigured so the client base URL can be used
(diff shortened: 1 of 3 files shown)


seratch:
We generally recommend doing `unset` instead of handling all empty patterns on the SDK side.

PR #1897, opened 2026-09-12
fix: return generic Cloudflare worker errors

jbeckwith-oai:
### Summary

Return a generic HTTP 500 body from the Cloudflare integration worker when request handling fails. Preserve server-side diagnostics, successful responses, and trace flushing.

Add two regression tests through the exported worker handler and include them in the normal test suite, without requiring live API calls or the integration registry.

### Test plan

- The failure-response regression fails against the original implementation and passes with the change; the success control passes in both cases.
- Focused handler tests pass on Vitest 5.
- Full repository verification, including Docker tests with local Colima.
- Independent security/compatibility review and two rounds of fresh-context adversarial review.

### Checks

- [x] Added relevant tests.
- [x] Reviewed documentation requirements: no documentation changes needed for this private test fixture.
- [x] Ran the required install, build, type, declaration, lint, test, and format checks.
- [x] Completed independent review before submission.
- Changeset and live integration runs are not applicable: no published packages changed.

Maintainer review requested for the worker's error-response boundary.

### DIFF
--- a/integration-tests/cloudflare-workers/worker.test.ts
+++ b/integration-tests/cloudflare-workers/worker.test.ts
@@ -1,0 +1,65 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { run } from '@openai/agents';
+import worker from './worker/src/index';
+
+const { forceFlush } = vi.hoisted(() => ({ forceFlush: vi.fn() }));
+
+// Stub the SDK boundary to exercise HTTP handling without external services.
+vi.mock('@openai/agents', () => ({
+  Agent: vi.fn(),
+  Runner: vi.fn(),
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Reviewed the complete three-file diff at `387c63b8b75d46bf6ddb55394ae1e9606af06e4c`. No actionable findings.

The direct response replacement preserves server diagnostics and the existing trace-flush lifecycle. The two HTTP-boundary tests cover failure and success without adding runtime machinery.

Verified exact-head CI evidence: both new tests passed within the 6,907-test Node 22 suite; Node 24.3, coverage, build, lint, and type checks are green. No local dependency installation or provider/Cloudflare execution was performed.

PR #1899, opened 2026-09-12
fix: trim Blaxel GCS mount prefixes in linear time

jbeckwith-oai:
### Summary

Trim Blaxel GCS mount prefixes with a linear boundary scan while preserving the generated `--only-dir` argument. Add caller-level coverage for omitted, empty, slash-only, internal-slash, Unicode, and whitespace prefixes, plus an extensions patch changeset.

### Test plan

- Full install, build, type, distribution, lint, test, and formatting stack passed locally: 6,912 tests across 225 files.
- Docker tests used the existing local Colima daemon and a temporary directory shared with that daemon.
- Two consecutive review rounds, each with two independent reviewers, were clean; reviews covered compatibility, performance, and command safety.

### Checks

- [x] Added regression tests and a patch changeset.
- [x] Completed `pnpm test`, `pnpm -r build-check`, and the full verification script.
- [x] Completed independent review before submission.
- Documentation and public APIs are unchanged.

Maintainer review requested for the GCS mount command construction; argument quoting and credential lifecycle are unchanged.

### DIFF
--- a/.changeset/tidy-gcs-prefix.md
+++ b/.changeset/tidy-gcs-prefix.md
@@ -1,0 +1,5 @@
+---
+'@openai/agents-extensions': patch
+---
+
+fix: trim Blaxel GCS mount prefixes in linear time while preserving mount behavior.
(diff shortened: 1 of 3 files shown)


markstuart-oai (approved):
Reviewed the complete diff at `4d28446b`, with an independent reviewer. The two boundary scans preserve native GCS prefix semantics, option ordering and shell quoting while making slash trimming linear. The public client tests cover absent, empty, slash-only, interior-slash and Unicode prefixes; no actionable findings.

Validation: independent dependency-free equivalence and shell-argument probes passed; all 14 current-head hosted checks passed. The full local SDK suite and live Blaxel/GCS mounts were not run.

PR #1904, opened 2026-09-13
ci: fix changeset pull request label permissions

jbeckwith-oai:
### Summary

Fixes the Changeset workflow's PR-label permission after label synchronization returned HTTP 403 on #1895. Replace the job's `issues: write` grant with `pull-requests: write`, which GitHub documents for labeling pull requests. The job continues to use `contents: read` and its existing label calculation.

### Test plan

- Passed actionlint, Prettier, and `git diff --check` for the workflow.
- Completed two clean adversarial-review rounds, each with two independent reviewers, including the token and trusted-base execution boundary.
- TruffleHog scanned the workflow with zero detected secrets.
- SDK tests and a changeset are not applicable to this workflow-only permission change. No new test duplicates the permission declaration; local tests cannot emulate GitHub's token authorization.

The `pull_request_target` check uses the base workflow. Effective token permissions and a successful label update must be verified in a subsequent run after this fix reaches `main`.

### Security review

Maintainer security review requested for the job's `GITHUB_TOKEN` permission change. Review confirmed that actions remain pinned, checkout stays on the trusted base SHA with credential persistence disabled, and PR head content is only read as data. The job receives no content-write permission.


seratch (approved):
The check still fails as the main branch doesn't have the permission but getting it back from https://github.com/openai/openai-agents-js/commit/ccd2f9a5409f4de3c59ca00ca612cf340bbf9923 changes looks good to me

PR #1909, opened 2026-09-14
fix(realtime): encode large audio buffers without call stack overflow

sbguangha:
## Summary

`arrayBufferToBase64` in `@openai/agents-realtime` spreads every byte into a single `String.fromCharCode(...)` call, so once a buffer passes the engine's argument limit the spread throws `RangeError: Maximum call stack size exceeded`. In V8 that limit lands around 125KB — roughly 2.5 seconds of 24kHz PCM16 audio — and JavaScriptCore caps out even lower.

This sits on the main send path: `session.sendAudio(buffer)` → `input_audio_buffer.append`. The push-to-talk pattern documented on `sendAudio` (buffer a whole utterance with turn detection / VAD disabled, then send with `{ commit: true }`) hits it with ordinary inputs, and the Twilio transport forwards model audio through the same helper.

The sibling decoder `base64ToArrayBuffer` in the same file already uses a byte-by-byte loop, and `@openai/agents-core` ships `encodeUint8ArrayToBase64`, which handles large payloads in both Node (via `Buffer`) and browser environments. This PR reuses that shared encoder, matching how the realtime package already imports other helpers from `@openai/agents-core/utils`. Output is unchanged for buffers that previously worked.

Supersedes #1425, which was closed by the stale bot without review.

## Repro

```js
function enc(buf) {
  return btoa(String.fromCharCode(...new Uint8Array(buf)));
}
enc(new ArrayBuffer(150_000)); // RangeError: Maximum call stack size exceeded
```

## Tests

- Added a 512KB encode/decode round-trip regression test in `packages/agents-realtime/test/utils.test.ts`; it throws `RangeError` on the old implementation and passes with the fix.
- `pnpm exec vitest run packages/agents-realtime` — 399 passed.
- `build-check` (tsc --noEmit) on agents-realtime — clean.
- Changeset included (`@openai/agents-realtime` patch).


seratch (approved):

seratch (changes requested):
Can you rebase on the latest main branch?

sbguangha:
Rebased on the latest main (00a92287) - the branch is now at 887a97df with no conflicts. CI is re-running.

jbeckwith-oai (approved):
Thanks for the fix!

PR #1915, opened 2026-09-14
fix: pass normalized arguments to needsApproval

### LINKED ISSUE #1914: needsApproval gets null for omitted optional fields on Zod and strict JSON schema tools
### Describe the bug

For Zod tools and strict JSON schema tools, `needsApproval` gets the raw `JSON.parse` of the model's arguments, while `execute` gets the input the SDK parsed.

In strict mode (the default), an `.optional()` field is sent to the model as required and nullable, so when the model doesn't set it, it sends `null`. The SDK removes that `null` before calling `execute`, so `execute` sees the field as absent. `needsApproval` still sees `null`, even though its input type says `string[] | undefined`.

So an approval callback that's correct for its declared type can throw, and `run()` rejects with `ToolCallError`. Checks like `cc !== undefined` also make the wrong approval decision.

Standard Schema tools already get the parsed value here (#1648). It's only Zod and plain JSON schema tools that don't.

### Debug information

- Agents SDK version: `@openai/agents` 0.18.0, and main at 8831eae9
- Related package versions: `zod` as pinned in the repo
- Runtime and version: Node.js 25 locally, but it's plain argument handling so the runtime doesn't matter
- Operating system: macOS
- Model and model provider: none, the repro uses `ScriptedModel`
- Does the issue reproduce with the latest Agents SDK release? Yes, 0.18.0
- Does the issue occur consistently or intermittently? Consistently

### Repro steps

```typescript
import { z } from 'zod';
import { Agent, run, tool, Usage } from '@openai/agents';
import {
  ScriptedModel,
  assistantMessage,
  functionCall,
  modelResponse,
} from '@openai/agents/testing';

const model = new ScriptedModel([
  modelResponse({
    output: [
      functionCall('send_email', '{"to":"bob@example.com","cc":null}', {
        callId: 'call_1',
      }),
    ],
    usage: new Usage(),
  }),
  modelResponse({ output: [assistantMessage('done')], usage: new Usage() }),
]);

const sendEmail = tool({
  name: 'send_email',
  description: 'Send an email',
  parameters: z.object({
    to: z.string(),
    cc: z.array(z.string()).optional(),
  }),
  // cc is typed as string[] | undefined here.
  needsApproval: async (_ctx, { cc }) => cc !== undefined && cc.length > 0,
  execute: async () => 'sent',
});

await run(new Agent({ name: 'a', model, tools: [sendEmail] }), 'send it');
```

What I get:

```text
ToolCallError: Failed to run function tools: TypeError: Cannot read properties of null (reading 'length')
```

If the callback just records its input instead, it gets `{"to":"bob@example.com","cc":null}` while `execute` gets `{"to":"bob@example.com"}`. A strict JSON schema tool with `cc` left out of `required` behaves the same way.

### Expected behavior

`needsApproval` gets the same parsed input that `execute` gets, `{ to: 'bob@example.com' }`. The callback above then returns `false` and the run finishes normally.

dajiaohuang:
## Summary

- pass schema-normalized function arguments to dynamic approval callbacks for Zod and strict JSON-schema tools
- preserve independent execution input by preparing a separate invocation value
- add regression coverage for optional null fields in both schema forms

Closes #1914

## Validation

- pnpm exec vitest run packages/agents-core/test/run.test.ts
- pnpm exec vitest run packages/agents-core/test/runner/toolExecution.test.ts
- pnpm -F @openai/agents-core build-check
- pnpm exec prettier --check packages/agents-core/src/runner/toolExecution.ts packages/agents-core/test/run.test.ts .changeset/normalized-approval-input.md
- pnpm exec eslint packages/agents-core/src/runner/toolExecution.ts packages/agents-core/test/run.test.ts


dajiaohuang:
Pushed 16607a3 to address the approval/execution isolation finding. Dynamic and Standard Schema tools now reuse one normalized execution snapshot while passing an independent, prototype-preserving clone to `needsApproval`; the regression covers approval mutation and the existing Standard Schema lifecycle.

Validation: focused Runner tests passed (7 selected normalization/snapshot tests and the isolated Standard Schema test), `@openai/agents-core` build-check passed, and Prettier passed on all changed files.

@codex review
@codex security review

dajiaohuang:
Follow-up on 5c45c02: the approval isolation now clones only plain objects/arrays and exact built-ins. Custom schema instances with private fields or other internal slots are reparsed for the approval callback, while execution keeps the original prepared snapshot; approval mutations therefore cannot alter execution and custom-instance behavior remains valid.

Validation: full `packages/agents-core/test/run.test.ts` passed (221/221), the private-field Standard Schema isolation test passed, `@openai/agents-core` build-check and Prettier passed, and `git diff --check` is clean. Two independent local reviews report no remaining blocker; hosted Codex reviews are running on 5c45c02.

dajiaohuang:
Final test correction is included in f7e8bd8: the async Standard Schema regression now returns an unclonable private-field instance on its first validation, so the custom-instance approval fallback is exercised and the second asynchronous validation failure is still rejected before callbacks.

Validation: the targeted async-validation test passes; full `run.test.ts` passed 221/221; the private-field isolation test passed; build-check, Prettier, diff-check, and the repository pre-commit secret scan passed. Hosted Codex review is running on f7e8bd8.

PR #1924, opened 2026-09-14
fix(memory): update Prisma mysql2 dependency to 3.23.1

jbeckwith-oai:
Prisma 7.10.0 pins mysql2 3.15.3 in the private memory example. Add a parent-specific override to mysql2 3.23.1, which includes the upstream authentication and compressed-packet fixes covering Dependabot alerts 500 and 501.

The lockfile includes mysql2's sql-escaper 1.5.1 replacement and Node types peer. Existing SQLite schema, transaction behavior, adapter patch, unrelated resolutions, and the seven-day release policy are preserved. Published SDK dependency manifests are unchanged.

## Validation

Final head `9a1c541b`: all 13 hosted checks pass, including CodeQL, Node 22/24, Prisma regressions, and coverage with all 6,931 tests passing (including Docker).

- Rebased onto current main after #1925 and #1926, preserving both merged dependency fixes. Two fresh consecutive independent adversarial-review rounds on the rebased diff, with two reviewers per round: clean.
- Prisma generation and all 15 memory persistence, contention, recovery, and cleanup tests pass.
- Frozen install, SDK build, workspace type checks, distribution checks, lint, formatting, and changed-file secret scan pass.
- Local full-suite verification passed 6,928 tests; three Docker cases hit the established remote-daemon DNS and bind-mount limitations. Host-injected proxy warnings were selectively disabled during the run. Hosted CI validates these Docker cases on the final head.

Maintainer security review is requested for this dependency update. No live MySQL exploit path is claimed for the SQLite example.

## Separate follow-up

Dependabot alert 437 remains open: @prisma/config 7.10.0 pins deepmerge-ts 7.1.5. Version 8 changes merge behavior, so this PR leaves its remediation for an upstream-supported Prisma release or a separately approved compatibility assessment.


markstuart-oai (approved):
Reviewed both changed files and the private Prisma dependency boundary, with an independent dependency-security pass. No actionable findings.

The parent-version-scoped override is justified: Prisma 7.10.0 pins mysql2 3.15.3 exactly, so refreshing the lock alone cannot select the fixes. The removal condition is documented, the new Node types peer resolves consistently, and published SDK manifests are unchanged. This is dependency remediation in the SQLite example's tooling; it does not establish an exploitable MySQL path in that example.

Verified the upstream advisory fixes, tarball/lock integrities and lifecycle hooks. Focused probes against the actual package source confirmed default cleartext authentication is rejected while explicit opt-in remains, compressed output is bounded by the declared size, and the replacement queue preserves ordering. Representative escaping checks passed; SQL Escaper intentionally differs from SQLString for some object/placeholder cases, so this approval is scoped to the existing SQLite usage rather than universal API equivalence.

Validation: all 13 checks pass for `ea9083f`. The inspected Node 22 CI log verifies this head merged into `b54ca2b`, install policy checks, build/declarations/examples, all 6,931 tests, Prisma generation and all 15 memory regressions. No live MySQL test or full local suite was run during this review.

PR #1926, opened 2026-09-14
fix(deps): remediate Blaxel generator and TOML alerts

jbeckwith-oai:
### Summary

Removes vulnerable workspace dependency resolutions introduced by the Blaxel development dependency: `@hey-api/openapi-ts` 0.96.1 → 0.97.3 and `toml` 3.0.0 → 4.2.0. Covers GHSA-hhx9-57xq-r5rw, GHSA-v5mp-jgw5-2x6j, and GHSA-82x6-q7mm-w9cf.

Uses an exact, Blaxel-version-scoped TOML override until upstream ships a patched dependency. The refreshed generator also requires a parser-scoped `js-yaml` 4.3.2 override to avoid its vulnerable 4.1.1 pin. Blaxel 0.2.82, its optional public peer range, unrelated resolutions, existing patches, and the seven-day release-age policy stay unchanged.

These overrides apply to this workspace; they do not change consumers' optional dependency installations or regenerate third-party clients. TOML 4.2.0 requires Node ≥20, within the workspace's Node 22/24 CI matrix.

### Test plan

- 104 Blaxel adapter tests passed.
- Actual installed Blaxel configuration-loader probes passed for ordinary scalar/array/date values, malformed-input fallback, the published prototype-pollution payload, and excessive nesting.
- Generator prototype protection, patched YAML resolution, and generation from a synthetic YAML specification passed.
- Two consecutive clean adversarial-review rounds, each with two independent fresh-context reviewers, on the complete final diff.
- Dependency manifests, install scripts, release age, integrity, and source reviewed; no new installation hooks. Changed-file secret scan found no secrets.

- Final full stack: frozen install, build, type checks, declaration checks, lint, and formatting passed. The suite had 6,922 passing tests and nine environment-related failures (injected proxy warnings, remote Docker DNS/bind-path errors, and a file-helper timeout). All 141 tests across the five affected non-Docker-integration files passed on a focused rerun with only the injected proxy warning disabled. Three live Docker tests remain unverified locally; local Docker socket access is denied in the verification sandbox. No tests or repository configuration were changed to hide these failures.

No changeset is needed for this workspace-only dependency change.

Maintainer security review requested for the local configuration parsing and code-generation dependency boundaries.


markstuart-oai (approved):
Reviewed the full workspace override and lockfile diff at `1fa43e1`, including independent dependency security review. No actionable findings. The two parent/version-scoped overrides are justified: Blaxel's `toml` range cannot select the patched major, and the updated generator parser otherwise pins an affected `js-yaml` release. They keep the remediation at the dependency boundary without adding SDK wrappers or upgrading the provider/public peer contract.

Verified changed registry artifacts and lock integrity, dependency and install-hook changes, Node compatibility, the generator's null-prototype parameter handling, and the TOML fixes. Focused checks against the published Blaxel loader with synthetic filesystem/process inputs preserved ordinary values and fallback behavior; patched TOML rejected the disclosed pollution/depth cases. Generator-template and YAML parsing probes also passed. The existing lazy optional-provider import, SQLite/other provider code, patches and release-age settings are untouched.

All 13 hosted checks are green at this head. The Node 22 log confirms the complete 6,931-test suite, including 104 Blaxel tests, plus build, declarations, lint, example compilation and Prisma regressions. I did not rerun the full suite or a live Blaxel service locally. These workspace overrides do not remediate consumers' independently installed optional peers or regenerate previously generated third-party clients.

PR #1929, opened 2026-09-14
fix: Use cryptographic randomness for sandbox PTY IDs

jbeckwith-oai:
### Summary

Use Web Crypto with unbiased range mapping for sandbox PTY process IDs, preserving the existing numeric range and collision handling. If allocation fails, retain existing registry entries and await cleanup of newly created E2B and Daytona PTYs before returning the original error.

Includes focused regression coverage and a patch changeset for `@openai/agents-extensions`. The affected boundary is PTY registration and provider resource cleanup; maintainer security review is requested.

### Test plan

- Full repository verification wrapper passed: frozen install, build, workspace type checks, declaration checks, lint, formatting, and all 6,936 tests across 227 files.
- Docker tests passed with local Colima and a shared temporary directory.
- Focused PTY and provider suite: 466 tests passed across six files.
- Two consecutive clean independent adversarial review rounds, covering correctness, security, failure cleanup, compatibility, and scope.
- Changeset validation and `git diff --check` passed.

### Checks

- [x] Added focused regression tests
- [x] Ran the full code-change verification wrapper; all steps passed
- [x] Completed independent review before submission
- [x] Added and validated a patch changeset
- Documentation: no public API or usage changes
- Examples and integration runners: not run separately; no examples or integration behavior changed


packages/agents-extensions/src/sandbox/shared/pty.ts
```diff
@@ -526,11 +526,16 @@ function resolvePtyWriteYieldTimeMs(
...
+  const range = PTY_PROCESS_ID_MAX_EXCLUSIVE - PTY_PROCESS_ID_MIN;
+  // Reject the incomplete range at the top of uint32 to avoid modulo bias.
+  const limit = 2 ** 32 - (2 ** 32 % range);
+  const random = new Uint32Array(1);
   while (true) {
-    const processId =
-      Math.floor(
-        Math.random() * (PTY_PROCESS_ID_MAX_EXCLUSIVE - PTY_PROCESS_ID_MIN),
-      ) + PTY_PROCESS_ID_MIN;
+    globalThis.crypto.getRandomValues(random);
```
markstuart-oai, line 534: Please establish the supported requirement for unpredictable PTY IDs before adding the cryptographic allocator and its new provider failure paths. The shell capability already captures the owning session, each adapter looks up this number in its own registry, and the existing allocator rejects collisions; changing the random source does not change those routing or ownership boundaries. The PR currently gives no concrete consumer consequence that the existing path gets wrong. Please provide that scenario or an established unpredictability contract; without one, retaining the current allocator is the simpler design. The new RNG-failure tests establish conditional cleanup correctness, but do not establish that underlying need.

markstuart-oai (commented):
Need status: plausible but unproven. I left an inline request for the supported outcome or contract that requires unpredictable session-local PTY routing IDs. The existing collision-checked registry already provides routing and separation; the evidence does not yet justify expanding its allocation and provider-failure behavior. I am withholding approval pending that rationale, without claiming a demonstrated security vulnerability.

The implementation review found no introduced correctness or cleanup-ownership defect. Range mapping is unbiased, allocation precedes pruning, and failure cleanup captures only the new attempt's handle or socket. I traced E2B, Daytona, Blaxel and Cloudflare, including cleanup overlapping another registration or session teardown. Node-only randomness and UUIDs would be worse fits for the shared runtime boundary and numeric contract; retaining the current implementation remains the strongest alternative until the need is established.

All 12 hosted checks passed at this head. I inspected the Node 24 job: 6,936 tests across 227 files passed, including the changed PTY/provider suites, with build/type/lint checks. The patch changeset is present and its hosted validation passed. This was a desk review with independent review and existing hosted evidence; no local tests or live sandbox operations were run.

packages/agents-extensions/src/sandbox/shared/pty.ts
jbeckwith-oai: Agreed. I rechecked the supported path and found no contract or concrete consumer consequence requiring unpredictable PTY IDs. The shell capability captures the owning session, and each affected adapter resolves the number only in its own collision-checked registry. Released Unix-local and Modal implementations also use sequential IDs through the same interface.

The new failure tests establish cleanup correctness conditional on this allocator choice; they do not establish the underlying need. With maintainer approval, I am closing #1929 without merging and retaining the existing allocator. No replacement patch is needed.

PR #1930, opened 2026-09-15
fix: serialize hosted MCP approval tool references without live configuration

jbeckwith-oai:
Hosted MCP approval snapshots retain the configured tool even though resume resolves that tool from the current agent configuration. Store only its identity while preserving approval request metadata and compatibility with existing snapshots.

Regression tests cover streaming and non-streaming serialization, unchanged live configuration, replacement configuration on resume, approval response identity, and older snapshots.

Validation: two consecutive independent review rounds passed; 145 focused tests passed. Install, build, type checks, declaration checks, lint, formatting, and changeset validation passed. The local full suite passed 6,925 tests and failed 8 unchanged environment-dependent cases: five from injected Node proxy warnings in expected output and three from Docker DNS/bind-mount failures. CI will provide the remaining full-suite validation.

Please include maintainer security review of the configuration-to-persistence boundary.


markstuart-oai (approved):
Reviewed the configuration-to-persistence boundary and approval/resume behavior at `a419626a`. No consequential findings. The projection preserves approval metadata and tool identity, leaves live configuration unchanged, and reuses current-tool rehydration. The latest release, v0.18.0, already accepts these retained fields and resolves the current tool, so this preserves the existing schema's reader meaning.

Validation: 216 focused tests passed on Node 24.20.0. Both new regressions fail against the verified base/released serializer. Independent streaming and non-streaming probes confirmed stored configuration cannot override the replacement tool or its current approval callback. I also verified the exact-head Node 22 CI log: all 6,933 tests passed, including the new snapshot tests; all PR checks are green.

No live API/MCP calls or local packaged/full-suite validation. This review covers the changed tool-configuration boundary, not general sanitization of application context or provider payloads.

PR #1932, opened 2026-09-15
fix: Apply current MCP tool filters to cached discovery

jbeckwith-oai:
MCP tool enumeration currently reuses an earlier filter decision when a cached discovery entry is available. Cache raw tool descriptions and apply the current static or callable filter on every lookup, so changing run contexts and filter settings produce the current tool list while discovery remains cached.

Preserves cache partitioning, invalidation, descriptor isolation, current-server wrapper binding, and prefixed tool names. Includes an `@openai/agents-core` patch changeset.

## Validation

- 113 focused MCP tests pass; the four new regression cases fail against the original source.
- Core source/test typecheck, build, workspace typechecks, declaration checks, lint, and formatting pass.
- Two consecutive clean independent review rounds, with two fresh reviewers per round.
- Full local suite: 6,927 passed and eight environment failures in unchanged tests: five assertions affected by injected Node warnings and three Docker connectivity/mount failures. CI will provide the remaining verification.


markstuart-oai (approved):
Reviewed `fc5c2041`, including the tool-exposure boundary, cache invalidation, descriptor ownership, and current-server binding. No consequential findings. Caching raw discovery while applying the current filter on every lookup removes stale filter decisions without adding a second cache or changing tool conversion.

Validation: 117 focused tests passed on Node 24.20.0; all four selected new regressions fail against the exact base implementation. An independent controlled concurrency probe confirmed a suspended old filter cannot repopulate invalidated discovery, and checked descriptor isolation, replacement-server invocation, and replacement filters. I verified the exact-head Node 22 CI run passed all 6,935 tests, including the 68 cache tests. All PR checks are green.

No live MCP calls, local package build, or local full-suite run. These checks assess the SDK cache/filter boundary, not application-wide authorization.

PR #1935, opened 2026-09-15
fix(agents-core): handle same-server MCP tool name collisions

### LINKED ISSUE #1934: getAllMcpTools only rejects duplicate tool names across MCP servers, so two tools on the same server that normalize to one function-tool name are both returned
### Describe the bug

`getAllMcpTools` only rejects duplicate function-tool names **across** MCP servers. Two tools exposed by the **same** MCP server whose names differ only by characters that `toFunctionToolName` rewrites (for example `search-a` and `search_a`, or `a.b` and `a_b`) are both returned under one identical function-tool name.

In the default path (`includeServerInToolNames: false`, `packages/agents-core/src/mcp.ts:890-900`) the guard is:

```ts
const serverToolNames = new Set(serverTools.map((t) => t.name));
const intersection = [...serverToolNames]
  .filter((n) => toolNames.has(n))
  .sort();
if (intersection.length > 0) {
  throw new UserError(
    `Duplicate tool names found across MCP servers: ${intersection.join(', ')}`,
  );
}
```

`toolNames` accumulates names only from *previously processed* servers, and `serverToolNames` is a `Set`, so a collision inside the current server's own batch is collapsed before the comparison and never detected. The `includeServerInToolNames: true` branch (`mcp.ts:865-873`) has the same shape.

Each tool is then built from the raw MCP name by `mcpToFunctionTool` and re-normalized by `tool()` via `toFunctionToolName` (`packages/agents-core/src/utils/tools.ts:87`), which replaces every non-`[A-Za-z0-9]` character with `_`. Two distinct MCP tool names therefore collapse to a single function-tool name.

Downstream, `resolveModelVisibleToolNameCollisions` (`packages/agents-core/src/runner/modelPreparation.ts:186`, wired at `modelPreparation.ts:99`, default policy `warn` per `modelPreparation.ts:93` and `packages/agents-core/src/runner/runConfig.ts:19`) keeps only one entry per colliding name and logs a generic warning — one remote MCP tool becomes unreachable. With `toolNameCollisionPolicy: 'error'` the run instead throws `Duplicate enabled function tool or handoff names found`. The cross-server equivalent throws a hard `UserError` from `getAllMcpTools` itself.

**Root cause** — `packages/agents-core/src/mcp.ts:892-900` (default branch) and `packages/agents-core/src/mcp.ts:865-873` (namespaced branch) inside `getAllMcpTools` (`mcp.ts:805`): the intersection is computed against the cross-server `toolNames` set only, never against the names allocated earlier for the same server. In the namespaced branch specifically, `getSafeToolNamePart` (`mcp.ts:916-925`) treats both `-` and `_` as safe characters and preserves them, so `search-a` and `search_a` yield *different* prefixed base names, no hash is forced by `buildPrefixedToolNameOverrides` (`mcp.ts:1050`), and the collision only appears later when `toFunctionToolName` rewrites both to `mcp_docs__search_a`.

The repo's own test `it('allocates normalized MCP tool name collisions stably')` (`packages/agents-core/test/mcpCache.test.ts:870`) asserts this invariant for one server using `['search', 'search!']`. That case passes because `getSafeToolNamePart` maps `!` to `_` and strips the trailing `_`, so both base names become `mcp_docs__search` and a hash is forced. `-` vs `_` slips through that path.

### Debug information

- Agents SDK version: 0.18.0 (`@openai/agents-core`)
- Related package versions (optional, e.g. `openai`, `zod`, or `@openai/agents-extensions`): none — the reproduction uses only `@openai/agents-core`
- Runtime and version (e.g. Node.js, Deno, or Bun): Node.js v22.23.2, pnpm 11.24.0
- Operating system: macOS (Darwin 25.6.0)
- Model and model provider: none — `getAllMcpTools` runs before any model call and needs no model or API key
- Does the issue reproduce with the latest Agents SDK release? Yes. Reproduced on current `main` (`d187747e9b905a515608bc1d6720d0e52b766b5e`) and on `00a92287e61a28f9bf89e760bd00d3656a998827`; the affected code is identical in both.
- Does the issue occur consistently or intermittently? Consistently.

No error or stack trace is produced — that is the defect. The observed result object, produced by the test in **Repro steps**:

```text
{"c1_oneServer_dash_vs_underscore":["search_a","search_a"]}
{"c2_oneServer_dot_vs_underscore":["a_b","a_b"]}
{"c3_oneServer_with_prefix":["mcp_docs__search_a","mcp_docs__search_a"]}
{"c4_distinct_names":["search_a","search_b"]}
{"c5_cross_server_duplicate":"THREW Duplicate tool names found across MCP servers: q"}
```

### Repro steps

Save as `packages/agents-core/test/repro-mcp-dup.test.ts` and run:

```bash
pnpm exec vitest run packages/agents-core/test/repro-mcp-dup.test.ts --project @openai/agents-core
```

```typescript
import { describe, it } from 'vitest';
import { appendFileSync, rmSync } from 'node:fs';
import { getAllMcpTools } from '../src/mcp';
import type { MCPServer } from '../src/mcp';
import { RunContext } from '../src/runContext';
import { Agent } from '../src/agent';
import { withTrace } from '../src/tracing';

const OUT = '/tmp/repro-mcp-dup.out';
try {
  rmSync(OUT);
} catch {}
const log = (obj: unknown) => appendFileSync(OUT, JSON.stringify(obj) + '\n');

function stubServer(name: string, toolNames: string[]): MCPServer {
  return {
    name,
    cacheToolsList: false,
    connect: async () => {},
    close: async () => {},
    listTools: async () =>
      toolNames.map((n) => ({
        name: n,
        description: '',
        inputSchema: {
          type: 'object',
          properties: { q: { type: 'string' } },
          required: ['q'],
          additionalProperties: false,
        },
      })),
    callTool: async () => [],
    invalidateToolsCache: async () => {},
  } as unknown as MCPServer;
}

async function names(servers: MCPServer[], includeServerInToolNames = false) {
  return withTrace('repro', async () => {
    const tools = await getAllMcpTools({
      mcpServers: servers,
      runContext: new RunContext({}),
      agent: new Agent({ name: 'AgentOne' }),
      includeServerInToolNames,
    });
    return tools.map((t) => t.name);
  });
}

describe('repro', () => {
  it('runs', async () => {
    log({ c1_oneServer_dash_vs_underscore: await names([stubServer('docs', ['search-a', 'search_a'])]) });
    log({ c2_oneServer_dot_vs_underscore: await names([stubServer('docs', ['a.b', 'a_b'])]) });
    log({ c3_oneServer_with_prefix: await names([stubServer('docs', ['search-a', 'search_a'])], true) });
    log({ c4_distinct_names: await names([stubServer('docs', ['search_a', 'search_b'])]) });
    try {
      const r = await names([stubServer('s1', ['q']), stubServer('s2', ['q'])]);
      log({ c5_cross_server_duplicate: r });
    } catch (e) {
      log({ c5_cross_server_duplicate: `THREW ${(e as Error).message}` });
    }
  });
});
```

`c1`/`c2` are the default (`includeServerInToolNames: false`) path, `c3` is the same server with `includeServerInToolNames: true`, `c4` is the non-colliding control, and `c5` shows the cross-server case that *is* rejected today.

### Expected behavior

Two distinct MCP tools must not be returned under the same function-tool name. Either the collision should be rejected up front — the same way the cross-server case already is — or both tools should be allocated distinct names, which is what the existing single-server test (`packages/agents-core/test/mcpCache.test.ts:870`) already requires for the `includeServerInToolNames: true` path. Silently returning a duplicate name means one remote tool gets dropped at `resolveModelVisibleToolNameCollisions` with only a generic warning.

Basis:

- The sibling branch in the same function throws a hard error for exactly this condition between servers: `mcp.ts:865-873` and `mcp.ts:892-900`, both `throw new UserError('Duplicate tool names found across MCP servers: ' + intersection.join(', '))`. Only the same-server instance of the same condition is missed.
- The repo's own test `it('allocates normalized MCP tool name collisions stably')` (`packages/agents-core/test/mcpCache.test.ts:870`) asserts, for a single server, `expect(new Set(Object.values(firstOrder)).size).toBe(2)` and that neither tool keeps the un-hashed `mcp_docs__search` name.
- `docs/src/content/docs/guides/running-agents.mdx:116` states: "Existing strict validation for namespaced tools, deferred tools, and duplicate local MCP tool names is unchanged."
- The sibling SDK treats duplicate function-tool names as an error: openai/openai-agents-python#4116 "Reject duplicate function tool names" and #464 "Duplicate tool names across MCP servers cause errors".

Related references found while searching for duplicates: no existing issue in this repository covers the intra-server case. openai/openai-agents-js#1901/#1902 (`fix(mcp): treat an empty static allowlist as allowing nothing`) are open and touch `packages/agents-core/src/mcp.ts`, but are unrelated to tool-name allocation.

Happy to open a PR with the approach above (reject or allocate distinct names for duplicates within the current server batch in both branches of `getAllMcpTools`) if that direction is acceptable.

BlueX888:
### Summary

`getAllMcpTools` only detected duplicate function tool names **across** MCP servers. Two tools from the **same** server whose names normalize to one function tool name were both returned under that name, so only one of them stayed reachable by the model.

#### Why

Every MCP tool name is passed through `toFunctionToolName`, which replaces every non-alphanumeric character with `_` (`packages/agents-core/src/utils/tools.ts:87`). Raw MCP names such as `search-a` and `search_a` therefore both become `search_a`.

The guard in `getAllMcpTools` compared the current server's names against only the names allocated by earlier servers, and collapsed the current batch through a `Set` first, so a repeat inside the batch never reached the comparison:

- `packages/agents-core/src/mcp.ts:892-900` on `main` (default path, `includeServerInToolNames: false`)
- `packages/agents-core/src/mcp.ts:865-873` on `main` (`includeServerInToolNames: true`)

In the prefixed path the collision originated in `buildPrefixedToolNameOverrides` (`packages/agents-core/src/mcp.ts:1050`): `getSafeToolNamePart` treats both `-` and `_` as safe and preserves them, so `search-a` and `search_a` produced different prefixed base names, `baseNameCounts` counted one occurrence of each, no hash suffix was forced, and the two names collided only once `toFunctionToolName` normalized them.

#### Changes

- `packages/agents-core/src/mcp.ts` — both `getAllMcpTools` paths now use a `findDuplicateToolNames` helper that reports a name repeating inside the current batch as well as a name already allocated by an earlier server. The non-prefixed path raises the existing `UserError`, matching the cross-server behavior.
- `packages/agents-core/src/mcp.ts` — `buildPrefixedToolNameOverrides` counts and reserves prefixed names by their normalized function tool name, so a same-server collision forces the same deterministic hash suffix that other prefixed collisions already use. Both tools stay reachable instead of one being dropped by `resolveModelVisibleToolNameCollisions`.
- `.changeset/mcp-intra-server-tool-name-collisions.md` — patch changeset for `@openai/agents-core`.

### Test plan

Added `throws when one server returns tool names that normalize to one function tool name`, and extended the existing `allocates normalized MCP tool name collisions stably` with the `search-a` / `search_a` pair, in `packages/agents-core/test/mcpCache.test.ts`.

On the unpatched tree, before the fix:

```
$ pnpm exec vitest run packages/agents-core/test/mcpCache.test.ts --project @openai/agents-core

 FAIL  |@openai/agents-core| test/mcpCache.test.ts > MCP tools uniqueness > throws when one server returns tool names that normalize to one function tool name

- Expected:
Error {

+ Received:
[

 FAIL  |@openai/agents-core| test/mcpCache.test.ts > MCP tools uniqueness > allocates normalized MCP tool name collisions stably

AssertionError: expected 1 to be 2 // Object.is equality

- Expected
+ Received

- 2
+ 1

 Test Files  1 failed (1)
      Tests  2 failed | 67 passed (69)
```

With the fix:

```
$ pnpm exec vitest run packages/agents-core/test/mcpCache.test.ts --project @openai/agents-core

 Test Files  1 passed (1)
      Tests  69 passed (69)
```

Full verification on `d187747e9b905a515608bc1d6720d0e52b766b5e` plus this commit:

```
$ bash .agents/skills/code-change-verification/scripts/run.sh

[build-check] pnpm -r build-check exited with code 0
[dist-check] pnpm -r -F "@openai/*" dist:check exited with code 0
[test]  Test Files  227 passed (227)
[test]      Tests  6936 passed (6936)
[test] pnpm test exited with code 0
code-change-verification: all commands passed.
```

`pnpm test:examples` also passes. One earlier local run of the full suite reported two PTY-signal failures in `packages/agents-core/test/sandboxes/unixLocal.signals.test.ts`; that file contains no MCP code and fails the same way with this change stashed away, so it is unrelated to this diff.

### Issue number

Closes #1934

### Checks


packages/agents-core/test/mcpCache.test.ts
```diff
@@ -914,6 +931,23 @@ describe('MCP tools uniqueness', () => {
...
+        'search-a',
+        'search_a',
+      ]);
+      const reversedDashOrder = await publicNamesByOriginalTool([
+        'search_a',
+        'search-a',
+      ]);
+
+      expect(dashOrder).toEqual(reversedDashOrder);
+      expect(new Set(Object.values(dashOrder)).size).toBe(2);
```
jbeckwith-oai, line 996: Non-blocking coverage opportunity: this verifies stable, distinct names, while the user-visible failure is that one remote tool disappears from the runner's usable capabilities. A regression through `run()` that has a `ScriptedModel` call both advertised tools and asserts `callTool('search-a', ...)` and `callTool('search_a', ...)` would directly protect that outcome. I ran this locally in both streaming and non-streaming modes: the base exposes/invokes only `search_a`, and this head exposes/invokes both. The implementation works; this is an optional strengthening of the committed regression coverage.

jbeckwith-oai (approved):
Reviewed the complete three-file change at 32818822455a7382c700c7cd75c2bd01c60073e9 against d187747e9b905a515608bc1d6720d0e52b766b5e, including #1934, the existing allocation/dispatch paths, and the v0.18.0 compatibility baseline.

Need status: demonstrated. This is a supported tool-identity failure with a caller-visible consequence, and the patch is merge-worthy as-is. I found no blocking correctness, compatibility, or scope issues.

The root cause is twofold: the default check converts each server's names to a Set before checking only previously processed servers, hiding duplicates within a server; the prefixed allocator compares names before the final toFunctionToolName normalization, so search-a and search_a appear distinct until both become mcp_docs__search_a. The runner then retains only one function under its default collision policy.

I independently reproduced both levels of the problem using source snapshots of the exact base and head, synthetic MCP discovery/call stubs, and the SDK's ScriptedModel:
- Running the PR's mcpCache tests against the base produces exactly the two reported failures: 67 pass and 2 fail. Both failures disappear on the PR head.
- Through the public run API, the base exposes only one prefixed tool and invokes only the original search_a tool. The head exposes two unique tools and invokes both original names, search-a and search_a. This holds for streaming and non-streaming runs.
- Without prefixing, the head raises UserError before any model request or MCP invocation, rather than silently allowing the runner to drop a tool.
- Additional controls confirm that a normalized local function name is reserved correctly, and filtering out one colliding tool avoids an unnecessary error.

The implementation is placed at the right boundary. findDuplicateToolNames checks both intra-batch and prior-server duplicates without mutating the caller's set. Applying the existing normalization to the allocator's counts, reservation checks, and used-name checks closes the same identity gap throughout allocation. Keeping raw tool names in the hash seed and dispatch closure preserves deterministic disambiguation and correct remote invocation. Changing global function-name normalization would have a much broader compatibility impact; reusing this existing allocator is the narrower fix. Configuring prefixes alone does not solve the problem on the base.

I also inspected the affected identity/security boundary: server-name redaction still occurs before prefix/hash allocation, and this patch does not change MCP approval, guardrail, credential, or execution configuration. Existing focused coverage for redaction, guardrail isolation, cache rebinding/filtering, reserved handoff names, and long-name allocation passes.

Validation: 139 tests passed across mcpCache, mcpToFunctionTool, mcpToolFilter, runner/modelPreparation, and six independent reproduction/control cases. The six controls also pass against the base with expectations for its observed broken behavior. ESLint passed for both changed TypeScript files; Prettier passed for all three changed files; git diff --check passed.

Limits: these were deterministic local tests, not live model/MCP service calls. I did not rerun the full monorepo verification stack. GitHub CI for this head currently reports action_required, so this code approval does not establish CI readiness.

packages/agents-core/test/mcpCache.test.ts
BlueX888: Added in 6351c77, following that shape.

`invokes both tools of one server through run()` runs the agent with a `ScriptedModel` that calls each advertised name once and asserts both original names reached the server (`callTool` receives `search-a` and `search_a`), parameterized over streaming and non-streaming. It reads the advertised names from `agent.getAllTools()` rather than hard-coding them, so it stays correct if the allocator changes.

I checked it protects the outcome rather than just the allocation: on the base commit it fails with `expected 1 to be 2` in both modes, since only one tool is advertised, and the pair passes here. `mcpCache.test.ts` is 71 passed, `tsc --noEmit -p tsconfig.test.json` is clean and prettier passes.

Your point about the assertion being the useful half is why I took this rather than another name-shape assertion — the RPC that a user hits is the one that silently stopped existing.

seratch (approved):

PR #1937, opened 2026-09-15
fix: Honor computer safety-check rejection without aborting runs

jbeckwith-oai:
## Summary

Fixes computer safety-check handling so an explicit `false` rejects the associated computer call before actions or screenshots execute. The run receives the existing rejection result and model-visible message and can continue without a new exception or catch/retry requirement.

Public types and omitted, `void`, `true`, and acknowledgement-object callback behavior remain unchanged. The intentional behavior change is that an explicitly denied action no longer executes.

## Validation

- 484 focused tests passed, covering normal and streamed runs, approval/resume, cleanup, hooks, and cancellation during rejection formatting.
- Two consecutive independent review rounds passed, including compatibility, security, and resource lifecycle review.
- Install, build, type checks, declaration checks, lint, formatting, and patch changeset validation passed.
- Full local suite: 6,929 passed; 8 failed in five unchanged files. Five exact-output assertions receive injected proxy warnings; three Docker tests have daemon DNS or bind-path failures. CI must confirm the full suite in its runner environment.

## Review focus

Please include maintainer security review of the tool-approval boundary: a host callback's explicit denial must prevent all associated computer actions while preserving run continuation and cleanup.


markstuart-oai (approved):
Reviewed the computer safety-denial boundary and continuation behavior. Explicit `false` now prevents all associated actions and screenshots and reuses the existing rejection result/message. Human approval still runs first; absent/void/true and acknowledgement-object callbacks retain their prior behavior. The public callback type is unchanged. The shared rejection builder is a proportionate change, and I found no consequential issues.

Independent checks covered normal and streamed continuation, serialized approval/resume with the current callback, Responses request conversion, mixed accepted/denied calls, and cancellation during the callback and rejection formatter. Resource ownership remains coherent: ordinary failure disposes the prepared computer, while cancelled streaming preserves it for resume and disposes it after completion.

Validation: 51 focused tests passed; all four selected regressions failed against the exact base as expected. Four lifecycle scenarios and the serialized-resume/conversion probes passed. CI passed 6,957 tests on Node 22 and 24 and in the coverage lane, plus builds, lint, type/example checks, Windows checks, and released API checks. No live provider or real computer-environment calls were made.

PR #1938, opened 2026-09-15
fix: withhold unvetted assistant output after guardrail errors

jbeckwith-oai:
When an output guardrail throws without returning a tripwire verdict, the run rejects but can still persist its unvetted final assistant response. A later run then replays that response from the session.

Route failed assistant-output validation through the existing blocked-output persistence policy in both streaming and non-streaming runs. Preserve the primary guardrail error, existing safe-history handling, and successful-output behavior. This adds no public API or persisted-format changes and includes a patch changeset.

## Review and validation

- Reproduced through the public runner in both run modes and both session capability categories; regression tests verify persistence and replay exclusion, accepted-output controls, and primary-error preservation.
- 236 focused tests passed; two independent adversarial/security review rounds were clean.
- Full repository verification passed: frozen install, build, workspace and distribution type checks, lint, all 6,962 tests, and formatting. Docker tests used local Colima.
- Maintainer security review requested for the model-output-to-session persistence boundary.


markstuart-oai (approved):
Reviewed `457dfff8ce30d19ccbd4a3796ec482356dc19dfa` against `8ac97dfedd0395beeb38edb0a17b83a9b89c3354`, including the model-output-to-session persistence boundary. No actionable findings. Failed assistant-output validation now uses the existing blocked-output policy in both run modes, excluding unvetted output from later replay while preserving the primary error, established safe-history handling, and terminal-tool behavior. The fix stays at the shared decision point and adds no parallel persistence mechanism or durable format.

Validation: 71 focused tests passed; all five new regressions fail against the exact base as expected. Six independent concurrency/cancellation probes passed, including late guardrail completion and concurrent session appends. Exact-head CI passed, with 6,962 tests on Node 24 and green Node 22, Windows, coverage, and changeset checks. The patch changeset covers the changed package. Local checks used synthetic sessions/models; no live API or external database was exercised.

joinwell52-AI:
@jbeckwith-oai Thanks for separating a failed check from a tripwire verdict. This prompted a persistence-and-replay experiment for our agent-runtime research. At 457dfff8, we used the public run API with scripted models, MemorySession and an append-only session, in both run modes. The control replaced only guardrails.ts with its exact base version; no live model API or external database was involved.

Across 32 observations, the error and mixed pass/error cases each went from 4/4 unvetted responses replayed to 0/4. Passing responses stayed at 4/4, with prior accepted history and the current user input preserved. That made the distinction between an audit record of generated text and working memory for the next turn especially concrete for us.

A follow-up design question, outside this patch: if an application later retries verification after the checker recovers, what identity should bind that verdict to the original answer and turn? We would want to avoid attaching a delayed pass to edited or regenerated content. We have not tested such a deferred-verification path; is there an existing pattern you would recommend for keeping it separate from automatic session replay?

[Probe, fixed versions and saved observations](https://github.com/joinwell52-AI/joinwell52/blob/e6d8c818070e3f6e3b5080cd601b4cc2d295c63c/research/manual-runs/2026-09-16-current-context/README.en.md)

PR #1940, opened 2026-09-16
fix: prevent stack overflow when encoding large Realtime audio buffers

jbeckwith-oai:
### Summary

Fix `sendAudio()` throwing `RangeError: Maximum call stack size exceeded` when encoding large Realtime audio buffers. Convert bytes to a binary string in bounded chunks before base64 encoding, preserving exact audio bytes and the existing API and append/commit behavior.

### Test plan

- Reproduced the original failure with a 1,048,578-byte buffer; the same regression passes after the fix. Empty and small buffers are also covered.
- All 128 focused utility, base transport, and WebSocket tests pass.
- Dependency install, package build, workspace type checks, distribution declaration checks, lint, formatting, and patch changeset validation pass.
- Full local suite: 6,957 passed, 8 environment-related failures. Five process/CLI tests receive an injected proxy warning in captured output; three Docker tests encounter unavailable daemon DNS or host bind mounts. GitHub CI must still pass.
- Two consecutive independent adversarial review rounds, with two fresh reviewers per round, found no blocking issues.

### Checks

- [x] Added regression tests
- [x] Ran the code-change-verification script; local environment failures are detailed above
- [ ] All verification steps pass
- [x] Completed independent code review
- [x] Added a patch changeset for `@openai/agents-realtime`


markstuart-oai (approved):
Reviewed at `c4e3944b0fcd1fee7bee8e6c9bdbe94fc5276270`. No actionable findings.

The fix stays in the shared encoder and bounds only the byte-to-string conversion. Concatenating the binary chunks before a single `btoa` call preserves base64 grouping and exact bytes. The public API, connection guard, and append/optional-commit ordering remain unchanged. The regression tests exercise public `sendAudio` with an independent `Buffer` oracle, including a large buffer with a partial final chunk. The patch changeset is appropriate.

Validation: independent source review, all 128 focused utility/base-transport/WebSocket tests passed locally on Node 24, formatting and diff checks passed, and all 14 returned exact-head CI checks were successful. I did not rerun the full local suite or exercise live Realtime/browser sessions.

PR #1948, opened 2026-09-20
fix: align conditional tool approvals with normalized execution input

### LINKED ISSUE #1914: needsApproval gets null for omitted optional fields on Zod and strict JSON schema tools

seratch:
This pull request supersedes #1915 and fixes conditional function-tool approvals in the core runner and Realtime so policies inspect isolated, schema-normalized input matching the prepared execution value. Policy mutations cannot change execution input, and Realtime retains prepared input while awaiting approval.

Outputs that cannot be inspected as independent plain data now require manual approval. Both affected packages receive a minor changeset for this compatibility change. Validation may still run application code before approval.

Fixes #1914.


markstuart-oai (approved):
Reviewed 73e19f7edbb795ac5e9337360d86d7def23f6c6f against merge base 9623a226a79268b7dc4e850e4b45025024214041. Conditional policies now inspect isolated input from the same canonical schema preparation used for execution. The synchronous pristine snapshot protects against policy/application mutation; in-memory approvals retain it, while durable resumes re-evaluate current input and policy without persisting transform-only values or comparison evidence. Invalid-input handling, nested approvals and Realtime pending-call ownership remain consistent.

The shared plain-data copier and invocation-scoped retained state serve those concrete lifecycle requirements; the new resume coverage is kept in dedicated modules, and no file newly crosses 1,000 lines. Full-diff review with independent core and Realtime reviewers found no actionable remaining issues.

Validation: all 14 hosted checks passed on this head, including Node 22/24.3 test jobs, package builds, lint, coverage and released/built API checks. This review used pinned source and hosted CI evidence; no local tests or live Realtime calls were run.

dpiet-oai (approved):

PR #1949, opened 2026-09-20
fix: preserve multiline skill frontmatter descriptions

seratch:
This pull request fixes truncated skill descriptions from wrapped plain values and folded/literal frontmatter blocks in both discovery paths. It preserves paragraphs and meaningful indentation, prevents nested fields from overwriting root metadata, and adds regression coverage without a YAML dependency.


markstuart-oai (approved):
Reviewed the complete diff at 31c8662. Both skill-discovery paths share the same parser; wrapped descriptions, folded/literal blocks, paragraph boundaries and nested-field isolation are covered through rendered skill indexes. The bounded scalar parser preserves existing single-line behavior and keeps explicit metadata/index alternatives for unsupported YAML forms. No actionable correctness or structural findings.

All 14 hosted checks passed at this exact head, including Node 22/24.3 tests, coverage, Windows paths and changeset validation. The patch changeset matches the affected core package. Source review only; I did not run local tests or a live sandbox.

PR #1952, opened 2026-09-20
fix: bound recursive sandbox source copy concurrency

seratch:
This pull request fixes recursive sandbox source copies exceeding localDirFiles by using one bounded worker pool per directory traversal. Each worker holds its slot through the host file read and awaited upload, preserving existing path checks, metadata, and separate manifestEntries limits.

Materialization waits for active work to settle after a failure before temporary clones or provider workspaces can be cleaned up. Regression tests cover nested directories, Git clones, explicit and default limits, and failure cleanup.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed a1c44f4f9a9532cba7371beac250286e59f1c06d against the actual merge base. No actionable correctness or structural findings.

One lazy traversal now feeds one file-worker pool per local directory or Git source, so recursive descendants share the configured limit through both the stable file read and awaited upload. The shared limiter drains started work before propagating the first failure, keeping temporary clones and provider workspaces alive until their users settle. Existing path checks, metadata ordering, and the separate manifest-entry concurrency limit remain intact. The controlled-concurrency tests cover nested and Git sources, default/explicit limits, traversal/upload failures, and sibling-entry cleanup.

Validation: independent source review and all 15 hosted checks passed on this head, including Node 22/24.3.x tests, coverage, Windows sandbox checks, changeset validation, and CodeQL. No local tests or live sandbox-provider operations were run.

PR #1953, opened 2026-09-21
fix: preserve filtered history during automatic compaction

seratch:
This pull request fixes automatic Responses compaction reloading and replacing stored history that `sessionInputCallback` or `callModelInputFilter` excluded from the model request. The runner captures the latest successful filtered input and response, and automatic compaction proceeds only when that exchange covers every stored item in order, including repeated occurrences.

Uncovered history is retained without a compaction request. Input compaction uses the covered model-visible representations, and response-ID compaction uses the response associated with that successful exchange. The checks apply to streaming and non-streaming runs, preserve the existing generation ownership guard, and leave explicit manual compaction caller-controlled.


markstuart-oai (approved):
Reviewed the full change at 984002f2b33ae3304595b6ac8e51a44ba632afda. Automatic compaction now requires ordered coverage of every stored occurrence by a successful model exchange; filtered history, terminal tool output and failed streams are retained. The existing ownership guard and explicit manual compaction remain intact. No actionable findings from source and independent review.

All 14 hosted checks passed, including Node 22/24 tests, builds, lint, API contracts and changeset validation. No local tests were run.

dpiet-oai (approved):

PR #1954, opened 2026-09-21
fix: reject privileged Docker storage with read-only host grants

seratch:
This pull request rejects explicit read-only hostPath grants combined with in-container Docker storage requiring SYS_ADMIN. It reuses the existing mount privilege classifier to reject the combination before provider or filesystem effects during creation, manifest updates, and live session reuse, while preserving the existing in-container resume restriction.

Ordinary host grants, provider-native storage, and privileged storage without these grants remain supported. The change includes focused regression coverage and a patch changeset for `@openai/agents-core`.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed b3f49eb, including the prior bind-coverage feedback. The shared validation uses the existing Docker bind and privilege classifiers, covers implicit and default read-only binds, and rejects incompatible create/update/reuse configurations before provider or filesystem effects. The regression tests also check preserved state and supported provider-native storage. No actionable findings.

Validation: source review plus an independent review; all 14 checks passed at this commit. Verified hosted Node 22/24 builds and tests, lint, coverage, Windows sandbox-path checks, and Node 22 API contract checks. No local tests or Docker execution.

PR #1955, opened 2026-09-21
feat: configure phase-two memory consolidation turns

seratch:
This pull request adds `MemoryGenerateConfig.phaseTwoMaxTurns` to control the separate memory consolidation run. The default remains 500, and phase one is unchanged.

The option counts model turns, not time, tokens, or spending. Exhaustion preserves existing failure handling and does not record a successful consolidation selection. Earlier tool writes are not rolled back.


packages/agents-core/src/sandbox/capabilities/memory.ts
```diff
@@ -283,6 +293,7 @@ function normalizeGenerateConfig(
     phaseTwoModel: phaseTwoModel ?? DEFAULT_PHASE_TWO_MODEL,
     phaseOneModelSettings,
     phaseTwoModelSettings,
+    phaseTwoMaxTurns: config.phaseTwoMaxTurns ?? 500,
```
dpiet-oai, line 296: **[Medium] Validate the phase-two turn limit**

`packages/agents-core/src/sandbox/capabilities/memory.ts:296`

The new option is copied straight into `Runner.run`, whose max-turn check is a numeric comparison with no validation. Passing `NaN` or `Infinity` therefore never reaches the limit, while a fractional value such as `1.5` allows two turns; a caller computing the budget can silently lose the cap and incur an unbounded consolidation run.

**Suggested fix:** Reject non-finite, non-integer, or non-positive values in `normalizeGenerateConfig` (and add boundary tests), matching the existing `maxRawMemoriesForConsolidation` validation.

dpiet-oai (commented):
The new phase-two budget needs input validation so non-finite values cannot disable the intended cap; details are inline.

markstuart-oai (approved):
Reviewed the full diff at `5edf5d749efae690b3b2445b6f6e977ac1a80412`. `phaseTwoMaxTurns` reaches the separate consolidation run through the existing normalization and Runner paths. The default remains 500, phase one is unchanged, and manager reuse distinguishes different configured limits.

The public Runner regressions cover explicit/default limits, exhaustion after a tool write, streaming, and owned-session cleanup. They verify that failed consolidation preserves previous selection metadata and completed writes, and does not rerun on close. The implementation stays small and uses the existing turn accounting; no actionable correctness or structural findings after source and independent review.

Validation: all 14 hosted checks passed on this commit. CI run 35574222408 passed Node 22/24 tests, builds, lint, API contract checks, Windows sandbox paths and coverage; changeset validation also passed. No local tests were run.

packages/agents-core/src/sandbox/capabilities/memory.ts
seratch: thanks, but if that's intentional, I think it could be okay and actually the memory ops won't require such turns.

PR #1956, opened 2026-09-21
fix: bound agent tool streaming callback backlogs

seratch:
This pull request limits pending Agent.asTool() stream notifications to 1024 events per invocation. onStreamMaxPendingEvents configures the limit; null restores unlimited buffering.

Overflow aborts the nested producer and closes its iterator through normal tool error handling. Normal draining, event ordering, and non-fatal callback errors are preserved. The limit counts events, not bytes or total memory.

Includes a minor changeset for the changed default. Documentation updates are separately timed with the release.


markstuart-oai (approved):
Reviewed 36a934c. The backlog bound is enforced in the existing producer queue before enqueueing, with a separate policy and abort controller for each invocation. Normal delivery remains ordered; overflow, cancellation and producer failures close the iterator and await producer cleanup without waiting forever on an application callback. The regressions cover sibling isolation, secondary cleanup failures, unlimited buffering and standalone streams. No actionable findings.

The minor changeset appropriately records the changed default and null escape hatch. Validation: source review plus an independent review; all 14 checks passed at this commit. Verified hosted Node 22/24 builds and tests, lint, coverage, Windows sandbox-path checks, and Node 22 API contract checks. No local tests executed.

dpiet-oai (approved):

PR #1957, opened 2026-09-21
fix(core): preserve guarded final outputs in agent tools

seratch:
This pull request fixes default `Agent.asTool()` extraction returning preliminary model text after output guardrails validated a different final output. Guarded runs now return the validated final output, including empty strings, and use the final agent's output type when serializing results after a handoff.

Custom extractors, unguarded output selection, and handled-error precedence retain their existing behavior. Includes regression coverage for streaming, structured handoffs, guardrail rejection, and serialized approval resume, plus a patch changeset for `@openai/agents-core`.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed `9c0b067f2c04ddd98659ff563ec018b462e13837` against the merge base, including an independent review. No actionable findings.

The default extractor now forwards the exact guardrail-checked value without rerunning schema transforms, preserves accepted empty strings, and uses the final agent's output type after handoffs. Custom extraction, approval interruptions, unguarded selection, and handled-error precedence remain intact. The regression cases check the actual parent tool result and next model input across streaming, handoffs, transforms, rejection, and serialized approval resume.

Verified hosted CI at this commit: Node 22 and 24 builds, lint and tests; coverage and Windows path tests; released API compatibility on Node 22. No local tests were run.

PR #1958, opened 2026-09-21
fix: snapshot tool candidates during async enablement

seratch:
This pull request fixes duplicate or skipped tool candidates when application code mutates `agent.tools` in place while an asynchronous `isEnabled` predicate is pending. `Agent.getAllTools()` takes one shallow candidate-array snapshot for each enablement pass, preserving sequential evaluation and exact tool identities while allowing the next resolution to observe updated configuration.

A controlled asynchronous regression verifies evaluation order, candidate identity, shared configuration, and next-call freshness. Includes a patch changeset for `@openai/agents-core`.


markstuart-oai (approved):
Reviewed 16504eb. Taking a shallow candidate snapshot at the start of the enablement pass preserves sequential evaluation and tool identity while preventing concurrent array reordering from duplicating or skipping tools. The next call still sees current configuration. The controlled async regression exercises these observable behaviors, and the patch changeset matches the scope. No actionable findings.

Validation: source review plus an independent review; all 14 checks passed at this commit. Verified hosted Node 22/24 builds and tests, lint, coverage, Windows sandbox-path checks, and Node 22 API contract checks. No local tests executed.

dpiet-oai (approved):

PR #1959, opened 2026-09-21
fix: Isolate agent tool transport settings for explicit models

seratch:
This pull request fixes parent transport overrides being implicitly forwarded to an agent tool's independently configured Model client. When the child Agent or its `asTool.runConfig` selects an explicit Model object, inherited header, query, and body override maps are omitted before applying the child's settings.

Named/default model inheritance, common model settings, and explicit child overrides remain supported. Applications that need transport overrides for an explicit child Model, including a reused Model instance, should supply them in `asTool.runConfig.modelSettings.providerData`.

Intercepted HTTP regression tests cover both explicit Model entry points, model precedence, ordinary inheritance, child credentials and overrides, and parent immutability. Maintainer security review is requested for the parent-to-child client configuration boundary.


markstuart-oai (approved):
Reviewed caf8abc8b3092f0fd7b94d547b8180274d757d61. The inheritance filter reuses the canonical model-selection and transport-alias rules, preserves common settings and named/default-model inheritance, and applies explicit child overrides afterward without mutating the parent. The HTTP regression tests cover both explicit-model routes, model precedence, child request options and parent immutability. No actionable correctness or structural findings after source review and an independent review.

All 14 hosted checks passed at this head, including CI run 35585630558. No local tests were run.

dpiet-oai (approved):

PR #1960, opened 2026-09-21
fix: reject filtered handoffs with server-managed history

seratch:
This pull request fixes handoff filtering with server-managed history by rejecting selected filtered handoffs that use `conversationId` or `previousResponseId`. Local filters cannot remove server-stored context, so the runner raises an actionable `UserError` before the handoff callback, filter invocation, or receiving-agent request.

The restriction covers per-handoff and runner-level filters, including no-op filters, streaming, and serialized approval resume. Callers that need filtering must use explicit client-managed history or a Session without either continuation option. Unfiltered handoffs and client-managed filtering retain their behavior.


markstuart-oai (approved):
Reviewed the selected-handoff validation in both turn-resolution paths, filter precedence, streaming and serialized approval resume, cancellation, and the Responses adapter coverage. The check rejects locally filtered handoffs before the callback/filter or receiving-agent request when continuation IDs keep history on the server; unfiltered and client-managed paths retain their existing behavior. No actionable findings.

Validation: immutable-source and independent review; all 14 hosted checks passed at this head. I did not run local tests or live API calls.

packages/agents-core/src/runner/turnResolution.ts
```diff
@@ -1404,6 +1407,9 @@ export async function resolveTurnAfterModelResponse<
     (run) => !suppressedToolCalls.has(run.toolCall),
   );
   if (handoffRuns.length > 0) {
+    if (!signal?.aborted) {
+      validateHandoffInputFilter(handoffRuns[0].handoff, runner, state);
```
dpiet-oai, line 1428: **[Medium] Validate before custom tool-search callbacks**

`packages/agents-core/src/runner/turnResolution.ts:1411`

This check runs only after `processModelResponseAsync` has awaited `executeCustomClientToolSearch`. A response containing both a selected filtered handoff and an unresolved client `tool_search` call can therefore let the custom executor perform an external mutation or request, then throw this `UserError`; the new cancellation regression itself demonstrates that the executor starts before this validation, so the non-aborted path remains reachable.

**Suggested fix:** Run the same selected-handoff validation from the `beforeClientToolSearch` preflight before invoking the custom executor, and add a regression asserting that the executor is not called.

dpiet-oai (commented):
One actionable current-head finding is included inline.

dpiet-oai (commented):
**[Medium] Validate before custom tool-search callbacks**

`packages/agents-core/src/runner/turnResolution.ts:1411`

The existing current-head inline finding remains verified. A response containing a selected filtered handoff and an unresolved client `tool_search` call can run and await the custom executor before this validation rejects server-managed history, allowing an external side effect that the failed turn cannot safely account for.

**Suggested fix:** Run the same selected-handoff validation from the `beforeClientToolSearch` preflight and cover that the executor is not called.

dpiet-oai (commented):
**[Medium] Validate before custom tool-search callbacks**

`packages/agents-core/src/runner/turnResolution.ts:1411`

The existing current-head inline finding remains independently verified. A response containing a selected filtered handoff and an unresolved client `tool_search` call can run and await the custom executor before this validation rejects server-managed history, allowing an external side effect that the failed turn cannot safely account for.

**Suggested fix:** Run the selected-handoff validation from the `beforeClientToolSearch` preflight and cover that the executor is not called. No duplicate inline comment was added.

dpiet-oai (commented):
**[Medium] Validate before custom tool-search callbacks**

`packages/agents-core/src/runner/turnResolution.ts:1411`

The existing current-head inline finding remains independently verified. A response containing a selected filtered handoff and an unresolved client `tool_search` call can run and await the custom executor before this validation rejects server-managed history, allowing an external side effect that the failed turn cannot safely account for.

**Suggested fix:** Run the selected-handoff validation from the `beforeClientToolSearch` preflight and cover that the executor is not called.

seratch:
@dpiet-oai Can you take another look when you have a chance?

PR #1962, opened 2026-09-21
fix: redact default function-tool failure details

seratch:
This pull request fixes default function-tool failures exposing exception details in model feedback and handled-error traces. Default failures now use fixed messages, and custom-handler trace details respect the active runner's sensitive-data setting.

Applications needing specific feedback should provide an `errorFunction` that selects application-approved information, or use `errorFunction: null` to propagate the original failure. Custom handler output, structured-output validation, and cancellation behavior remain unchanged.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed default function-tool failure feedback and traces, Runner parse failures, Realtime approval parsing, explicit custom/null handlers, and invocation-scoped trace policy through timeout wrappers and overlapping runs. Default handled failures now return fixed feedback without inspecting exception details, while the documented application-controlled escape hatches remain intact. No actionable findings.

Validation: immutable-source and independent review; all 14 hosted checks passed at this head. I did not run local tests or live API calls.

dpiet-oai (approved):

PR #1964, opened 2026-09-21
fix: omit reasoning from default OpenAI trace exports

seratch:
This pull request omits structured generation reasoning from traces sent to the default OpenAI ingest endpoint before applying payload limits. It covers AI SDK reasoning items and assistant content blocks, plus supported Chat Completions reasoning fields, while preserving ordinary text, tool data, usage, original spans, model results, and custom destinations.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed `cf2fb8f` including the exporter, both regression suites and the actual AI SDK/Chat Completions trace and replay producers. Filtering covers the supported reasoning locations before truncation while preserving original spans, model results, ordinary data, usage and custom destinations. The earlier replay-content finding is addressed. The patch changeset covers both changed packages. No actionable findings after source and independent review. All 14 hosted checks passed; no local tests or live ingest requests were run.

dpiet-oai (approved):

PR #1966, opened 2026-09-21
fix: require interactive approval for host shell examples

seratch:
This pull request updates the local shell example to require explicit interactive approval for every command batch and excludes both host-shell aliases from automatic example runs. It safely displays commands, restricts child-process environment inheritance, and closes child stdin while preserving existing execution results.

Approved commands retain access to host files and the network; the example points to the existing container shell examples for isolated execution.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed `dad5d75` including the host-shell example, automation aliases, tests and approval dispatch. Each supported command batch requires a fresh interactive decision; automatic/non-TTY runs are denied, command display is escaped, and child environment/stdin handling preserves the existing execution outcomes. The example clearly retains host file/network access and points to container alternatives. No actionable findings after source and independent review. All 14 hosted checks passed; no local tests or live shell/model execution were run.

dpiet-oai (approved):

PR #1972, opened 2026-09-21
fix: pass changeset prompt Git arguments without a shell

seratch:
This pull request fixes local changeset prompt generation by passing Git refs and package paths as argument arrays instead of shell command text. It preserves committed, staged, unstaged, and untracked diff collection and existing prompt output.

Adds isolated regression tests that record subprocess arguments and check prompt sections, release-only filtering, CI output mode, and optional command failures. The tests run through the existing changeset fixture runner.


jbeckwith-oai (approved):
Reviewed exact head 6edcdeffabf048211cdf201c9ac0703d644a49c7. Git subprocess arguments are now passed as argv, package paths are separated from options, and the regression coverage preserves committed, staged, unstaged, untracked, CI-output, and optional-failure behavior. All exact-head checks are green.

PR #1973, opened 2026-09-21
fix: preserve current approval decisions during nested restore

seratch:
This pull request fixes nested agent resumes restoring an older permanent function approval after the application has changed the current decision to alwaysReject. Serialized owner-scoped function approvals now preserve an existing permanent decision, including its per-call exceptions and rejection message.

Saved approvals still restore when no current permanent decision exists. The change preserves owner and qualified-tool identities, existing context strategies, and the serialized schema. Recording-only regressions cover nested resumes with and without an outer serialization round trip, restoration strategies, both permanent-decision directions, and owner separation.


markstuart-oai (approved):
Reviewed nested restoration through RunState and Agent.asTool. The canonical owner-scoped merge now preserves the current permanent function decision, including exceptions and rejection messages, while empty-context restoration and explicit replacement retain their intended behavior. The regression tests exercise real nested resumes and qualified/owner-separated tools. No actionable findings.

All 14 exact-head hosted checks passed, including Node 22/24 tests and coverage. Source review only; no local tests run.

PR #1974, opened 2026-09-21
fix: Apply static MCP filters without run context

seratch:
This pull request fixes static MCP tool filtering when `getAllMcpTools` is called without both an agent and a run context. Previously, these calls could return executable tools excluded by the server's allow or block list.

Apply static filters in the shared discovery path for both public call forms while preserving the full-context requirement for callable filters and reapplying policy to raw cached listings. Add recording-only discovery and Runner regressions for missing and partial context, cache reuse, permitted execution, and excluded tool calls.


markstuart-oai (approved):
Reviewed static MCP filtering through both discovery overloads, cached listings and Agent/Runner dispatch. Static policies now apply without complete context, while callable filters retain their existing context requirement and cached discovery still reevaluates policy. The change stays in the shared discovery boundary. No actionable findings.

All 14 exact-head hosted checks passed, including Node 22/24 tests and coverage. Source review only; no local tests run.

